PR #35, opened 2018-05-06
Automatically convert types to string in string concatenation operators

ryzhyk:
The compiler automatically inserts code to convert abitrary values to strings when they occur
inside interpolated strings or as a second argument to string concatenation.  
Values of primitive types (`string`, `int`, `bit`, and `bool`) are converted using 
builtin function
```
function __builtin_2string(x:'A): string
```

For example, the assignment statement in
```
x: int;
y = $"x:{x}";
```
is equivalent to:
```
y = $"x:{__builtin_2string(x)}";
```

For user-defined types, the compiler injects a call to a function whose name is formed from
the type name by changing the first letter of the type name to lower case (if it is in upper case) 
and adding the `"2string"` suffix.  The function must take exactly one 
argument of the given type and return a string.  
Compilation fails if a function with this name and signature is not found.

For example, the last statement in
```
typedef udf_t = Cons1 | Cons2{f: int}
function udf_t2string(x: udf_t): string = ...
x: udf_t;

y = $"x:{x}";
```
is equivalent to:
```
y = $"x:{udf_t2string(x)}";

### CODE REVIEW

File: doc/language_reference/language_reference.md [hunk 1]
```diff
@@ -335,6 +335,45 @@ pattern ::= (* tuple pattern *)
           | "_"             (* wildcard, matches any value *)
 ```
 
+### Automatic string conversion
+
+The compiler automatically inserts code to convert abitrary values to strings when they occur
```
mbudiu-vmw: Frankly I believe that this should not appear in the language reference. Users don't need to know what the compiler does internally. Some of this discussion should be moved into a comment, but the description of the conversion function naming should be kept.

src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -557,3 +523,46 @@ exprCheckMatchPatterns d ctx e@(EMatch _ x cs) = do
...
+exprInjectStringConversions d ctx e@(EBinOp p Concat l r) | (te == tString) && (tr /= tString) = do
+    -- find string conversion function
+    fname <- case tr of
+                  TBool{}     -> return bUILTIN_2STRING_FUNC
+                  TInt{}      -> return bUILTIN_2STRING_FUNC
+                  TString{}   -> return bUILTIN_2STRING_FUNC
+                  TBit{}      -> return bUILTIN_2STRING_FUNC
+                  TUser{..}   -> return $ mk2string_func typeName
+                  TOpaque{..} -> return $ mk2string_func typeName
+                  TTuple{}    -> err (pos r) "Automatic string conversion for tuples is not supported"
```
mbudiu-vmw: I don't see why this is difficult. Can't you just iterate over the tuple members?

src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -557,3 +523,46 @@ exprCheckMatchPatterns d ctx e@(EMatch _ x cs) = do
...
+                  TVar{..}    -> err (pos r) $
```
mbudiu-vmw: Is this a type variable?
If so maybe the conversion should be done after unification, and there shouldn't be any more type variables left.

src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -557,3 +523,46 @@ exprCheckMatchPatterns d ctx e@(EMatch _ x cs) = do
...
+                                     "Cannot automatically convert " ++ show r ++ 
+                                     " of variable type " ++ tvarName ++ " to string"
+                  TStruct{}   -> error "unexpected TStruct in exprInjectStringConversions"
+    f <- case lookupFunc d fname of
+              Nothing  -> err (pos r) $ "Cannot find declaration of function " ++ fname ++ 
+                                        " needed to convert expression " ++ show r ++ " to string"
+              Just fun -> return fun
+    let arg0 = funcArgs f !! 0
+    -- validate its signature
+    assert (isString d $ funcType f) (pos f)
```
mbudiu-vmw: I don't know what assert does in Haskell, but in general you should not report errors that are caused by user programs using assertions.

test/datalog_tests/function.fail.ast.expected
```diff
@@ -15,3 +15,6 @@ error: ./test/datalog_tests/function.fail.dl:5:1-9:1: Recursive function definit
 error: ./test/datalog_tests/function.fail.dl:9:5-9:17: Type constructor in the left-hand side of an assignment is only allowed for types with one constructor,  but "option_t" has multiple constructors
 
 error: ./test/datalog_tests/function.fail.dl:11:16-11:27: Type constructor in the left-hand side of an assignment is only allowed for types with one constructor,  but "option_t" has multiple constructors
+
+error: ./test/datalog_tests/function.fail.dl:6:5-7:1: Expression parameterized(string, int) has unknown type in CtxTop
```
mbudiu-vmw: This error message looks confusing. Shouldn't the error indicate that unification failed between the types of x and y?

src/Language/DifferentialDatalog/Validate.hs
ryzhyk: this is our assert function that will generate a nice compilation error

test/datalog_tests/function.fail.ast.expected
ryzhyk: You're right. This is a result of how type inference works right now. I think I know how to fix this. I will add an issue for the moment and will fix this in a separate PR.

src/Language/DifferentialDatalog/Validate.hs
mbudiu-vmw: What is the name of the conversion function for tuples?

src/Language/DifferentialDatalog/Validate.hs
ryzhyk: Shall I call it `check`?

mbudiu-vmw: That's a good name.

doc/language_reference/language_reference.md
ryzhyk: OK, I will rephrase this to not refer to what compiler does.

src/Language/DifferentialDatalog/Validate.hs
ryzhyk: This is doable, but not easy. Problem is, we will generate polymorphic Rust functions, with unification (specialization, actually) being done by the Rust compiler. In a richer type system, we'd be able to write that the type variable must be of a printable type, but I really don't want to add type classes or inheritance to the language. Alternatively, we could infer that a type variable must be printable from function implementation, and convert its 2string function to the Rust Display trait implementation.

I'd rather not spend time on this now, as I want to make progress. How about I add this as an issue for now?

src/Language/DifferentialDatalog/Validate.hs
ryzhyk: There isn't one. You have to write your own function for each user-defined tuple.

mbudiu-vmw: How do you associate a function with a tuple type?

PR #38, opened 2018-05-09
Improved type error reporting

ryzhyk:
propagate the reason why expression type cannot be determined to the compiler error message.

Old error message:
```
error: ./test/datalog_tests/function.fail.dl:6:5-7:1: Expression parameterized(string, int) has unknown type in CtxTop
```

New error message
```
error: ./test/datalog_tests/function.fail.dl:6:5-8:1: Conflicting types string and int for type variable 'A in call to function parameterized (x: 'A, y: 'A): 'A
```

### DIFF
--- a/src/Language/DifferentialDatalog/Type.hs
+++ b/src/Language/DifferentialDatalog/Type.hs
@@ -28,9 +28,10 @@ module Language.DifferentialDatalog.Type(
     typeUserTypes,
     typeIsPolymorphic,
     unifyTypes,
-    exprType, exprType', exprType'', exprTypeMaybe, exprNodeType,
---    exprTypes,
-    exprTraverseTypeM,
+    exprType,
+    exprType',
+    exprType'',
+    exprNodeType,
     typ', typ'',
     isBool, isBit, isInt, isString, isStruct, isTuple,
     checkTypesMatch,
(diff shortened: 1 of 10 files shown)


src/Language/DifferentialDatalog/Type.hs
```diff
@@ -28,9 +28,10 @@ module Language.DifferentialDatalog.Type(
     typeUserTypes,
     typeIsPolymorphic,
     unifyTypes,
-    exprType, exprType', exprType'', exprTypeMaybe, exprNodeType,
---    exprTypes,
-    exprTraverseTypeM,
+    exprType,
+    exprType',
```
mbudiu-vmw: I am a bit worried that you need to export primed names; this means either that the names are wrong, or that you are not hiding enough.

src/Language/DifferentialDatalog/Type.hs
```diff
@@ -101,39 +102,40 @@ typeIsPolymorphic TOpaque{..} = any typeIsPolymorphic typeArgs
...
+        (TVar _ n1       , t)                -> return $ M.singleton n1 t
+        _                                    -> err p $ "Cannot match expected type " ++ show a ++ " and actual type " ++ show c ++ " " ++ ctx  
     where a' = typ'' d a
           c' = typ'' d c
 
 -- | Visitor pattern implementation that carries type information
 -- through the syntax tree.
-exprTraverseTypeM :: (Monad m) => DatalogProgram -> (ECtx -> ExprNode (Maybe Type) -> m ()) -> ECtx -> Expr -> m ()
-exprTraverseTypeM d = exprTraverseCtxWithM (\ctx e -> return $ exprNodeType d ctx e)
+--exprTraverseTypeM :: (Monad m) => DatalogProgram -> (ECtx -> ExprNode (Maybe Type) -> m ()) -> ECtx -> Expr -> m ()
```
mbudiu-vmw: why not just delete this code?

test/datalog_tests/function.fail.ast.expected
```diff
@@ -16,5 +16,6 @@ error: ./test/datalog_tests/function.fail.dl:9:5-9:17: Type constructor in the l
 
 error: ./test/datalog_tests/function.fail.dl:11:16-11:27: Type constructor in the left-hand side of an assignment is only allowed for types with one constructor,  but "option_t" has multiple constructors
 
-error: ./test/datalog_tests/function.fail.dl:6:5-7:1: Expression parameterized(string, int) has unknown type in CtxTop
-CtxFunc            use_parameterized
+error: ./test/datalog_tests/function.fail.dl:6:5-8:1: Conflicting types string and int for type variable 'A in call to function parameterized (x: 'A, y: 'A): 'A
```
mbudiu-vmw: This is very good.

src/Language/DifferentialDatalog/Type.hs
ryzhyk: I guess you're right, but these functions (`typ`, `typ'`, `typ''` and `exprType`, `exprType'`, `exprType'''`) are ubiquitous and I like using short names for them. Otherwise they'd have to be called, e.g., `typ`, `typCanonical`, `typ????` (not even sure what the name should be)

src/Language/DifferentialDatalog/Type.hs
ryzhyk: done

PR #42, opened 2018-05-16
change string interpolation syntax: $"foo{x}var" -> $"foo${x}var"

ryzhyk:


test/datalog_tests/ovn-ftl.ast.expected
```diff
@@ -27,7 +27,7 @@ for (lrp in Logical_Router_Port if lrp.enabled)
     {
         let P = lrp.name, E = lrp.mac in
             {
-                Flow ( lrp.lr, LR_IN_ADMISSION{}, 50, (((("inport == $" ++ P{}) ++ " && (eth.mcast || eth.dst == $") ++ E{}) ++ ")"), "{ next; }" )
+                Flow ( lrp.lr, LR_IN_ADMISSION{}, 50, (((("inport == " ++ P{}) ++ " && (eth.mcast || eth.dst == ") ++ E{}) ++ ")"), "{ next; }" )
```
mbudiu-vmw: This looks wrong: why is it P{} instead of just P?

ryzhyk: Capital letters are reserved for type constructors, so expression P is interpreted as a call to a type constructor with no arguments. Once you've hooked expression validation code into FTL, it will complain about P being invalid variable name.

PR #46, opened 2018-05-26
Convert ftl

mbudiu-vmw:
The only case I haven't done is the match. 
This PR is submitted to get some comments on the coding style, we should merge it only after I do the match too.

Also, once this is checked-in the ovn-ftl.dl does not compile any more, because the relation declarations are missing. We'll have to do something about it.

You should configure your text editor to delete spaces at the end of lines, we'll have smaller diffs.


src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -90,34 +91,102 @@ checkNoRecursion :: (MonadError String me) => DatalogProgram -> me ()
...
+     map (\r -> Rule (rulePos r) (ruleLHS r) (toAdd : ruleRHS r)) rules
+
+convertFtlStatement :: (MonadError String me) => Statement -> me [Rule]
+convertFtlStatement (ForStatement p e r Nothing s) = do
+    rules <- convertFtlStatement s
+    let atom = Atom p r [("", e)]
+    let rhs = RHSLiteral True atom
+    let rules' = addRhsToRules rhs rules
+    return rules'
+convertFtlStatement (ForStatement p e r (Just c) s) = do
```
ryzhyk: One way to avoid two near-identical implementations for `ForStatement`:

```
convertFtlStatement (ForStatement p e r mc s) = do
    rules <- convertFtlStatement s
    let atom = Atom p r [("", e)]
    let rhs0 = RHSLiteral True atom
    let rhs1 = map RHSCondition $ maybeToList mc
    return $ map (\r -> r{ruleRHS = rhs0 : rhs1 ++ ruleRHS r}) rules
```

src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -90,34 +91,102 @@ checkNoRecursion :: (MonadError String me) => DatalogProgram -> me ()
...
+
+-- Desugar atoms: convert all relational atoms to named field syntax.
 progDesugarAtoms :: (MonadError String me) => DatalogProgram -> me DatalogProgram
 progDesugarAtoms d = do
-    rules' <- mapM (\r -> do lhs <- mapM (atomDesugarArgs d) $ ruleLHS r
-                             rhs <- mapM (rhsDesugarAtoms d) $ ruleRHS r 
-                             return r{ruleLHS = lhs, ruleRHS = rhs})
-                   $ progRules d
-    return d{progRules = rules'}    
+    -- Drop rules with an empty LHS
```
ryzhyk: It would be cleaner if the function that introduces rules with empty LHS filtered them out before adding them to the spec.

mbudiu-vmw:
I cannot approve my won PR.
Please file an issue to keep track of the commented-out code, and let's merge this. 
It looks very good.

PR #47, opened 2018-06-05
Record types

ryzhyk:
Elements of relations now have types. The old relation declaration syntax now works as syntactic sugar that creates a new type and a relation whose elements belong to this type. The type has the same name as the relation and has a single type constructor with the same name. 

```
relation R(f1: int, f2: bool)
```
is equivalent to:
```
typedef R = R{f1: int, f2:bool}
relation R[R]
```
where the second form is the new, more general, declaration syntax. Relations can have arbitrary types, not just structs, e.g.,

```
relation R2[(int, int)]
```

There is also new syntax for atoms:
```
atom ::= rel_name "(" expr (,expr)* ")"
       | rel_name "(" "." arg_name "=" expr [("," "." arg_name "=" expr)*] ")"
       | rel_name "[" expr "]"
```

The first two forms are the same as before. The third one is new and more general, e.g., atom

```
Rel(val1, val2)
```
expands to
```
Rel[Rel1(val1, val2)]
```

### DIFF
--- a/src/Language/DifferentialDatalog/Parse.hs
+++ b/src/Language/DifferentialDatalog/Parse.hs
@@ -152,25 +152,40 @@ removeTabs = do s <- getInput
                 setInput s'
 
 withPos x = (\ s a e -> atPos a (s,e)) <$> getPosition <*> x <*> getPosition
+withPosMany x = (\ s as e -> map (\a -> atPos a (s,e)) as) <$> getPosition <*> x <*> getPosition
 
-data SpecItem = SpType         TypeDef
-              | SpRelation     Relation
-              | SpRule         Rule
-              | SpFunc         Function
-              | SpStatement    Statement
+data SpecItem = SpType      TypeDef
+              | SpRelation  Relation
+              | SpRule      Rule
+              | SpFunc      Function
+              | SpStatement Statement
+
+instance WithPos SpecItem where
+    pos   (SpType         t)   = pos t
+    pos   (SpRelation     r)   = pos r
+    pos   (SpRule         r)   = pos r
+    pos   (SpFunc         f)   = pos f
+    pos   (SpStatement    s)   = pos s
+    atPos (SpType         t) p = SpType      $ atPos t p
+    atPos (SpRelation     r) p = SpRelation  $ atPos r p
+    atPos (SpRule         r) p = SpRule      $ atPos r p
+    atPos (SpFunc         f) p = SpFunc      $ atPos f p
+    atPos (SpStatement    s) p = SpStatement $ atPos s p
+   
 
 datalogGrammar preamble = removeTabs *> ((optional whiteSpace) *> spec preamble <* eof)
 exprGrammar = removeTabs *> ((optional whiteSpace) *> expr <* eof)
 
 spec preamble = do
-    items <- many decl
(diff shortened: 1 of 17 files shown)


src/Language/DifferentialDatalog/Parse.hs
```diff
@@ -191,37 +206,49 @@ spec preamble = do
...
+relation = do
+    ground <-  True <$ reserved "ground" <* reserved "relation" 
+           <|> False <$ reserved "relation"
+    relName <- relIdent
+    ((do start <- getPosition
+         fields <- parens $ commaSep arg
+         end <- getPosition
+         let p = (start, end)
+         let tspec = TStruct p [Constructor p relName fields]
+         let tdef = TypeDef nopos relName [] $ Just tspec
```
mbudiu-vmw: Maybe you can use the positions you have.

src/Language/DifferentialDatalog/Parse.hs
```diff
@@ -191,37 +206,49 @@ spec preamble = do
...
+           return [SpRelation rel]))
+        
 arg = withPos $ (Field nopos) <$> varIdent <*> (colon *> typeSpecSimple)
 
-parseForStatement = withPos $ ForStatement nopos <$ reserved "for"
-                                                 <*> (symbol "(" *> expr)
-                                                 <*> (reserved "in" *> relIdent)
-                                                 <*> (optionMaybe (reserved "if" *> expr))
-                                                 <*> (symbol ")" *> statement)
+parseForStatement = ForStatement nopos <$ reserved "for"
```
mbudiu-vmw: Why remove the withPos?

mbudiu-vmw: Or if you want to remove it here why not elsewhere as well?

src/Language/DifferentialDatalog/Parse.hs
```diff
@@ -250,14 +276,13 @@ parseLetStatement = withPos $ (LetStatement nopos) <$ reserved "let"
...
 
-rule = withPos $ Rule nopos <$>
-                 (commaSep1 atom) <*>
-                 (option [] (reservedOp ":-" *> commaSep rulerhs)) <* dot
+rule = Rule nopos <$>
+       (commaSep1 atom) <*>
+       (option [] (reservedOp ":-" *> commaSep rulerhs)) <* dot
 
-rulerhs =  (do _ <- try $ lookAhead $ (optional $ reserved "not") *> relIdent *> symbol "("
+rulerhs =  (do _ <- try $ lookAhead $ (optional $ reserved "not") *> relIdent *> (symbol "(" <|> symbol "[")
```
mbudiu-vmw: I was expecting a closed "]" somewhere.

ryzhyk: This is a lookahead, it does not consume any input; just checks that we are looking at an atom before actually parsing it, so seeing the opening bracket is enough

src/Language/DifferentialDatalog/Parse.hs
ryzhyk: I implemented `WithPos` for `SpecItem`, so now I only need to invoke `withPos` once in `decl` and not for each type of `SpecItem`

Issue #73, opened 2018-08-24
Get rid of `let` syntax


ryzhyk:
We use `let` in FTL, and `var` in the rest of the language. There is not reason for these two forms of variable declaration.

PR #74, opened 2018-08-28
Fix for issue #73

mbudiu-vmw:


src/Language/DifferentialDatalog/Syntax.hs
```diff
@@ -419,18 +419,19 @@ instance Show Rule where
...
+    (==) (Assignment _ l1 t1 r1) (Assignment _ l2 t2 r2) =
+      l1 == l2 && t1 == t2 && r1 == r2
 
 instance WithPos Assignment where
     pos = assignPos
     atPos r p = r{assignPos = p}
 
 instance PP Assignment where
-    pp (Assignment _ l r) = pp l <+> "=" <+> pp r
+    pp (Assignment _ l t r) = pp l <> ":" <+> pp t <+> "=" <+> pp r
```
ryzhyk: This does not look correct. Should be something like:
```
pp l <> (maybe empty (\t' -> ":" <+> pp t') t) <+> "=" <+> pp r
```

src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -127,6 +127,10 @@ addRhsToRules:: RuleRHS -> [Rule] -> [Rule]
 addRhsToRules toAdd rules =
      map (\r -> r{ruleRHS=(toAdd : ruleRHS r)}) rules
 
+convertAssignment :: Assignment -> Expr
+convertAssignment (Assignment p l Nothing r) = eSet (E $ EVarDecl p l) r
```
ryzhyk: It's better not to lose source position for correct error reporting:

```
E $ ESet p (E $ EVarDecl p l) r
```

PR #78, opened 2018-09-05
Hash functions + faster 128-bit arithmetic

ryzhyk:
- this adds two new extern functions: hash64 and hash128
- it also switches to using Rust's u128 type for 128-bit bitvectors instead of the more expensive BigUint


rust/cmd_parser/from_record.rs
```diff
@@ -108,6 +108,32 @@ fn test_u64() {
     assert_eq!(u64::from_record(&Record::Int((-0xabcdef).to_bigint().unwrap())), Err("cannot conver -11259375 to u64".to_string()));
 }
 
+impl FromRecord for u128 {
+    fn from_record(val: &Record) -> Result<Self, String> {
+        match val {
+            Record::Int(i) => {
+                match i.to_u128() {
+                    Some(x) => Result::Ok(x),
+                    None    => Result::Err(format!("cannot conver {} to u128", i))
```
mbudiu-vmw: typo in error message here and below.

test/datalog_tests/rules.dl
```diff
@@ -32,3 +32,14 @@ relation R7[R7]
...
+typedef S = S {
+    f1: (bool, bool),
+    f2: bit<32>
+}
+input relation Table1(id: bit<64>, name: string)
+input relation Table2(name: string, f2: bigint, f3: S)
+
+relation Table12(id: bit<128>, name: string, f3: S)
+
+Table12(hash128((id,name,f2,f3)), name, f3) :- Table1(id, name), Table2(name, f2, f3).
```
mbudiu-vmw: do you want to make this hex?

PR #83, opened 2018-09-07
Main DDlog program

ryzhyk:


app/Main.hs
```diff
@@ -81,24 +87,26 @@ main = do
...
+parseValidate Config{..} = do
+    d <- parseDatalogFile True confDatalogFile
+    case validate d of
+         Left e   -> errorWithoutStackTrace $ "error: " ++ e
+         Right d' -> return d'
+
+compileProg :: Config -> IO ()
+compileProg conf@Config{..} = do
+    let specname = takeBaseName confDatalogFile
+    prog <- parseValidate conf
```
mbudiu-vmw: will this handle errors properly?

ryzhyk: You mean readFile? It will fail with a meaningful error message. And so should parseValidate

Issue #85, opened 2018-09-10
OVN integration TODOs


ryzhyk:
TODO list for DDlog/OVN integration:

- [ ] generate DDlog schema from OVSDB schema.
   - This can be done mostly automatically from OVSDB schema files; however we will need some way to specify which OVSDB relations are inputs and which our outputs (can they be both?). Should we add those as annotations to the OVSDB schema or just specify them as command line arguments to the schema conversion tool?
- [ ] port rules from `northd.c` to `northd.dl`
- [ ] create data sets for standalone testing of `northd.dl` (i.e., outside of OVN)
- [ ] generate runtime OVSDB input/output adapters for DDlog
- [ ] create OVN infrastructure to test `northd.dl` against `northd.c` on a live system
- [ ] integration testing
- [ ] scale testing (and CPU/memory profiling)

HehuaTang:
when will be  this feature released?

ryzhyk:
There is a functional implementation of OVN northd in DDlog in a branch: https://github.com/ovn-org/ovn/tree/ddlog-dev-v2 (see  documentation here: https://github.com/ovn-org/ovn/tree/ddlog-dev-v2/northd/docs).  However not all recent northd features have been ported to DDlog and it is not part of an official OVN release. I am hoping both issues will be resolved sometime soon, but I am not an OVN maintainer, so can't make any promises.

HehuaTang:
Thanks

PR #86, opened 2018-09-11
convert OVSDB schema to DDlog relation declarations

ryzhyk:
@justinpettit, @blp : this tool takes multiple OVSDB schema files and generates an equivalent DDlog schema that looks like [this](https://github.com/ryzhyk/differential-datalog/blob/ovsdb_adapter/test/ovsdb/ovn.dl.expected).

There are a few issues with this translation that we will gradually fix:
- Long relation names that include schema name, e.g., `relation OVN_Northbound_Logical_Switch_Port`. This should go away once we've implemented name spaces.
- DDlog currently does not yet support multi-file programs, meaning that there is currently no convenient way to combine generated and manual code.
- we add underscore to OVSDB field names that clash with DDlog and Rust keywords: `match` and `type`.


src/Language/DifferentialDatalog/OVSDB/Compile.hs
```diff
@@ -0,0 +1,127 @@
...
+         ColumnTypeComplex ct -> mkComplexType isinput tname c ct
+
+mkColName :: String -> Doc
+mkColName "type" = "_type"
+mkColName "match" = "_match"
+mkColName n      = pp $ map toLower n
+
+mkAtomicType :: (MonadError String me) => AtomicType -> me Doc
+mkAtomicType IntegerType{} = return "integer"
+mkAtomicType (RealType p)  = err p "\"real\" types are not supported"
```
mbudiu-vmw: Could this be an external type?
Is this needed?

src/Language/DifferentialDatalog/OVSDB/Compile.hs
```diff
@@ -0,0 +1,127 @@
...
+                      Unlimited -> fromIntegral (maxBound::Word64))
+                    maxComplexType
+    check (max >= min) (pos t) $ "min bound exceeds max bound"
+    check (min == 0 || min == 1) (pos t) $ "min bound must be 0 or 1"
+    check (max > 0) (pos t) $ "max bound must be greater than 0"
+    check (max /= 1 || isNothing valueComplexType) (pos t)
+          $ "Cannot handle key-value pairs when max bound is 1"
+    key <- mkBaseType keyComplexType
+    case (min, max) of
+         (1,1) -> return $ Left $ mkColName columnName <> ":" <+> key
```
mbudiu-vmw: This looks like a weird case. I wonder whether it should be an error. Mathematically it makes sense, but I don't expect anyone will write a program like that.

src/Language/DifferentialDatalog/OVSDB/Compile.hs
```diff
@@ -0,0 +1,127 @@
...
+         _     -> Right <$> mkSubtable isinput tname columnName t
+
+mkSubtable :: (MonadError String me) => Bool -> String -> String -> ComplexType -> me Doc
+mkSubtable isinput tname cname t@ComplexType{..} = do
+    keytype <- mkBaseType keyComplexType
+    valtype <- maybe (return []) (\v -> (\vt -> ["value:" <+> vt]) <$> mkBaseType v) valueComplexType
+    let cols = [ mkColName tname <> ": uuid"
+               , "key:" <+> keytype] ++ valtype
+    let prefix = if isinput then "input" else empty
+    return $ prefix <+> "relation" <+> pp tname <> "_" <> pp cname <> "(" $$
```
mbudiu-vmw: Are you checking whether this name clashes with the existing ones?

test/ovsdb/ovn.dl.expected
```diff
@@ -0,0 +1,611 @@
+
+typedef integer      = bit<64>
+typedef uuid         = bit<128>
+typedef option_t<'A> = Some{x: 'A}
+                     | None
+
+input relation OVN_Northbound_NB_Global(
```
mbudiu-vmw: The next step is to validate this automatically somehow.
For example by loading some data into this database.

src/Language/DifferentialDatalog/OVSDB/Compile.hs
ryzhyk: Yep, could be `extern`. There is not a single occurrence of this type in OVN, though, so probably not worth it.

src/Language/DifferentialDatalog/OVSDB/Compile.hs
ryzhyk: True. I'll check with Ben/Justin.

src/Language/DifferentialDatalog/OVSDB/Compile.hs
ryzhyk: Nope, but DDlog will complain if it does.

test/ovsdb/ovn.dl.expected
ryzhyk: Yep, it's on the TODO list (#85) :)

PR #91, opened 2018-09-17
Performance improvements

ryzhyk:
See https://github.com/frankmcsherry/differential-dataflow/issues/113 and https://github.com/frankmcsherry/differential-dataflow/issues/114


rust/differential_datalog/profile.rs
```diff
@@ -0,0 +1,109 @@
+//! Memory profile of a DDlog program.
```
mbudiu-vmw: Isn't there some overhead for adding all this code, even if profiling is disabled?

ryzhyk: Should be negligible. The runtime callback is invoked once for every batch and basically just updates two keys in two different maps.

PR #93, opened 2018-09-18
Change test script to run on MacOS

remysucre:
When using `cwd` in `withCreateProcess`, absolute paths should be used to ensure portability. See doc for [`cmdspec :: CmdSpec`](http://hackage.haskell.org/package/process-1.6.4.0/docs/System-Process.html). See also haskell/process#87


ryzhyk:
Thanks for the contribution!

PR #100, opened 2018-09-20
Parallelize Travis tests

remysucre:
I've set up `tutorial`, `ovn_ftl` and `path` to run in parallel both on linux and osx. The `ovn` test didn't print anything for 10 min therefore timed out; if we make these tests more verbose, it may be possible to run all tests in parallel (as long as they each finish within 50 min and prints every 10 min).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,9 @@ env:
 
 addons: {apt: {packages: [ghc-8.2.2], sources: [hvr-ghc]}}
 
-os: osx
+os: 
+  - osx
+  - linux
 
 matrix:
   fast_finish: true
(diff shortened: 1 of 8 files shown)


ryzhyk:
I wonder if tests slow each other down as they run in parallel. Adding `-j 1` to test arguments (`stack test --ta '-j 1 -p .....'`) might help. I can make test more verbose, but this will make test output look messy...

ryzhyk:
That's a good idea. I'll work on it. Travis sometimes takes >10minutes even for a single test without anything else running in the background, so tests definitely need to be more noisy.

PR #101, opened 2018-09-20
Module system.

ryzhyk:
See the update tutorial and language reference for description.


src/Language/DifferentialDatalog/Syntax.hs
```diff
@@ -659,25 +590,53 @@ funcShowProto Function{..} = render $
...
+                     , importModule :: ModuleName
+                     , importAlias  :: ModuleName 
+                     }
+
+instance WithPos Import where
+    pos = importPos
+    atPos i p = i { importPos = p }
+
+instance PP Import where
+    pp Import{..} = "import" <+> pp importModule <+> "as" <+> pp importAlias
```
mbudiu-vmw: does this work when there is no alias?

test/datalog_tests/function.ast.expected
```diff
@@ -1,11 +1,10 @@
 typedef Alt = C0{x: bit<32>} | C1{x: bit<32>}
 typedef C = C{f1: string, f2: string}
-extern type Set<'A>
-extern type Vec<'A>
 typedef nested_t = N{field: C}
 typedef option_t<'A> = None{} | Some{value: 'A}
+extern type std.Set<'A>
```
mbudiu-vmw: What happens if you (re)declare a function from std?
This may be a problem if you output a program and try to compile that one. It should not have the std definitions...

test/datalog_tests/statements.ast.expected
```diff
@@ -67,13 +42,38 @@ Z(.field=(a - 1)) :- X(.field=a), match (a) {
                                       2 -> false,
                                       _ -> true
                                   }.
+Z(.field=a) :- X(.field=a), match (a) {
```
mbudiu-vmw: Should you dump these in alphabetical order? Would make it easier to diff.

src/Language/DifferentialDatalog/Syntax.hs
ryzhyk: good catch! Imports are always eliminated before printing, which is why I did not spot this.

test/datalog_tests/function.ast.expected
ryzhyk: DDlog will complain just like with any name clash. Maybe I misunderstand the issue.

mbudiu-vmw: I type a program, and then I pretty-print it (is there an option?). Then I can't compile the result anymore, because it has duplicate declarations.

PR #105, opened 2018-09-25
Allow underscores in numeric literals

mbudiu-vmw:


src/Language/DifferentialDatalog/Parse.hs
```diff
@@ -480,16 +484,16 @@ sradval =  ((try $ string "'b") *> parseBin)
...
        <|> parseDec
 parseBin :: Stream s m Char => ParsecT s u m Integer
-parseBin = readBin <$> (many1 $ (char '0') <|> (char '1'))
+parseBin = readBin . stripUnder <$> (many1 $ (char '0') <|> (char '1') <|> (char '_'))
 parseOct :: Stream s m Char => ParsecT s u m Integer
-parseOct = (fst . head . readOct) <$> many1 octDigit
+parseOct = (fst . head . readOct . stripUnder) <$> many1 (octDigit <|> (char '_'))
 parseDec :: Stream s m Char => ParsecT s u m Integer
-parseDec = (fst . head . readDec) <$> many1 digit
+parseDec = (fst . head . readDec . stripUnder) <$> many1 (digit <|> (char '_'))
```
ryzhyk: This does not seem right. I don't think we should allow leading underscores, as the same literal, e.g., _0 can then be interpreted as a number of a valid identifier...

PR #108, opened 2018-09-27
Some optimizations derived from the Souffle test case

ryzhyk:


rust/template/main.rs
```diff
@@ -31,6 +31,10 @@ use differential_datalog::program::*;
 use cmd_parser::*;
 use time::precise_time_ns;
 
+// uncomment to enable profiling
```
mbudiu-vmw: Can't you do this with a dynamic boolean check?

PR #109, opened 2018-10-11
OVN integration

ryzhyk:
Progress on OVN integration-related features:

1. A more direct OVSDB schema encoding that stores sets and maps as atomic values in DDlog records (as opposed to generating separate tables for them). I also finally implemented aggregates in DDlog, so one can easily flatten a map into a set of records, compute on those records and then aggregate them back into a set.

2. Support for input relations with primary keys. This allows updating or deleting a record only knowing its primary key and not the complete value (which is how OVSDB updates/deletions work).

3. Generate Delta and Realized tables in the OVSDB frontend. No rules to compute those yet.


src/Language/DifferentialDatalog/OVSDB/Compile.hs
```diff
@@ -21,57 +49,103 @@ import Language.DifferentialDatalog.PP
...
+                 else empty
+    return $ prefix <+> "relation" <+> pp tname <+> "("            $$
+             (nest' $ vcat $ punctuate comma $ uuidcol ++ columns) $$
+             ")"                                                   $$
+             key
+
+mkCol :: (MonadError String me) => OVSDBSchema -> TableKind -> String -> TableColumn -> me Doc
+mkCol schema tkind tname c@TableColumn{..} = do
+    check (columnName /= "_uuid") (pos c) $ "Reserved column name _uuid in table " ++ tname
+    check (columnName /= "uuid-name") (pos c) $ "Reserved column name uuid-name in table " ++ tname
```
mbudiu-vmw: is this minus or underscore?

src/Language/DifferentialDatalog/OVSDB/Compile.hs
```diff
@@ -102,23 +176,44 @@ mkComplexType isinput tname TableColumn{..} t@ComplexType{..} = do
...
+                  Nothing -> return $ "Set<" <> key <> ">"
+                  Just v  -> do vt <- mkBaseType tkind v
+                                return $ "Map<" <> key <> "," <> vt <> ">"
+
+mkBaseType :: (MonadError String me) => TableKind -> BaseType -> me Doc
+mkBaseType _     (BaseTypeSimple at)   = mkAtomicType at
+mkBaseType tkind (BaseTypeComplex cbt) | isJust (refTableBaseType cbt) && tkind == TableOutput
+                                       = return "string"
+mkBaseType tkind (BaseTypeComplex cbt) | isJust (refTableBaseType cbt) && elem tkind [TableDeltaPlus, TableDeltaUpdate]
+                                       = return "uuid_or_string_t"
```
mbudiu-vmw: did you reserve this identifier?

src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -272,18 +278,41 @@ ruleRHSValidate d rl@Rule{..} (RHSFlatMap v e) idx = do
     exprValidate d [] ctx e
     case exprType' d ctx e of
          TOpaque _ tname [_] | elem tname sET_TYPES -> return ()
-         t  -> err (pos e) $ "FlatMap expression must be of these types: " ++ intercalate ", " sET_TYPES ++ ", but its type is " ++ show t
+         TOpaque _ "std.Map" [_,_]                  -> return ()
+         t  -> err (pos e) $ "FlatMap expression must have one of these types: " ++ intercalate ", " sET_TYPES ++ ", or std.Map but its type is " ++ show t
```
mbudiu-vmw: Does this take care of the case where you typedefed one of these standard types and are using the new name?

doc/language_reference/language_reference.md
```diff
@@ -595,14 +605,26 @@ extern function extract_ips(addrs: string): Set<ip_addr_t>
 ```
 
 The sixth form groups records computed so far by a subset of fields,
-computes an aggreagate for each group using specified aggregate function
-(*requirements on aggregate function signature to be specified*), and
-binds the result to a new variable:
+computes an aggreagate for each group using specified aggregate function,
```
mbudiu-vmw: I wonder whether we should not call it GroupByAndAggregate. It is more in line with its semantics, although it's a bit long.

test/datalog_tests/simple.dump.expected
```diff
@@ -155,3 +155,34 @@ Rel2:
 Rel2{0,Option1{0,IP4{300},(true, "foo")}}
 Rel3:
 Rel3{0,IP4{100},Option1{0,IP4{300},(true, "foo")}}
+Aggregate1
+Aggregate1{"a",3}
+Aggregate1{"b",3}
+Aggregate2
+Aggregate2{"a",std_Set { x: {"1", "2", "3"} }}
```
mbudiu-vmw: Can you also parse sets and vectors in an input file?

test/ovn/OVN_Northbound.dl.expected
```diff
@@ -1,132 +1,69 @@
 import types
 
-input relation NB_Global(
+input relation NB_Global (
```
mbudiu-vmw: this is very nice.

test/ovn/OVN_Northbound.dl.expected
```diff
@@ -1,132 +1,69 @@
...
     _uuid: uuid,
     nb_cfg: integer,
     sb_cfg: integer,
     hv_cfg: integer,
+    external_ids: Map<string,string>,
```
mbudiu-vmw: I am wondering whether this will work for nested maps: `Map<string, Map<string, string>>`.

test/ovn/ovn.dl
```diff
@@ -1,2 +1,17 @@
 import OVN_Northbound
 import OVN_Southbound
+
+
+// FlatMap test
+input relation HostAddresses(host: bit<64>, addrs: Set<string>)
+relation HostAddress(host: bit<64>, addr: string)
+
+HostAddress(host, addr) :- HostAddresses(host, addrs),
+                           FlatMap(addr=addrs).
```
mbudiu-vmw: I also find this syntax a little jarring. I would prefer something like FlatMap(addr in addrs).

doc/language_reference/language_reference.md
```diff
@@ -242,9 +243,18 @@ typedef R = R{f1: bigint, f2:bool}
 relation R[R]
 ```
 
+The optional primary key clause is only allowed in `input` relations and defines a mapping from a record to
+its unique key. For example, the following declares a relation with primary key `f1`:
+
+```
+relation R(f1: bigint, f2: bool) 
+primary key (r) r.f1
```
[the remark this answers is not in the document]
ryzhyk: Relations don't really have fields. A relation can have any type, including scalar type without fields like `bool`, a tuple that also does not have named fields or a variant type with multiple constructors. And even in the common case where relation is just a struct with fields, it might be necessary to access the record as a whole, e.g., to pass it as an argument to a function. This is why I needed a way to refer to the entire record by name. This is essentially a lambda, except that we don't support lambdas right now.

Here are some options:
- we can come up with a shortcut syntax for the common case when the primary key is just a subset of fields, e.g., `primary key (f1, f2)`
- we can come up with a more general syntax for lambda expressions and use it here, so that we don't need to change the primary key syntax when we add support for lambdas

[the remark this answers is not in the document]
mbudiu-vmw: For now you can leave it as it is, but we should do both these things.
For lambdas I would suggest `(r) -> r.f1`. Shouldn't be much harder to parse.

src/Language/DifferentialDatalog/OVSDB/Compile.hs
ryzhyk: Actually, both should be illegal. Thanks for catching this!

src/Language/DifferentialDatalog/OVSDB/Compile.hs
ryzhyk: No need, OVSDB does not have user-defined types, so there is no risk of collision here.

src/Language/DifferentialDatalog/Validate.hs
ryzhyk: These are declared in `stdlib.dl` that is automatically imported by all other modules. DDlog will complain if the user attempts to re-define these types.

test/datalog_tests/simple.dump.expected
ryzhyk: yep, e.g.,

```

start;
insert HostAddresses(0, ["10.10.10.101", "10.10.10.102", "10.10.10.103"]),
insert HostAddresses(0, ["10.10.10.103", "10.10.10.104", "10.10.10.105"]),
insert HostAddresses(1, ["192.168.0.1", "192.168.0.2", "192.168.0.3"]);
commit;
```

test/ovn/ovn.dl
ryzhyk: makes sense.

src/Language/DifferentialDatalog/Validate.hs
mbudiu-vmw: The question is the other way round:
`typedef X Set<string, string>`
And you are using X.

ryzhyk: Yes, the `exprType'` method will expand type definitions.

PR #117, opened 2018-11-04
minor improvements in souffle converter 

remysucre:
Handles new syntax for output relation declarations and remove hard-coded file names in soufflé to DDlog converter


mbudiu-vmw:
You should use git rebase to keep the commit logs cleaner

PR #129, opened 2018-12-17
Ref<T> type and convenience syntax around it

ryzhyk:
- `Ref<T>` is the DDlog shortcut to Rust's `Arc<T>` type
- syntax for declaring relations whose records are references:
 ```
 relation &Router(
     lr:                 nb.Logical_Router,
     dpname:             string,
     l3dgw_port:         Option<nb.Logical_Router_Port>,
     redirect_port_name: string,
     is_gateway:         bool,
     snat_external_ips:  Set<string>
 )
 ```
- Pattern matching relations and structs wrapped in a `Ref`
```
&RouterPort(.lrp = lrp,
                 .router = &Router{.lr = lr, .dpname = dpname, .snat_external_ips = external_ips},
                 .json_name = json_name,
                 .networks = networks)
```
- syntax in flux; no documentation yet

### DIFF
--- a/src/Language/DifferentialDatalog/Validate.hs
+++ b/src/Language/DifferentialDatalog/Validate.hs
@@ -365,7 +365,7 @@ exprValidate d tvars ctx e = {-trace ("exprValidate " ++ show e ++ " in \n" ++ s
 -- This function does not perform type checking: just checks that all functions and
 -- variables are defined; the number of arguments matches declarations, etc.
 exprValidate1 :: (MonadError String me) => DatalogProgram -> [String] -> ECtx -> ExprNode Expr -> me ()
-exprValidate1 d _ ctx e@EVar{..} | ctxInRuleRHSPattern ctx
+exprValidate1 d _ ctx e@EVar{..} | ctxInRuleRHSPositivePattern ctx
                                           = do
     when (ctxInBinding ctx) $
         check (isJust $ lookupVar d ctx exprVar) (pos e) $ "Variable declarations not allowed inside @-bindings"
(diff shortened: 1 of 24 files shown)


src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -414,8 +411,13 @@ exprValidate1 _ _ ctx (EPHolder p)        = do
                    CtxStruct EStruct{..} _ f -> "Missing field " ++ f ++ " in constructor " ++ exprConstructor
                    _               -> "_ is not allowed in this context"
     check (ctxPHolderAllowed ctx) p msg
-exprValidate1 d _ ctx (EBinding p v _)    = checkNoVar p d ctx v
+exprValidate1 d _ ctx (EBinding p v _)    = do
+    checkNoVar p d ctx v
+    
 exprValidate1 d tvs _ (ETyped _ _ t)      = typeValidate d tvs t
+exprValidate1 _ _ ctx (ERef p _)          =
+    -- Rust does not allow pattern matching inside 'Arc'
```
mbudiu-vmw: this is a strange comment. what should matter is what ddlog allows, not what rust allows.

test/datalog_tests/simple.dl
```diff
@@ -681,3 +681,24 @@ Allocation(name, id, x) :-
...
+input relation Referenced(x: bool, y: Option<string>)
+
+output relation Referee(r: Ref<Referenced>)
+
+Referee(ref_new(r)) :-
+    r in Referenced(_, Some{_}).
+
+output relation Filtered(r: Ref<Referenced>)
+
+Filtered(r.r) :- r in Referee(.r = &Referenced{.x = true}).
```
mbudiu-vmw: Is equality for references well-defined? I suspect they behave like pointers. But this means you cannot just copy them.

mbudiu-vmw: I.e., if you delete a value and insert it back again you will have the same value in the relation, but the other relation will have a different reference. I will have to think about the semantics of this.

src/Language/DifferentialDatalog/Validate.hs
ryzhyk: Sure, but it is really hard to implement things that don't directly map to Rust. The comment just documents the reason for the limitation.

test/datalog_tests/simple.dl
ryzhyk: The current semantics compares values inside references. In the scenario you described, there should still only be one reference, but there theoretically exist more complex dataflows where this is indeed possible. In any case, multiple references should not affect the functionality; only memory footprint.

PR #135, opened 2019-01-07
Several improvements based on OVN work

ryzhyk:
Several improvements based on OVN work:

- allow DDlog libraries to add new Rust dependencies
- `min` and `max` stdlib functions
- improved allocator API
- `deref()` function
- fix the antijoin implementation

### DIFF
--- a/lib/allocate.dl
+++ b/lib/allocate.dl
@@ -1,19 +1,21 @@
-// Allocate a unique 64-bit identifier in the range `1..max_val` for each element in `toallocate`.
+// Allocate a unique identifier in the range `min_val..max_val` for each element in `toallocate`.
 // `allocated` stores already allocated ids that must not be re-used.
 //
 // Returns new allocations.  May only assign a subset of values if the address space has been
 // exhausted.
-extern function allocate_u64(
-    allocated: Set<bit<64>>,
+extern function allocate(
+    allocated: Set<'N>,
     toallocate: Vec<'B>,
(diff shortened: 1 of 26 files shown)


lib/allocate.dl
```diff
@@ -1,19 +1,21 @@
-// Allocate a unique 64-bit identifier in the range `1..max_val` for each element in `toallocate`.
+// Allocate a unique identifier in the range `min_val..max_val` for each element in `toallocate`.
```
mbudiu-vmw: there is some constraint on the type 'N. What is it?

lib/allocate.rs
```diff
@@ -1,39 +1,78 @@
...
     for b in toallocate.x.iter() {
         loop {
-            last = if last + 1 <= *max_val {last + 1} else {1};
-            if last == first {
-                return res;
-            };
-            if allocated.x.contains(&last) { continue; };
-            res.x.push(((*b).clone(), last));
-            break;
+            next = if next == *max_val { *min_val } else { next + N::one() };
```
mbudiu-vmw: won't this get into an infinite loop if all values are allocated?

lib/allocate.dl
ryzhyk: ack

lib/allocate.rs
ryzhyk: It uses `offset` to terminate the loop, so should be fine.

PR #138, opened 2019-01-12
support for record modification commands

ryzhyk:
This set of changes introduces support for record modification commands that
change an existing record in the database.  This is needed, e.g., to
support the OVSDB update-2 format.  A modification command only
contains values of modified columns; therefore in order to decompose it
into a delete+insert sequence, we must fetch existing value from the DB,
and apply update to it to compute a new value.

To enables this, multiple changes, mostly to the Rust template were
required:
- add methods to `RunningProgram` to read current input relation state
- `Mutator` trait for objects that can modify other objects
- implement `Mutator` trait for `Record` and all Rust types that can be
used in a datalog program, including built-in, library, and user-defined
types
- introduce `Modify` command to `Record` and `RunningProgram` APIs
- parse update-2 OVSDB records

In addition, this patch introduces the notion of default values.
Previously, missing fields in a `Record::NamedStruct` were treated as
errors.  We now use default values for missing columns instead.  This
allows reducing the communication overhead and is also required to
support update-2 format that relies on this behavior.


lib/std.rs
```diff
@@ -186,6 +193,13 @@ impl<T: IntoRecord + Ord> IntoRecord for std_Set<T> {
     }
 }
 
+impl<T: FromRecord + Ord> Mutator<std_Set<T>> for Record
+
```
mbudiu-vmw: empty line

rust/template/differential_datalog/record.rs
```diff
@@ -477,21 +481,19 @@ impl FromRecord for u8 {
...
+    }
+}
+
 impl IntoRecord for u8 {
     fn into_record(self) -> Record {
         Record::Int(BigInt::from(self))
     }
 }
 
-#[test]
```
mbudiu-vmw: why delete the tests?

rust/template/differential_datalog/record.rs
```diff
@@ -917,50 +925,36 @@ impl<T: IntoRecord+Ord> IntoRecord for BTreeSet<T> {
     }
 }
 
-
-#[test]
-fn test_vec() {
-    assert_eq!(<vec::Vec<bool>>::from_record(&Record::Array(CollectionKind::Unknown, vec![Record::Bool(true), Record::Bool(false)])),
-               Ok(vec![true, false]));
+/* Set update semantics: update contains values that are in one of the maps but not the
```
mbudiu-vmw: one of the sets

rust/template/differential_datalog/record.rs
mbudiu-vmw: found them in a different file

test/datalog_tests/rules.fail.ast.expected
```diff
@@ -6,648 +6,4 @@ expecting letter or digit, "_", ".", "(" or "["
   ./test/datalog_tests/rules.fail.dl:5:1-7:1
   ./test/datalog_tests/rules.fail.dl:7:1-9:1
 
-(typedef R1 = R1{a1: string}
```
mbudiu-vmw: this is a big change... can't figure out why

ryzhyk: This is something I missed a long time ago. This was meant to test that ddlog rejects non-linear rules. However, we now allow such rules. So I moved the test from `simple.fail.dl` to `simple.dl`, since it does not fail anymore.

PR #148, opened 2019-02-06
Generate binary ddlog distribution using travis

ryzhyk:
- Statically link ddlog executable on Linux to eliminate shared library dependencies
- Configure travis to package ddlog executable and libraries and upload them as github releases on each tagged commit.


ryzhyk:
> The build has 1/2 failed.

Yeah, I canceled a redundant build.

PR #149, opened 2019-02-08
Optimized antijoin implementation

ryzhyk:
Previously, the antijoin operator created three arrangements of the
collection being antijoined with: two to apply `distinct()` to the
collection, and one more to arrange the resulting collection by self.

This PR reduces this to 1 or 2 arrangements, potentially shared
across multiple operators.


mbudiu-vmw:
I can't say I understand all the Haskell code, but the comments make some sense.
I wonder whether in the end this could become a PLDI paper that also talks about Frank's arrangements. I expect his SIGMOD paper didn't make it. I don't know the literature on compiling datalog, maybe this is obvious.

PR #154, opened 2019-02-28
Optimization: share the first arrangement in the body of a rule

ryzhyk:
This is an important optimization. Consider a rule of the form

```
A(x) :- B(x,y), C(y,z).
```

To evaluate this rule, we arrange relations `C` and `B` by field `y`. However, while the former arrangement was shared by all rules that require this arrangement, the latter (i.e., the arrangement of the first literal in the rule) was computed per rule.  This PR enables sharing of this arrangement when possible.


ryzhyk:
> Does it help?

Arrangement reuse is definitely happening, but I don't have good benchmarks to test it.

Issue #153, opened 2019-02-27
Java API improvements


ryzhyk:
- [x] Add a flag to store output tables in DDlog
- [x] Add a dump table callback
- [ ] Profile the runtime of callbacks (not Java-specific)
- [x] Transform DDlog errors into Java exceptions
- [x] Batch records per thread (don't invoke callback for each record)

PR #155, opened 2019-03-02
Expose flag to store relations in API; better documentation

mbudiu-vmw:
Addresses part of #153


mbudiu-vmw:
The build failed with a timeout. I don't know why it took so long.

ryzhyk:
> The build failed with a timeout. I don't know why it took so long.

I need to fix the Travis script. That test case combines two slow span tests and I think Travis became slower recently.

PR #161, opened 2019-03-27
Java example with nested sets for redist.dl

mbudiu-vmw:


mbudiu-vmw:
You will probably need to update some dump outputs.

mbudiu-vmw:
Yes, I was using the sorted map just for testing.
I will also convert the outer map to a fast map.
Are you getting the correct results?

ryzhyk:
> Are you getting the correct results?

After 034fa980b8c3bae2ca442cabf78cc313074ed235 - yes (with TreeMap).

mbudiu-vmw:
I have pushed a new change which has two implementations: one for testing with a RBTree and one for production with OpenHashMap. Unfortunately the code is a bit ugly now. In production you can delete much of the code. The OpenHashMap is about 2x faster than the RBTree.

PR #162, opened 2019-03-28
Two new library functions

ryzhyk:
- `std.string_slice()`
- `allocate.allocate_with_hint()`

### DIFF
--- a/lib/std.dl
+++ b/lib/std.dl
@@ -58,6 +58,10 @@ extern function string_join(strings: Vec<string>, sep: string): string
 extern function string_split(s: string, sep: string): Vec<string>
 extern function string_contains(s1: string, s2: string): bool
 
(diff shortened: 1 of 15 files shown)


lib/std.dl
```diff
@@ -58,6 +58,10 @@ extern function string_join(strings: Vec<string>, sep: string): string
 extern function string_split(s: string, sep: string): Vec<string>
 extern function string_contains(s1: string, s2: string): bool
 
+/* Substring of `s` that begins at `start` and continues up to, but not
+ * including, `end`.*/
+extern function string_slice(s: string, start: bit<64>, end: bit<64>): string
```
mbudiu-vmw: slice in python, javascript, etc. means something else. You should really call this substring or substr.
Also, why 64-bit indexes?

PR #164, opened 2019-04-04
String interner

ryzhyk:
See commit comment

### DIFF
--- a/lib/intern.dl
+++ b/lib/intern.dl
@@ -1,0 +1,21 @@
+/* Object interning library.
+ * Generates a unique ID for an object, which can be converted back to the original value.
+ */
+
+/* Interned object of type `'A`.
+ * While this type is defined for any `'A`, interning is only supported for strings.
+ * There is simply no way to obtain an interned object of a different type.
+ */
+extern type IObj<'A>
+
+/* Interned string
+ */
+typedef IString = IObj<string>
+
+/* Intern a string
+ */
+extern function string_intern(s: string): IString
+
+/* Extract the value of an interned string
+ */
+extern function istring_str(s: IString): string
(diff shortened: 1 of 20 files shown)


lib/intern.rs
```diff
@@ -0,0 +1,115 @@
...
+type Symbol = u32;
+
+#[derive(Eq, PartialOrd, PartialEq, Ord, Clone, Hash)]
+pub struct intern_IObj<T> {
+    pub x: Symbol,
+    phantom: marker::PhantomData<T>
+}
+
+struct Interner<T> {
+    vec: vec::Vec<T>,
```
mbudiu-vmw: why do you need the vector?
isn't a Map<T, T> enough?

lib/intern.dl
```diff
@@ -0,0 +1,21 @@
...
+ */
+extern type IObj<'A>
+
+/* Interned string
+ */
+typedef IString = IObj<string>
+
+/* Intern a string
+ */
+extern function string_intern(s: string): IString
```
mbudiu-vmw: I find these names rather baroque. Why not intern and get_string?

ryzhyk: I try to follow the "object oriented" naming convention: function names start with the name of the type the function applies to followed by a verb. If I call this just `intern`, what should I call similar methods for other (non-string) types?   Ideally, `intern()` should be polymorphic, but I haven't figured out a way to do it, since the interner is a global variable and we need one per every type that we want to intern.

Open to better suggestions, though.

lib/intern.rs
ryzhyk: Sorry, I don't understand how `Map<T,T>` is enough.  We need `Map<T, Symbol>` to go from `T` to number and `Vec<T>` to go from number to `T`.

PR #168, opened 2019-04-08
Fix escaping of Unicode characters that would break parsing.

gfour:
In the Souffle converter, Unicode characters in strings were being escaped as '\uXXXX' sequences in ASCII. This would break the parser during `insert`. This PR keeps the original Unicode characters in place, so that parsing succeeds.


ryzhyk:
@gfour, thanks for contributing the fix!

PR #179, opened 2019-04-24
Use output statements to generate .dump.expected file

mbudiu-vmw:
The access1 test is taken from the souffle tests. The plan is to remove it in the future and to automatically download all tests when running tests. But I have submitted it so we can test the new functionality before we have the download code.

Also TBD is the concatenation of all tests into a single big one.


mbudiu-vmw:
I will close this PR, and I will submit a new one more complete.

PR #184, opened 2019-05-01
Record and replay feature for DDlog

ryzhyk:
It can be hard to debug and profile DDlog when it's running as a library
in a bigger program.  For instance, there is typically no easy way to modify
inputs to DDlog.

This commit adds the ability to record DDlog commands in a dat file that
can later be replayed against the standalone DDlog executable (`prog_cli`).
To enable this feature, the client calls the new `ddlog_record_commands`
method, supplying an open file descriptor as argument.  DDlog will write
command stream to this file.

This commit also introduces command syntax for the `modify` operation,
which is necessary to ensure that all commands supported via the API
also have command-line equivalents.

Finally, two new commands (and matching API functions) were introduced:
`profile cpu on`, `profile cpu off` enable and disable CPU profiling
respectively.  It turns out that intercepting all operator activations
is expensive in large dataflows, so it makes sense to only do it for
selected program sections and only when necessary.

### DIFF
--- a/rust/template/lib.rs
+++ b/rust/template/lib.rs
@@ -41,7 +41,12 @@ use std::borrow;
 use std::ptr;
 use std::ffi;
 use std::boxed;
+use std::fs;
 use std::ops::Deref;
+use std::io::Write;
+use std::os::unix;
+use std::os::unix::io::{IntoRawFd, FromRawFd};
+use std::mem;
 use libc::size_t;
 
 pub mod valmap;
@@ -50,7 +55,10 @@ pub mod ovsdb;
 pub struct HDDlog {
     prog: sync::Mutex<RunningProgram<Value>>,
     db: Option<sync::Arc<sync::Mutex<valmap::ValMap>>>,
-    print_err: Option<extern "C" fn(msg: *const raw::c_char)>
+    print_err: Option<extern "C" fn(msg: *const raw::c_char)>,
+    /* When set, all commands sent to the program are recorded in
+     * the specified `.dat` file so that they can be replayed later. */
+    replay_file: Option<sync::Mutex<fs::File>>
 }
 
 impl HDDlog {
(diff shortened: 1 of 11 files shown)


rust/template/lib.rs
```diff
@@ -172,43 +189,81 @@ pub extern "C" fn ddlog_run(
...
+    if prog.is_null() {
+        return -1;
+    };
+    let mut prog = sync::Arc::from_raw(prog);
+    let res = match sync::Arc::get_mut(&mut prog) {
+        Some(prog) => {
+            /* Swap out the old file and convert it into FD to prevent
+             * the file from closing on destruction (it is the caller's
+             * responsibility to close the file when they're done with it). */
+            let mut old_file = if fd == 0 {
```
mbudiu-vmw: 0 is a legal file descriptor in unix; it's stdout.

rust/template/lib.rs
```diff
@@ -249,13 +318,24 @@ pub unsafe extern "C" fn ddlog_transaction_commit(prog: *const HDDlog) -> raw::c
     res
 }
 
+unsafe fn record_transaction_commit(prog: &sync::Arc<HDDlog>) {
+    if let Some(ref f) = prog.replay_file {
+        if write!(f.lock().unwrap(), "commit;\n").is_err() {
+            prog.eprintln("ddlog_transaction_commit(): failed to record invocation in replay file");
```
mbudiu-vmw: your error messages are copy-paste; do you want a different message for each failure?

rust/template/lib.rs
ryzhyk: Good point. Let's use -1 instead.

rust/template/lib.rs
ryzhyk: They are not exactly the same; each carries the name of the function, which uniquely identifies its location.

Issue #165, opened 2019-04-04
Size annotations for `extern` types


ryzhyk:
At the moment DDlog does not know the size of types declared as `extern` and has to conservatively assume that they are large and place them in a `Box`.  Often times, extern types are however small, e.g.,  a fat pointer (like `Ref<>`) or an integer (like `IObj<>`).

Solution: allow users to annotate `extern type` declarations with size. Note, an incorrect annotation does not affect correctness, but may affect memory footprint: failing to box a large type will blow up the size of the `Value`.

PR #187, opened 2019-05-02
Support for meta-attributes

ryzhyk:
Meta-attributes annotate top-level DDlog declarations (types, relations,
functions, and rules).  They do not change the semantics of the program,
but affect compilation process.

Only one attribute is supported at the moment, namely the `size` attribute,
applicable to `extern type` declarations.  It specifies the size of the
corresponding Rust data type in bytes:

```
 #[size=4]
 extern type IObj<'A>
```

The compiler uses this attribute to more accurately compute data
structure sizes to avoid boxing small structs.

This reduces the runtime of the souffle benchmark from 11 to 7 seconds
(see #165)

### DIFF
--- a/src/Language/DifferentialDatalog/Syntax.hs
+++ b/src/Language/DifferentialDatalog/Syntax.hs
@@ -33,6 +33,8 @@ SOFTWARE.
 -- WithPos: manipulating position information
 
 module Language.DifferentialDatalog.Syntax (
+        Attribute(..),
+        ppAttributes,
         Type(..),
         typeTypeVars,
         tBool,
@@ -49,6 +51,8 @@ module Language.DifferentialDatalog.Syntax (
         structFieldGuarded,
         Field(..),
         TypeDef(..),
+        tdefCheckSizeAttr,
+        tdefGetSizeAttr,
         Constructor(..),
         consType,
         consIsUnique,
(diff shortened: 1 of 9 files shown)


src/Language/DifferentialDatalog/Syntax.hs
```diff
@@ -289,21 +323,37 @@ instance WithName TypeDef where
...
 
 instance Show TypeDef where
     show = render . pp
 
 instance Eq TypeDef where
-    (==) t1 t2 = name t1 == name t2 && tdefType t1 == tdefType t2
+    (==) t1 t2 = (name t1, tdefAttrs t1, tdefAttrs t1, tdefType t1) == (name t2, tdefAttrs t2, tdefAttrs t2, tdefType t2)
+
+tdefCheckSizeAttr :: (MonadError String me) => TypeDef -> me (Maybe Int)
+tdefCheckSizeAttr TypeDef{..} =
```
mbudiu-vmw: if i understand this right the size attribute will be ignored on types which are not extern

mbudiu-vmw: moreover, it won't be checked.

ryzhyk: `Validate.hs` has code to check that only extern types can have this attribute.  This function only checks that the attribute has correct format.

ryzhyk:
@Argc0, @gfour, with this PR you should see significant speedup on Souffle benchmarks.

Argc0:
@ryzhyk Yes, we see a speedup from 4x it gone to 3x, compared to original times of souffle (in our examples).

ryzhyk:
Thanks for the update, this is great!

PR #196, opened 2019-05-03
Three functions needed by the souffle converter

### LINKED ISSUE #190: The re_match function in souffle_lib.dl is not implemented
This function needs a native Rust implementation; it checks whether a string matches a regular expression.

ryzhyk:
Two new functions in the std library:
```
extern function pow32(base: 'A, exp: bit<32>): 'A`
extern function group_sum(g: Group<'A>): 'A
```

New `lib/regex` library, which currently contains a single function.  Note that this is based on the Rust regex syntax (https://docs.rs/regex/1.1.6/regex), which may differ from what souffle expects:
```
extern function regex_match(regex: string, str: string): bool
```

This should resolve #190


lib/std.dl
```diff
@@ -14,6 +14,11 @@ extern function deref(x: Ref<'A>): 'A
 extern function max(x: 'A, y: 'A): 'A
 extern function min(x: 'A, y: 'A): 'A
 
+/*
+ * Arithmetic functions
+ */
+extern function pow32(base:'A, exp: bit<32>): 'A
```
mbudiu-vmw: do you really expect we'll have other versions of pow?
Why don't you just call it pow?

ryzhyk: 32 bits is sufficient, but because DDlog is strict with types and doesn't have type casting right now, it may be convenient to pass other integer types without conversion. Also, I cannot easily make this function polymorphic. So lets keep the name.

PR #204, opened 2019-05-08
Support for user-defined aggregates

ryzhyk:
This addresses an important limitation: DDlog used to only support
aggregation functions written in Rust, which would have been a major
roadblock for most users.

With this commit, the `Aggregate` operator allows arbitrary user-defined
functions with the following signature:

```
function f(Group<X>): Y
```

where `X` and `Y` are arbitrary types.  `Group` is an `extern` type,
that can be manipulated using three functions declared in `std.dl`:

```
extern function group_count(g: Group<'A>): bit<64>
extern function group_first(g: Group<'A>): 'A
extern function group_nth(g: Group<'A>, n: bit<64>): Option<'A>
```

In addition, `Group` can be iterated over in a for-loop.

### DIFF
--- a/lib/std.dl
+++ b/lib/std.dl
@@ -77,28 +77,49 @@ extern function str_to_lower(s: string): string
 extern function hash64(x: 'X): bit<64>
 extern function hash128(x: 'X): bit<128>
 
+/* The `Group` type is used exclusively in aggregation operations.  It
+ * represents a non-empty list of objects sorted in ascending order.
+ */ 
+extern type Group<'A>
+
 /*
  * Standard aggregates
  */
 
-extern type Group<'A>
+/* Returns the number of elements in the group, which is guaranteed
+ * to be >0. */
+extern function group_count(g: Group<'A>): bit<64>
 
-extern function count(g: Group<'A>): bit<64>
+/* Returns the first element of the group.
+ * It always exists, as aggregation cannot return an empty group. */
+extern function group_first(g: Group<'A>): 'A
+
+/* Returns `n`th element of the group if the group is large enough
+ * or `None` otherwise.
+ */
+extern function group_nth(g: Group<'A>, n: bit<64>): Option<'A>
+
 extern function group2set(g: Group<'A>): Set<'A>
 extern function group2vec(g: Group<'A>): Vec<'A>
 extern function group2map(g: Group<('K,'V)>): Map<'K,'V>
-extern function group_unzip(g: Group<('X,'Y)>): (Vec<'X>, Vec<'Y>)
 extern function group_sum(g: Group<'A>): 'A
+
+function group_unzip(g: Group<('X,'Y)>): (Vec<'X>, Vec<'Y>) = {
+    var xs: Vec<'X> = vec_empty();
+    var ys: Vec<'Y> = vec_empty();
+    for (v in g) {
+        (var x, var y) = v;
+        vec_push(xs, x);
+        vec_push(ys, y)
+    };
+    (xs,ys)
+}
 
 extern function group_set_unions(g: Group<Set<'A>>): Set<'A>
 
 /* Optimized version of group_set_unions that, when the group consits of
  * a single set, simply returns the reference to this set. */
 extern function group_setref_unions(g: Group<Ref<Set<'A>>>): Ref<Set<'A>>
-
-/* Returns the first element of the group.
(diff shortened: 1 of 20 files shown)


lib/std.dl
```diff
@@ -77,29 +77,50 @@ extern function str_to_lower(s: string): string
 extern function hash64(x: 'X): bit<64>
 extern function hash128(x: 'X): bit<128>
 
+/* The `Group` type is used exclusively in aggregation operations.  It
+ * represents a non-empty list of objects sorted in ascending order.
```
mbudiu-vmw: Do you want to enforce the fact that the group is sorted?
Maybe you will want to change your mind later.
At least in LINQ IGroup is not sorted.

lib/std.dl
```diff
@@ -77,29 +77,50 @@ extern function str_to_lower(s: string): string
...
+ */ 
+extern type Group<'A>
+
 /*
  * Standard aggregates
  */
 
-extern type Group<'A>
+/* Returns the number of elements in the group, which is guaranteed
+ * to be >0. */
```
mbudiu-vmw: Here you are assuming that groups will be only produced by aggregation.
However, if you allow functions on groups perhaps they can be produced in some other ways as well.
So I don't think you should promise that a group cannot be empty.

lib/std.dl
```diff
@@ -77,29 +77,50 @@ extern function str_to_lower(s: string): string
...
+extern function group_nth(g: Group<'A>, n: bit<64>): Option<'A>
 
-extern function count(g: Group<'A>): bit<64>
 extern function group2set(g: Group<'A>): Set<'A>
 extern function group2vec(g: Group<'A>): Vec<'A>
 extern function group2map(g: Group<('K,'V)>): Map<'K,'V>
-extern function group_unzip(g: Group<('X,'Y)>): (Vec<'X>, Vec<'Y>)
 extern function group_sum(g: Group<'A>): 'A
 
+function group_unzip(g: Group<('X,'Y)>): (Vec<'X>, Vec<'Y>) = {
```
mbudiu-vmw: This would make sense for other datatypes that can be iterated upon. But do you want this to be a primitive? The generalization would take functions from elements to values instead of projecting on components.

lib/std.rs
```diff
@@ -302,6 +364,36 @@ pub struct std_Map<K: Ord,V> {
     pub x: BTreeMap<K,V>
 }
 
+/* This is needed so we can support for-loops over `Set`'s
```
mbudiu-vmw: over Maps

lib/std.rs
```diff
@@ -469,101 +556,161 @@ pub fn std_hash128<T: Hash>(x: &T) -> u128 {
...
+
+    fn size(&self) -> u64 {
+        self.group.len() as u64
+    }
+
+    fn first(&'a self) -> &'a X {
+        X::from_value(self.group[0].0)
+    }
+
+    fn nth_unchecked(&'a self, n: u64) -> &'a X {
```
mbudiu-vmw: if you decide to allow empty groups this should return None for empty groups.

lib/std.dl
ryzhyk: The sortedness is enforced by differential, not me. This may theoreticlly change in the future, breaking people's code. OTOH, an aggregation algorithm that requires a sorted group may be able to take advantage of this.

lib/std.dl
ryzhyk: Groups are immutable and can only be produced by the `Aggregate` operator, so there is no way for the programmer to forge an empty group.  In fact, I should not even allow copying groups or returning them from functions, as this will generate code that will fail to compile in Rust, but I don't have a clean infrastructure to do dataflow analysis atm.

lib/std.dl
ryzhyk: Unfortunately, we do not support lambdas or even function-typed variables at the moment.  This is only a primitive in that it is declared in the `std` lib; it is not a primitive in that it is actually implemented in DDlog.  Is your point that it is kind of arbitrary and should not be in the standard lib? I think I only use it in one of the projects at the moment and could just move this function there. But on the other hand, I don't see any harm in keeping it in `std`.  In fact, even Haskell, which support lambdas, provides unzip in the standard library.

lib/std.rs
ryzhyk: fixed

PR #211, opened 2019-05-11
Push initial facts one-by-one

ryzhyk:
For some weird reason, rustc runs out of memory if we declare 'init_data'
using the 'vec![]' macro (given enough facts), so push them one by one
instead.


ryzhyk:
But now compilation takes forever, so need to keep searching for a better solution...

PR #222, opened 2019-05-20
Multiset semantics for groups.

ryzhyk:
We used to treat groups as sets, e.g., given relation with records

```
|x|y|z|
|-----|
|1|1|1|
|1|2|1|
|1|3|1|
|-----|
```

the following rule fragment

```
R(x,y,z), var a = Aggregate((x), group_count(z))
```

will return the number of all distinct values of `z` for each `x`:

```
|z|a|
|---|
|1|1|
|---|
```

This is counterintuitive and is different from how other Datalog engines
work.  The new semantics treats each group as a multiset and will
produce

```
|z|a|
|---|
|1|3|
|---|
```

in the above example.

Implementation details
----------------------

The multiset semantics is maintained by keeping a tuple of all live
variables in each element in the group and applying a projection
function to it to obtain the argument to the aggregation function.

The good: this design avoids having to apply `distinct` before aggregation and,
furthemore enables arrangement reuse for aggregations (which was not
possible before)

The bad: currently, the projection function always returns an owned value, not
a reference, introducing an extra copy, which can be expensive for large
values.

### DIFF
--- a/src/Language/DifferentialDatalog/Compile.hs
+++ b/src/Language/DifferentialDatalog/Compile.hs
@@ -958,7 +958,7 @@ extractValue d t = parens $
         "|" <> vALUE_VAR <> ": Value| {"                                                              $$
         "match" <+> vALUE_VAR <+> "{"                                                                 $$
         "    Value::" <> mkValConstructorName' d t' <> "(x) => {" <+> boxDeref d t' "x" <+> "},"      $$
-        "    _ => panic!(\"unexpected constructor\")"                                                 $$
+        "    _ => unreachable!()"                                                                     $$
         "}}"
     where t' = typeNormalize d t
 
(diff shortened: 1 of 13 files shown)


src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -866,15 +851,15 @@ mkValType d types grp_types =
     tuple0 = "Value::" <> mkValConstructorName' d (tTuple []) <> (parens $ box d (tTuple []) "()")
     mkdisplay :: Type -> Doc
     mkdisplay t = "Value::" <> consname t <+> "(v) => write!(f, \"{:?}\", *v)"
-    mkfromval t =
+    {-mkfromval t =
```
mbudiu-vmw: do you want to delete this?

PR #228, opened 2019-05-23
Fix another bug in aggregates

mbudiu-vmw:
So far at least 70 Souffle tests pass, but I haven't finished trying all of them yet.


mbudiu-vmw:
this failed because of a timeout

PR #232, opened 2019-05-25
Support for signed numbers

mbudiu-vmw:
These are written as signed<W>.

### DIFF
--- a/src/Language/DifferentialDatalog/Compile.hs
+++ b/src/Language/DifferentialDatalog/Compile.hs
@@ -2054,12 +2054,12 @@ mkType' TBit{..} | typeWidth <= 8  = "u8"
                  | typeWidth <= 64 = "u64"
                  | typeWidth <= 128= "u128"
                  | otherwise       = "Uint"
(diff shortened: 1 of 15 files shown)


src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -2227,9 +2250,11 @@ mkBVMask w | w > 128   = "Uint::parse_bytes(b\"" <> m <> "\", 16)"
 mkTruncate :: Doc -> Type -> Doc
 mkTruncate v t =
     case t of
-         TBit{..} | needsTruncation typeWidth
-                  -> parens $ v <+> "&" <+> mask typeWidth
-         _        -> v
+         TBit{..}    | needsTruncation typeWidth
+                     -> parens $ v <+> "&" <+> mask typeWidth
+         TSigned{..} | needsTruncation typeWidth
```
ryzhyk: `needsTruncation` will always return false for signed ints (since they can only be 8,16,32,64, or 128 bit). If this was not the case, then this truncation logic wouldn't work anyway (as you said, it would have to sign-extend).

PR #234, opened 2019-05-28
Use dummy relation in front of assignement.

Argc0:


test/Spec.hs
```diff
@@ -96,7 +96,7 @@ sOUFFLE_DIRS = ["souffle0", -- large Doop example
                 "souffle1", "souffle2", "souffle3", "souffle4", "souffle5", "souffle6",
                  -- "souffle7", -- uses a recursive type
                  "souffle9", "souffle10", 
-                 "souffle11", "souffle12", "souffle13", "souffle14", "souffle15", "souffle16", "souffle17", "souffle18"]
+                 "souffle11", "souffle12", "souffle13", "souffle14", "souffle15", "souffle16", "souffle17", "souffle18", "souffle20"]
```
mbudiu-vmw: where is souffle19?

tools/souffle_converter.py
```diff
@@ -1067,7 +1067,8 @@ def process_rule(self, rule):
             convertedDisjunctions = [self.convert_terms(x) for x in terms]
             if len(convertedDisjunctions) > 0 and \
                     len(convertedDisjunctions[0]) > 0 and \
-                    convertedDisjunctions[0][0].startswith("not"):
+                    convertedDisjunctions[0][0].startswith("not") or\
```
mbudiu-vmw: We can do this for now, but in general we will need to reorder the clauses to handle many Souffle programs that we cannot compile today.

test/Spec.hs
Argc0: I am afraid that it was already measing, I did not change a thing, I only add souffle20.

mbudiu-vmw: I have added that; will be part of a future PR.

PR #238, opened 2019-05-29
Cast

### LINKED ISSUE #189: There are no casts between integers
One has to use bit manipulation operations.
We may want checked casts.
Also, if we add signed types we may want casts between signed and unsigned.

mbudiu-vmw:
The cast support is not complete, but it contains the grammar, and it works for casts between some integer types. Fixes #189.

### DIFF
--- a/lib/souffle_lib.dl
+++ b/lib/souffle_lib.dl
@@ -2,26 +2,27 @@
 
 import intern
 import regex
+import souffle_types
 
-function castTo32(i: bit<64>): bit<32> = i[31:0]
-function castTo64(i: bit<32>): bit<64> = 32'd0 ++ i
+function castTo32(i: bit<64>): Tnumber = i[31:0] as Tnumber
+function castTo64(i: Tnumber): bit<64> = 32'd0 ++ (i as bit<32>)
 function cat(s: IString, t: IString): IString = (string_intern(istring_str(s) ++ istring_str(t)))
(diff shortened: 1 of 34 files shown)


lib/souffle_lib.dl
```diff
@@ -2,26 +2,27 @@
 
 import intern
 import regex
+import souffle_types
 
-function castTo32(i: bit<64>): bit<32> = i[31:0]
-function castTo64(i: bit<32>): bit<64> = 32'd0 ++ i
+function castTo32(i: bit<64>): Tnumber = i[31:0] as Tnumber
```
ryzhyk: Shouldn't this be just `i as Tnumber`?

lib/souffle_lib.dl
```diff
@@ -2,26 +2,27 @@
...
+function castTo64(i: Tnumber): bit<64> = 32'd0 ++ (i as bit<32>)
```
ryzhyk: Likewise, this can now be written as `i as bit<64>`

mbudiu-vmw: P4 forbids casts that change both the width and the sign, because the semantics is unclear.
I expect we will want to do the same thing.

ryzhyk: So, is it `i as bit<32> as bit<64>`?

mbudiu-vmw: Should give the same result really.

Issue #229, opened 2019-05-23
Single quotes in strings are escaped on output


mbudiu-vmw:
For example, the following string "Luke isn't a good buddy!" in a relation is dumped as follows:
`Rresult{"Luke isn\'t a good buddy!"}`

ryzhyk:
Fixed by 3f2630f1e631ded33f1641958f21609cd6244799

mbudiu-vmw:
This still does not work for interned strings.

PR #240, opened 2019-05-30
Additional fix for single-quote escaping

ryzhyk:
- Single quoutes were still incorrectly escaped in interned strings
- My custom string formatting function had a bug


mbudiu-vmw:
Fixes #229

ryzhyk:
Closing this PR without merging. I will instead integrate it into the type cast PR.

PR #249, opened 2019-06-06
Better document `Aggregate` semantics

ryzhyk:
- Explain how variables go out of scope after `Aggregate`
- Explain grouping by multiple variables

### DIFF
--- a/doc/tutorial/tutorial.md
+++ b/doc/tutorial/tutorial.md
@@ -892,11 +892,27 @@ BestPrice(item, best_price) :-
 
 Here `group_min()` is one of several aggregation functions defined in
 `lib/std.rs`.  What if we wanted to return the name of the vendor along with the
-lowest price for each item?  This requires an aggregation function that takes a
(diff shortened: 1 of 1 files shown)


doc/tutorial/tutorial.md
```diff
@@ -892,12 +892,28 @@ BestPrice(item, best_price) :-
 
 Here `group_min()` is one of several aggregation functions defined in
 `lib/std.rs`.  What if we wanted to return the name of the vendor along with the
-lowest price for each item?  This requires an aggregation function that takes a
+lowest price for each item?  The following naive attempt will not work:
+
+```
+output relation BestVendor(item: string, vendor: string, price: bit<64>)
```
mbudiu-vmw: I think that a diagram would be more helpful than the text explanation.
We can talk about it in person.

ryzhyk: Merging for now.  This is better than nothing. We can add diagram later.

PR #255, opened 2019-06-12
Support non-DNF conversions in Souffle converter

Argc0:
This pull request adds a `--non-dnf` flag to souffle_converter.py, to transform non-DNF rules to DNF rules. The transformation is disabled by default and should only support cases that are found in Doop.


ryzhyk:
@Argc0: thanks!

@mbudiu-vmw, merging this, but it still needs your review when you're available. Plus we need to test if this fixes the souffle test cases that require support for disjunctions.

tools/souffle_converter.py
```diff
@@ -1046,14 +1047,87 @@ def convert_terms(self, terms):
                 result.append(term)
         return result + tail
 
+    def flatten(self, tail, listbody, bool_dis):
+        for x in tail.children:
+            if (x.symbol.name == "Term"):
+                if (getOptField(x, "Literal") != None) or (getOptField(x, "!") != None):
+                    listbody.append(x)
```
mbudiu-vmw: if "continue" would work here I think it would make the code easier to understand - and perhaps faster too.

tools/souffle_converter.py
```diff
@@ -1046,14 +1047,87 @@ def convert_terms(self, terms):
...
+                child = getOptField(x, "Disjunction")
+                if child != None:
+                    if getOptField(child, "OR") != None:
+                        bool_dis = True
+                    listbody.append(child)
+            elif (x.symbol.name in ["OR", "(", ")", ",", "!"]):
+                listbody.append(x)
+            else:
```
mbudiu-vmw: Can you add a comment describing when the else is taken and why ignoring the listbody result is OK?

tools/souffle_converter.py
```diff
@@ -1046,14 +1047,87 @@ def convert_terms(self, terms):
...
+                (_, bool_dis) = self.flatten(x, listbody, bool_dis)
+        return (listbody, bool_dis)
+
+    def findCommas(self, comma):
+        commas = False
```
mbudiu-vmw: In general I find it more readable to call such boolean variables hasCommas (or isXXX). Same for your bool_XXX naming above.

tools/souffle_converter.py
```diff
@@ -1046,14 +1047,87 @@ def convert_terms(self, terms):
...
+        for x in comma.children:
+            if x.symbol.name == ",":
+                return True
+            elif not (x.symbol.name in ["Term", "OR", "(", ")", "!"]):
+                commas = self.findCommas(x)
```
mbudiu-vmw: I think that this would be slightly more readable using no "commas" variable and returning directly here, and False below.

tools/souffle_converter.py
```diff
@@ -1046,14 +1047,87 @@ def convert_terms(self, terms):
...
+        return commas
+
+    def disconjunction(self, listbody):
```
mbudiu-vmw: how about calling it disConjunction?
or getConjunctions?

tools/souffle_converter.py
```diff
@@ -1046,14 +1047,87 @@ def convert_terms(self, terms):
...
+                (listbody, _) = self.flatten(choice, [], False)
+                if bool_comma:
+                    disconlist = self.condisjunction(listbody)
+                else:
+                    disconlist = self.disconjunction(listbody)
+                for elem in disconlist:
+                    dislist.append(elem)
+        return dislist
+
+    def condisjunction(self, listbody):
```
mbudiu-vmw: same comment as above

Argc0:
@mbudiu-vmw  Well, I saw your comments. I will change what it needs and I will update with a new PR.

PR #257, opened 2019-06-15
Update handler framework

ryzhyk:
This PR introduces a modular framework for handling update notifications
from differential.

At the low level differential supports a callback API that gets invoked
every time a record is added or deleted in an output relation.  
DDlog can perform one or more of the following actions inside the callback:
(1) invoke an extern "C" callback, (2) apply update to a copy of output relations
cached by DDlog, (3) (as of this PR) accumulate changes across one or
multiple transactions, (4) (as of this PR) forwarding the update to a worker
thread that can perform any combination of 1,2,3, and, in the future, (5) forwarding
update to a pool of worker threads.

To make arbitrary combinations of the above, as well as other possible future features,
this PR introduces a modular framework that allows chaining the above actions in arbitrary
ways. It also implements (3) and (4) above. (4) is now in use by default; (3) is available
through a new API method: `transaction_commit_dump_changes()`

### DIFF
--- a/java/ddlogapi/DDlogAPI.java
+++ b/java/ddlogapi/DDlogAPI.java
@@ -16,6 +16,8 @@ public class DDlogAPI {
      * The C ddlog API
      */
     native long ddlog_run(boolean storeData, int workers, String callbackName);
+    static native int ddlog_record_commands(long hprog, String filename);
+    static native int ddlog_stop_recording(long hprog, int fd);
     native int dump_table(long hprog, int table, String callbackMethod);
     static native int  ddlog_stop(long hprog);
     static native int ddlog_transaction_start(long hprog);
@@ -70,6 +72,9 @@ public class DDlogAPI {
 
     // This is a handle to the program; it wraps a void*.
     private final long hprog;
+
+    // File descriptor used to record DDlog command log
+    private int record_fd = -1;
 
     public static final int success = 0;
     public static final int error = -1;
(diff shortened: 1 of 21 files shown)


ryzhyk:
@mbudiu-vmw, can you please check that my Java API changes look correct.

java/test1/RedistTest.java
```diff
@@ -382,9 +386,20 @@ void parseLine(String line)
                         this.runCommands();
                     break;
                 case "profile":
-                    this.exitCode = this.api.profile();
-                    this.checkExitCode();
-                    this.checkSemicolon();
+                    if (rest.equals(" cpu on")) {
```
mbudiu-vmw: I am starting to think that this code should be moved into a utility class for processing dat files - but I should do this later.

rust/template/api.rs
```diff
@@ -0,0 +1,580 @@
+#![allow(non_snake_case, non_camel_case_types, non_upper_case_globals, unused_parens, non_shorthand_field_patterns, dead_code, overflowing_literals)]
```
mbudiu-vmw: A comment indicating the purpose of this file would be nice.

rust/template/api.rs
```diff
@@ -0,0 +1,580 @@
...
+use std::os::unix::io::{IntoRawFd, FromRawFd};
+use std::mem;
+use std::sync::{Mutex, Arc};
+use libc::size_t;
+
+use super::valmap::*;
+use super::update_handler::*;
+use super::*;
+
+pub struct HDDlog {
```
mbudiu-vmw: I can't guess what HDD is for.

rust/template/api.rs
```diff
@@ -0,0 +1,580 @@
...
+    pub fn eprintln(&self, msg: &str)
+    {
+        Self::print_err(self.print_err, msg)
+    }
+}
+
+pub fn updcmd2upd(c: &record::UpdCmd) -> Result<Update<Value>, String>
+{
+    match c {
+        record::UpdCmd::Insert(rident, rec) => {
```
mbudiu-vmw: I am guessing that if match fails you get automatically a Result that is not Ok.

java/ddlogapi/DDlogAPI.java
```diff
@@ -16,13 +16,17 @@
      * The C ddlog API
      */
     native long ddlog_run(boolean storeData, int workers, String callbackName);
+    static native int ddlog_record_commands(long hprog, String filename);
+    static native int ddlog_stop_recording(long hprog, int fd);
     native int dump_table(long hprog, int table, String callbackMethod);
     static native int  ddlog_stop(long hprog);
     static native int ddlog_transaction_start(long hprog);
     static native int ddlog_transaction_commit(long hprog);
+    native int ddlog_transaction_commit_dump_changes(long hprog, String callbackName);
```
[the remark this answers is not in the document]
ryzhyk: So that `jmethodID methodId = (*env)->GetMethodID(env, cbinfo->cls, methodstr, signature)` gets a correct `cls` value and does not fail.

rust/template/api.rs
ryzhyk: "Handle"

java/test1/RedistTest.java
ryzhyk: Agreed, we are duplicating code and my changes to parsing code are hacky.

rust/template/api.rs
ryzhyk: The pattern is exhaustive (and the compiler knows it)

PR #266, opened 2019-06-21
Test two programs in the same process

mbudiu-vmw:


mbudiu-vmw:
This seems to work correctly. It runs two transactions using two separate APIs running the same program.

mbudiu-vmw:
The transactions are concurrent.

PR #271, opened 2019-06-25
Make else optional; fixes #171

### LINKED ISSUE #171: Make the `else` clause optional
The `else` clause in an `if-else` expression is currently mandatory.  The thinking behind this design was that in an expression-oriented language all expressions must return something. In practice, `if-else` expressions often return an empty tuple (`()`), and have the trivial `else` clause: `if () {...} else {()}`. Rather than forcing the user to type this every time, we should allow skipping the `else` clause and imply empty tuple.  The type checker will complain in case the  `if`-branch has a different type.

mbudiu-vmw:


ryzhyk:
Thanks for the quick fix!

I confirm that my original program that triggered the bug also work correctly now.

PR #272, opened 2019-06-25
Api doc

remysucre:
Add documentation on using the API from rust.

### DIFF
--- a/test/datalog_tests/api_test/run.sh
+++ b/test/datalog_tests/api_test/run.sh
@@ -1,0 +1,4 @@
+#!/bin/sh
+
+# compile ../copy.dl, then compile and run api_test
+stack test --ta "-p copy" && cargo run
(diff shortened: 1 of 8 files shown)


test/datalog_tests/api_test/README
```diff
@@ -0,0 +1 @@
+A client program to [`copy.dl`](../copy.dl). Compile the DDlog code (e.g. with stack test --ta "-p copy") before compiling this client.
```
mbudiu-vmw: I would call this file README.md; then it displays nicely on github.
Perhaps it's better to point to the run.sh script in the readme too.

test/datalog_tests/api_test/run.sh
```diff
@@ -0,0 +1 @@
+stack test --ta "-p copy" && cargo run
```
mbudiu-vmw: Please add #!/bin/sh and a comment describing what this does

test/datalog_tests/api_test/src/main.rs
```diff
@@ -0,0 +1,34 @@
+use copy_ddlog::api::*;
```
mbudiu-vmw: This file also needs a comment indicating that it is a "main" entry point for a specific datalog program.

PR #275, opened 2019-06-25
mark public functions as unsafe

remysucre:
Public functions that dereference raw pointer arguments should be marked unsafe. See https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref


remysucre:
These are found by clippy (level of "error").

PR #277, opened 2019-06-27
Rust api

remysucre:
Refactor some of the C API into a safe Rust API + an unsafe C wrapper. 

- I feel comfortable for the methods at the bottom of `api.rs`
- Still not too sure how to handle `ddlog_run` and `update_handler.rs`
- ~~I'll try to make `record_commands` safe, but not too sure how to yet.~~ made it safe

### DIFF
--- a/rust/template/api.rs
+++ b/rust/template/api.rs
@@ -278,16 +278,17 @@ pub unsafe extern "C" fn ddlog_dump_input_snapshot(prog: *const HDDlog, fd: unix
     if prog.is_null() || fd < 0 {
         return -1;
     };
-    let mut prog = Arc::from_raw(prog);
+    let prog = Arc::from_raw(prog);
     let mut file = fs::File::from_raw_fd(fd);
-    let res = prog.dump_input_snapshot(&mut file);
+    let res = prog.dump_input_snapshot(&mut file)
+                  .map(|_|0)
+                  .unwrap_or_else(|e| {
+                      prog.eprintln(&format!("ddlog_dump_input_snapshot: error: {}", e));
+                      -1
+                  });
     file.into_raw_fd();
     Arc::into_raw(prog);
-    res.map(|_|0)
-       .unwrap_or_else(|e| {
-           prog.eprintln(&format!("ddlog_dump_input_snapshot: error: {}", e));
-           -1
-        })
+    res
 }
 
 impl HDDlog {
@@ -519,7 +520,7 @@ pub unsafe extern "C" fn ddlog_transaction_rollback(prog: *const HDDlog) -> raw:
 
 impl HDDlog {
     pub fn transaction_rollback(&self) -> Result<(), String> {
-        self.record_transaction_rollback();
+        let _ = self.record_transaction_rollback();
         self.prog.lock().unwrap().transaction_rollback()
     }
 
@@ -565,10 +566,8 @@ impl HDDlog {
 
         let upds_vec: Result<Vec<_>, _> = upds.map(|upd| updcmd2upd(upd))
                                               .collect();
-
-        upds_vec.map(|upds| {
-            self.prog.lock().unwrap().apply_updates(upds);
-        })
+        let upds = upds_vec?;
+        self.prog.lock().unwrap().apply_updates(upds)
     }
 
     pub fn record_updates<'a, I: iter::ExactSizeIterator<Item=&'a record::UpdCmd>>(&self, upds: I) {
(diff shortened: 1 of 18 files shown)


ryzhyk:
There are some `^^^^^^ unnecessary `unsafe` block` warnings now, that the API functions themselves are marked unsafe.

remysucre:
I think I've made the necessary changes for a useable rust API. There are some more code that could be refactored, but I refrained from spending too much more time on it and mostly preserved previous code structure. If these look OK I'll put the Rust methods in the same `impl` block and copy over the documentation, and also address some of the compiler warnings...

ryzhyk:
Fantastic, thank you! I will try to review quickly. But we will need to do some testing before merging into master.  The easiest way to do this without writing many new tests is to port `rust/template/main.rs` to your new API. This is not urgent and can be done in the background.

In addition, we should make sure that Java and OVN tests pass.

ryzhyk:
@remysucre, can you please review my changes to Rust code.
@mbudiu-vmw, can you please review my changes to Java bindings?

rust/template/main.rs
```diff
@@ -106,7 +103,7 @@ fn handle_cmd(handler: &Box<dyn IMTUpdateHandler<Value>>,
                 },
                 Some(rid) => rid as RelId
             };
-            let _ = db.lock().unwrap().format_rel(relid, &mut stdout());
+            let _ = hddlog.db.as_ref().map(|db|db.lock().unwrap().format_rel(relid, &mut stdout()));
```
remysucre: use the dump_table API?

rust/template/main.rs
```diff
@@ -28,74 +28,71 @@ use differential_datalog::program::*;
...
-                *deltadb.lock().unwrap() = Some(DeltaMap::new());
-            }
             // uncomment to enable profiling
-            handler.before_commit();
             //PROFILER.lock().unwrap().start("./prof.profile").expect("Couldn't start profiling");
-            let res = p.transaction_commit();
+            let mut current_table = None;
+            let res = if record_delta {
+                let cb = if print_deltas {
+                    Some(move |table, val: &Record, pol: bool| {
```
remysucre: The code still compiles without `move`, which is surprising. What's going on here with `current_table`? Maybe Rust is automatically converting it into a mutable reference?

rust/template/api.rs
```diff
@@ -90,18 +91,20 @@ pub extern "C" fn ddlog_get_table_id(tname: *const raw::c_char) -> libc::size_t
     if tname.is_null() {
         return libc::size_t::max_value();
     };
-    match get_table_id(tname) {
+    let table_str = unsafe{ ffi::CStr::from_ptr(tname) }.to_str().unwrap();
```
remysucre: Move `unsafe` to method signature, since this is a public API

rust/template/main.rs
ryzhyk: Actually, now that you mention this, I realize that the Rust `dump_table` and `commit_dump_changes` APIs should have been implemented to return iterators, which is a far more Rust-friendly approach that also makes it easier for the caller to observe the start and end of a transaction compared to getting callbacks. I need to move on to other things now, but I will create an issue so that one of us can eventually make this improvement.

rust/template/main.rs
remysucre: I see, it doesn't have to be a reference. `FnMut` can capture any mutable variable

rust/template/ddlog.h
```diff
@@ -125,14 +126,53 @@ extern ddlog_prog ddlog_run(unsigned int workers,
...
  * `fd` - file descriptor.  Passing -1 in this argument instructs DDlog to stop
  * recording. The caller is responsible for opening and closing the file.  They
  * can inject additional commands, e.g., `echo` to the file.
+ *
+ * IMPORTANT: this function is _not_ thread-safe and must not be invoked
+ * concurrently with other functions in this API.
  */
 extern int ddlog_record_commands(ddlog_prog hprog, int fd);
 
+/*
```
mbudiu-vmw: I find it a bit odd that the most extensive documentation is on the C API.
But I don't have an alternative proposal of where to locate it.
Perhaps we can make a table somewhere in an MD file with the function names across all languages and the corresponding documentation.

ryzhyk: I **really** want to figure out a way to write API documentation for multiple languages in a centralized way and without replicating the same text multiple times. Does not have to be MD.

remysucre: Swig might help in that regard: https://libpasta.github.io/blog/bindings/ https://github.com/Dushistov/rust_swig

PR #281, opened 2019-06-28
Nondeterministic behavior in tinyset implementation

ryzhyk:
`tinyset.Set64` implementation had several nondeterministic operations:
`cmp`, `partial_cmp`, `iter` (returned iterator that iterated through
the set in non-deterministic order), and `fmt`.

The fix implements deterministic operator (which converts the set into
a sorted vector first) and to use this iterator in all previously
non-deterministic functions.  This adds some overhead, but it's
necessary for correctness.

### DIFF
--- a/lib/tinyset.rs
+++ b/lib/tinyset.rs
@@ -13,7 +13,7 @@ use std::ops::BitOr;
 use std::collections;
 
 #[derive(Eq, Clone, Hash, PartialEq)]
-pub struct tinyset_Set64<T: u64set::Fits64 + Ord> {
+pub struct tinyset_Set64<T: u64set::Fits64> {
     pub x: u64set::Set64<T>
 }
 
@@ -53,11 +53,11 @@ impl <T: tinyset::Fits64 + Ord> tinyset_Set64<T> {
     }
 }
 
-impl <'a, T: tinyset::Fits64 + Ord> tinyset_Set64<T> {
+impl <T: tinyset::Fits64> tinyset_Set64<T> {
     /* In cases when order really, definitely, 100% does not matter,
      * a more efficient iterator can be used.
      */
-    pub fn unsorted_iter(&'a self) -> u64set::Iter64<'a, T> {
+    pub fn unsorted_iter(&self) -> u64set::Iter64<T> {
         self.x.iter()
     }
 }
@@ -116,7 +116,7 @@ impl<'de, T: u64set::Fits64 + serde::Deserialize<'de> + Ord> serde::Deserialize<
     }
 }
 
-impl <T: u64set::Fits64 + Ord> tinyset_Set64<T> {
+impl <T: u64set::Fits64> tinyset_Set64<T> {
     pub fn new() -> Self {
         tinyset_Set64{x: u64set::Set64::new()}
     }
@@ -128,7 +128,7 @@ impl <T: u64set::Fits64 + Ord> tinyset_Set64<T> {
     }
 }
 
-impl<T: FromRecord + u64set::Fits64 + Ord> FromRecord for tinyset_Set64<T> {
+impl<T: FromRecord + u64set::Fits64> FromRecord for tinyset_Set64<T> {
     fn from_record(val: &Record) -> Result<Self, String> {
         vec::Vec::from_record(val).map(|v|tinyset_Set64{x: u64set::Set64::from_iter(v)})
     }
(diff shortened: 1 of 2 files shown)


lib/tinyset.rs
```diff
@@ -41,18 +47,26 @@ impl<'a, X:u64set::Fits64 + Ord + 'a> Iterator for Set64Iter<'a, X> {
...
 
-impl <'a, T: tinyset::Fits64 + Ord> tinyset_Set64<T> {
-    pub fn iter(&'a self) -> Set64Iter<'a, T> {
+impl <T: tinyset::Fits64 + Ord> tinyset_Set64<T> {
+    pub fn iter(&self) -> Set64Iter<T> {
         Set64Iter::new(self)
     }
 }
 
+impl <'a, T: tinyset::Fits64 + Ord> tinyset_Set64<T> {
```
remysucre: Why is lifetime parameter needed here? It compiles fine without it.

lib/tinyset.rs
```diff
@@ -41,18 +47,26 @@ impl<'a, X:u64set::Fits64 + Ord + 'a> Iterator for Set64Iter<'a, X> {
...
+    /* In cases when order really, definitely, 100% does not matter,
+     * a more efficient iterator can be used.
+     */
+    pub fn unsorted_iter(&'a self) -> u64set::Iter64<'a, T> {
+        self.x.iter()
+    }
+}
+
 impl<T: fmt::Debug + u64set::Fits64 + Ord> fmt::Debug for tinyset_Set64<T> {
```
remysucre: Does this mean the elements have to implement `Ord` for debugging? Is this too stringent?

lib/tinyset.rs
```diff
@@ -65,13 +79,18 @@ impl<T: fmt::Debug + u64set::Fits64 + Ord> fmt::Debug for tinyset_Set64<T> {
...
+        } else {
+            self.iter().cmp(other.iter())
+        }
     }
 }
 
 impl<T: u64set::Fits64 + Ord> PartialOrd for tinyset_Set64<T> {
     fn partial_cmp(&self, other: &Self) -> Option<cmp::Ordering> {
-        self.x.iter().partial_cmp(other.x.iter())
+        Some(self.cmp(other))
```
remysucre: How about returning `None` for sets of different size?

lib/tinyset.rs
```diff
@@ -254,7 +274,7 @@ pub fn tinyset_group_setref_unions<X: u64set::Fits64 + Ord + Clone>(g: &std_Grou
         {
             let rres = std_Ref::get_mut(&mut res).unwrap();
             for gr in g.iter() {
-                for v in gr.iter() {
+                for v in gr.unsorted_iter() {
```
remysucre: Change line 258 to `gr.unsorted_iter()` too?

lib/tinyset.rs
remysucre: Same for Display

remysucre: Ah I see, you are requiring `Ord` for using sets at all time. Maybe we can also provide an "Unordered Set" type for those who really don't want / can't implement `Ord`?

lib/tinyset.rs
```diff
@@ -17,19 +17,25 @@ pub struct tinyset_Set64<T: u64set::Fits64 + Ord> {
     pub x: u64set::Set64<T>
 }
 
-/* This is needed so we can support for-loops over `Set64`'s
+/* This is needed so we can support for-loops over `Set64`'s.
+ *
+ * IMPORTANT: We iterate over Set64 by converting it into a sorted vector.
```
mbudiu-vmw: hopefully this conversion is itself deterministic

lib/tinyset.rs
ryzhyk: Ok, removed the lifetime

lib/tinyset.rs
ryzhyk: Thanks, I removed redundant trait bounds thoughout the file

lib/tinyset.rs
ryzhyk: My `Ord` implementation actually says that a smaller-size set is also smaller in the total order (which makes a big difference for performance), so `PartialOrd` just inherits this semantics.

lib/tinyset.rs
ryzhyk: good point

lib/tinyset.rs
ryzhyk: I think this is the case, since the element type must be totally ordered (`Ord`)

Issue #268, opened 2019-06-21
Generate Java classes for DDlog input relations


ryzhyk:

mbudiu-vmw:
I think that we can close this one, since it is superseded by the Flatbuf work.

PR #283, opened 2019-06-29
Java code generation with useful methods and classes

mbudiu-vmw:
The -j option of ddlog will cause it to generate a .java file which contains useful classes and method for interacting more efficiently with the generated code. Currently we only support base types for relation elements.

We generate:
* constants corresponding to table ids
* HashMaps mapping table ids to strings and viceversa
* functions that create DDlogRecords for each relation from raw column values
* one class for each relation with multiple methods:
  * constructor from DDlogRecord
  * toString matching the Rust string serialization used by the CLI
  * createCommand - creates an insert/delete command for the corresponding row
  * constructor from raw field values

Here is an example:
```Java
import java.util.*;
import ddlogapi.DDlogRecord;

import ddlogapi.DDlogCommand;

class Span_string
{
    public static final int TableId_Binding = 0;
    public static HashMap<Integer, String> idToName = new HashMap<Integer, String>();
    public static HashMap<String, Integer> nameToId = new HashMap<String, Integer>();
    static
    {
        idToName.put(TableId_Binding, "Binding");
        nameToId.put("Binding", TableId_Binding);
    }
    public static DDlogRecord create_Binding(String entity, String tn)
    {
        DDlogRecord rentity = new DDlogRecord(entity);
        DDlogRecord rtn = new DDlogRecord(tn);
        DDlogRecord[] a = {rentity, rtn};
        DDlogRecord r = DDlogRecord.makeStruct("Binding", a);
        return r;
    }
    public static class Binding
    {
        public String entity;
        public String tn;
        public Binding(DDlogRecord r)
        {
            this.entity = r.getStructField(0).getString();
            this.tn = r.getStructField(1).getString();
        }
        public Binding(String entity, String tn)
        {
            this.entity = entity;
            this.tn = tn;
        }
        public DDlogRecord asRecord()
        {
            return create_Binding(this.entity, this.tn);
        }
        @Override public String toString()
        {
            return "Binding" + "{" + this.entity.toString() + "," + 
                                     this.tn.toString() + "}";
        }
        public DDlogCommand createCommand(boolean insert)
        {
            return new DDlogCommand(insert ? DDlogCommand.Kind.Insert : DDlogCommand.Kind.DeleteVal,
            TableId_Binding,
            this.asRecord());
        }
    }
}
```

### DIFF
--- a/java/test3/SpanTest.java
+++ b/java/test3/SpanTest.java
@@ -1,0 +1,36 @@
+import java.io.IOException;
+import java.util.*;
+import java.lang.RuntimeException;
+
+import ddlogapi.DDlogAPI;
+import ddlogapi.DDlogCommand;
+import ddlogapi.DDlogRecord;
+
+public class SpanTest {
+    private final DDlogAPI api;
+
+    SpanTest() {
+        this.api = new DDlogAPI(1, r -> this.onCommit(r), false);
+    }
+
+    void onCommit(DDlogCommand command) {
+        System.out.println(command.toString());
+    }
+
+    void run() {
+        Span_string.Binding binding = new Span_string.Binding("1", "2");
+        DDlogCommand command0 = binding.createCommand(true);
+        Span_string.FWRule rule = new Span_string.FWRule("1");
+        DDlogCommand command1 = rule.createCommand(true);
+        DDlogCommand[] commands = { command0, command1 };
+        this.api.start();
+        this.api.applyUpdates(commands);
+        this.api.commit();
+        this.api.stop();
+    }
+
+    public static void main(String[] args) throws IOException {
+        SpanTest test = new SpanTest();
+        test.run();
+    }
+}
--- a/java/test3/SpanTest.java
+++ b/java/test3/SpanTest.java
@@ -10,7 +10,7 @@ public class SpanTest {
     private final DDlogAPI api;
 
     SpanTest() {
-        this.api = new DDlogAPI(1, r -> this.onCommit(r), false);
+        this.api = new DDlogAPI(1, null, false);
     }
 
     void onCommit(DDlogCommand command) {
@@ -25,7 +25,7 @@ public class SpanTest {
         DDlogCommand[] commands = { command0, command1 };
         this.api.start();
         this.api.applyUpdates(commands);
-        this.api.commit();
+        this.api.commit_dump_changes(r -> this.onCommit(r));
         this.api.stop();
     }
 
--- a/java/test3/run.sh
+++ b/java/test3/run.sh
@@ -18,7 +18,7 @@ esac
 
 # Compile the span_uuid.dl DDlog program
 ddlog -i ../../test/datalog_tests/span_string.dl -j -L../../lib
-# Compile the rust program; generates ../test/datalog_tests/span_ddlog/target/release/libspan_ddlog.a and .so
+# Compile the rust program; generates ../test/datalog_tests/span_ddlog/target/release/libspan_ddlog.a
 pushd ../../test/datalog_tests/span_string_ddlog
 cargo build --release
 popd
--- a/java/test3/run.sh
+++ b/java/test3/run.sh
@@ -27,7 +27,7 @@ make -C ..
 # Compile SpanTest.java and the generated Java file
 javac -cp .. SpanTest.java Span_string.java
 # Create a shared library containing all the native code: ddlogapi.c, libspan_string_ddlog.a
-${CC} -shared -fPIC -I${JAVA_HOME}/include -I${JAVA_HOME}/include/${JDK_OS} -I../../rust/template ../ddlogapi.c -L../../test/datalog_tests/span_string_ddlog/target/release/ -lspan_string_ddlog -o libddlogapi.${SHLIBEXT}
+${CC} -shared -fPIC -I${JAVA_HOME}/include -I${JAVA_HOME}/include/${JDK_OS} -I../../rust/template -I../../lib ../ddlogapi.c -L../../test/datalog_tests/span_string_ddlog/target/release/ -lspan_string_ddlog -o libddlogapi.${SHLIBEXT}
 # Run the java program pointing to the created shared library
 java -Djava.library.path=. -cp ../ddlogapi.jar:. SpanTest > spantest.dump
 diff spantest.dump spantest.dump.expected
--- a/java/test3/run.sh
+++ b/java/test3/run.sh
@@ -24,8 +24,6 @@ cargo build --release
 popd
 # Build the Java library with the DDlog API
 make -C ..
-# Force linking with the static library by deleting the dynamic library
-rm -f ../../test/datalog_tests/span_string_ddlog/target/release/libspan_string_ddlog.so
 # Compile SpanTest.java and the generated Java file
 javac -cp .. SpanTest.java Span_string.java
 # Create a shared library containing all the native code: ddlogapi.c, libspan_string_ddlog.a
--- a/java/test3/run.sh
+++ b/java/test3/run.sh
@@ -1,0 +1,37 @@
+#!/bin/bash
+# Shell script to build and run a Java program
+# tied to test/datalog_tests/span_string.dl program
+
+set -ex
+
+if command clang -v 2>/dev/null; then
+    export CC=clang
+else
+    export CC=gcc
+fi
+
+case `uname -s` in
+    Darwin*)    SHLIBEXT=dylib; JDK_OS=darwin;;
+    Linux*)     SHLIBEXT=so; JDK_OS=linux;;
+    *)          echo "Unsupported OS"; exit -1
+esac
+
+# Compile the span_uuid.dl DDlog program
+ddlog -i ../../test/datalog_tests/span_string.dl -j -L../../lib
+# Compile the rust program; generates ../test/datalog_tests/span_ddlog/target/release/libspan_ddlog.a and .so
+pushd ../../test/datalog_tests/span_string_ddlog
+cargo build --release
+popd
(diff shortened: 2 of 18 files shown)


java/test3/SpanTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+import ddlogapi.DDlogAPI;
+import ddlogapi.DDlogCommand;
+import ddlogapi.DDlogRecord;
+
+public class SpanTest {
+    private final DDlogAPI api;
+
+    SpanTest() {
+        this.api = new DDlogAPI(1, r -> this.onCommit(r), false);
```
ryzhyk: Can we pass this as a callback to `commit_dump_changes` instead? I'd like to discourage people from using this callback...

java/test3/run.sh
```diff
@@ -0,0 +1,37 @@
...
+# Compile the span_uuid.dl DDlog program
+ddlog -i ../../test/datalog_tests/span_string.dl -j -L../../lib
+# Compile the rust program; generates ../test/datalog_tests/span_ddlog/target/release/libspan_ddlog.a and .so
+pushd ../../test/datalog_tests/span_string_ddlog
+cargo build --release
+popd
+# Build the Java library with the DDlog API
+make -C ..
+# Force linking with the static library by deleting the dynamic library
+rm -f ../../test/datalog_tests/span_string_ddlog/target/release/libspan_string_ddlog.so
```
ryzhyk: This should no longer be needed, as `ddlog` now has flags to explicitly control static/dynamic lib generation:
```
           --dynlib         generate dynamic library
           --no-dynlib      do not generate dynamic library (default)
           --staticlib      generate static library (default)
           --no-staticlib   do not generate static library
```

so only the static lib will be generated by default

mbudiu-vmw:
I am still working on this file, so it's simpler to suggest changes than to make all of them.
I can make the changes once you have a suggestion.
I will merge yours for now, but you should not waste your time on beautification that I can also do.

ryzhyk:
Ok, sorry. Feel free to ignore.

The only thing that actually looked like a bug are the iterations of the form
`map XXX [0..(length typeTupArgs)]`.  I suspect what you meant is `[0..(length typeTupArgs - 1)]`

mbudiu-vmw:
I have rebased on your changes, so they are not lost.

mbudiu-vmw:
The last commit adds support for vectors and sets.
Still to be done: support for nested structs.

mbudiu-vmw:
The last commit adds support for nested structs.
The only thing that is missing is iterating the code generation until reaching a fixed-point; I didn't know how to do this using monads - see the TODO in the code.
The Java code generation is now good enough to handle many interesting test cases.

src/Language/DifferentialDatalog/CompileJava.hs
```diff
@@ -0,0 +1,487 @@
...
+generateJava :: DatalogProgram -> String -> Doc
+generateJava d sourceName =
+    preamble sourceName $+$
+    (braces' $ body d)
+
+-- Generates the preamble of a Java file
+preamble :: String -> Doc
+preamble sourceName =
+    "import java.util.*;"                       $+$
+    "import ddlogapi.DDlogRecord;\n"            $+$
```
ryzhyk: No need for `\n` here and in the next line.

src/Language/DifferentialDatalog/CompileJava.hs
```diff
@@ -0,0 +1,487 @@
...
+    typesCompiled  = S.empty,
+    typesToCompile = M.empty
+}
+
+-- Generates the body of the Java class containing all utility methods
+body :: DatalogProgram -> Doc
+body d =
+    evalState (
+        do relations <- mapM (genRelationCreate d) (M.keys $ progRelations d)
+           -- TODO: this should iterate until todo is empty
```
ryzhyk: The simplest way to do this is to add a function to `Type.hs` to recursively compute all types mentioned in a type declaration (I can do this), call this function for each type in `typesToCompile`, and union the results. Then you won't need fix point here, and `genCreate` won't have to return a monad.

src/Language/DifferentialDatalog/CompileJava.hs
```diff
@@ -0,0 +1,487 @@
...
+    let rtype = relType $ progRelations M.! relationName in
+    genClass d relationName rtype True
+
+-- Generate a Java class for a specific type
+genClass :: DatalogProgram -> String -> Type -> Bool -> Doc
+genClass d@DatalogProgram{} name rtype isRelation =
+    "public static class" <+> makeIdentifier name $+$
+    (braces' $ classBody d name rtype isRelation)
+
+-- given a type generate the body of a Java class storing the columns of the relation
```
ryzhyk: Isn't supporting multiple constructors just a matter of generating a class per constructor?

src/Language/DifferentialDatalog/CompileJava.hs
```diff
@@ -0,0 +1,487 @@
...
+  -- toString method
+  "@Override public String toString()" $+$
+  (braces' $ "return" <+> (doubleQuotes $ pp className) <+> "+ \"{\" +" <+> (toStringBody d t) <+> "+ \"}\"" <> semi) $+$
+  -- command to insert
+  if isRelation then
+      "public DDlogCommand createCommand(boolean insert)" $+$
+      (braces' $ ("return new DDlogCommand(insert ? DDlogCommand.Kind.Insert : DDlogCommand.Kind.DeleteVal," $+$
+              (pp $ "TableId_" ++ className ++ ",") $+$
+              "this.asRecord());"))
+  else ""
```
ryzhyk: `""` -> `empty`

src/Language/DifferentialDatalog/CompileJava.hs
```diff
@@ -0,0 +1,487 @@
...
+
+-- given a TStruct extract the first constructor if it is the only one;
+-- report an error otherwise.  Thsi is because we don't support (yet) structs
+-- with multiple constructors
+getSingleConstructor :: Type -> Constructor
+getSingleConstructor t@TStruct{..} =
+    let c0:tail = typeCons in
```
ryzhyk: `tail` is a function from Haskell preamble; better not to overload the name.

src/Language/DifferentialDatalog/CompileJava.hs
```diff
@@ -0,0 +1,487 @@
...
+genRelationCreate d relationName =
+    let rtype = relType $ fromJust $ M.lookup relationName $ progRelations d in
+    genCreate d relationName rtype
+
+-- given a type name and a type this creates a function to construct a
+-- DDlogRecord that can be inserted or deleted in the relation
+genCreate :: DatalogProgram -> String -> Type -> JavaMonad Doc
+genCreate d name rtype = do
+    done <- gets (\state -> S.member name $ typesCompiled state)
+    if done then return ""
```
ryzhyk: `""` -> `empty`

src/Language/DifferentialDatalog/CompileJava.hs
```diff
@@ -0,0 +1,487 @@
...
+    pp (CSet t)    = "Set<" <> (pp t) <> ">"
+
+-- resolves a TypeDef to its underlying type (which is a Collection description)
+resolveType :: DatalogProgram -> Type -> Collection
+resolveType d t =
+    -- trace (show t) $
+    case t of
+        -- it is arguably wrong to hardwire these type names in the compiler, but we assume that
+        -- they are more or less standard types.  Maybe the compiler should know about them?
+        TUser{..} -> if typeName == "std.Ref"
```
ryzhyk: You can write this as:
```
case t of
    TUser{typeName = "std.Ref", ..} -> ...
    TUser{..} | elem typeName ["std.Set", "tinyset.Set64"] -> ..
    ...
```

src/Language/DifferentialDatalog/CompileJava.hs
```diff
@@ -0,0 +1,487 @@
...
+                         let inner = resolveType d $ head typeArgs in
+                         case inner of
+                              CNone elem   -> CSet elem
+                              _            -> error $ "Nested collections not supported: " ++ show t
+                     else if typeName == "std.Vector" then
+                         let inner = resolveType d $ head typeArgs in
+                         case inner of
+                              CNone elem   -> CVector elem
+                              _            -> error $ "Nested collections not supported: " ++ show t
+                     else resolveType d actual
```
ryzhyk: There's a `typ'` function in `Type.hs` that recursively unwraps typedef's, returning a TOpaque or some other type that is not `TUser`. It might help to simplify this code...

src/Language/DifferentialDatalog/CompileJava.hs
mbudiu-vmw: I wanted an empty line

src/Language/DifferentialDatalog/CompileJava.hs
mbudiu-vmw: I thought about this; we also need to be able to tell whether a type comes from a relation or not, since we generate slightly different code.

src/Language/DifferentialDatalog/CompileJava.hs
mbudiu-vmw: That's one way to do it. These classes would also probably need to inherit from a common interface. Let's leave this until someone needs this feature.

mbudiu-vmw:
Fixes #268

mbudiu-vmw:
the macos build seems to have been stuck

ryzhyk:
`run-souffle-tests` tends to timeout due to not printing anything for 10 minutes. I will wrap it in `travis_wait` to hopefully fix this.

Issue #274, opened 2019-06-25
Address Clippy warning


remysucre:
Clipply warnings may be addressed at least for rust/template. For generated code, there should be a way to selectively turn off clippy.

ryzhyk:
Are there more warnings left? Is there a way to integrate Clippy to run during `cargo build` or `cargo test` or otherwise integrate it into the build process?

remysucre:
There are a lot more none-critical warnings. I just found a tool that auto-applies warnings, which should speed this up a lot (we should review the changes of course) https://github.com/rust-lang-nursery/rustfix. Yes there's a way to have this in the build process (https://github.com/mwillsey/egg/blob/master/check.sh) I can give it a try

remysucre:
Next steps are: 
1. fix more warnings
2. have clippy run as part of the build
3. run `cargo fmt`

d-e-s-o:
Let me take this one. I just enabled `rustfmt` (for https://github.com/vmware/differential-datalog/issues/338) and clippy should be similar in many respects.

d-e-s-o:
`clippy` is now running as part of Travis and checking various crates. Here are the remaining tasks I see:
- fix the individual lints we currently white list or investigate whether to keep the contentious ones in the white list:
  - [ ] clippy::char_lit_as_u8
  - [ ] clippy::many_single_char_names
  - [ ] clippy::enum_variant_names
  - [ ] clippy::expect_fun_call
  - [ ] clippy::extra_unused_lifetimes
  - [ ] clippy::for_kv_map
  - [ ] clippy::get_unwrap
  - [ ] clippy::into_iter_on_ref
  - [ ] clippy::just_underscores_and_digits
  - [ ] clippy::let_and_return
  - [ ] clippy::map_clone
  - [ ] clippy::needless_return
  - [ ] clippy::new_without_default
  - [ ] clippy::ptr_arg
  - [ ] clippy::ptr_offset_with_cast
  - [ ] clippy::redundant_closure
  - [ ] clippy::should_implement_trait
  - [ ] clippy::trivially_copy_pass_by_ref
  - [ ] clippy::single_match
  - [ ] clippy::type_complexity
  - [ ] clippy::unreadable_literal
- [ ] push white listing to the individual crates (we should not have to touch `.travis` for such policy decisions)
  - that will also allow us to run the same `clippy` command locally more easily

d-e-s-o:
[...]

PR #288, opened 2019-07-05
Clippy

remysucre:
Address some of the clippy warnings in rust/template(#274)

### DIFF
--- a/rust/template/api.rs
+++ b/rust/template/api.rs
@@ -57,7 +57,7 @@ impl HDDlog {
     {
         Self::do_run(workers,
                      do_store,
-                     cb.map(|f| CallbackUpdateHandler::new(f)),
+                     cb.map(CallbackUpdateHandler::new),
                      None)
     }
 
@@ -221,7 +221,7 @@ impl HDDlog {
                 let delta_handler = DeltaUpdateHandler::new(deltadb2);
 
                 let store_handler = if do_store {
-                    nhandlers = nhandlers + 1;
+                    nhandlers += 1;
                     Some(ValMapUpdateHandler::new(db2))
                 } else {
                     None
@@ -233,8 +233,8 @@ impl HDDlog {
                 } else {
                     let mut handlers: Vec<Box<dyn UpdateHandler<Value>>> = Vec::new();
                     handlers.push(Box::new(delta_handler));
-                    store_handler.map(|h| handlers.push(Box::new(h)));
-                    cb_handler.map(|h| handlers.push(h));
+                    if let Some(h) = store_handler { handlers.push(Box::new(h))};
+                    if let Some(h) = cb_handler { handlers.push(h) };
                     Box::new(ChainedUpdateHandler::new(handlers))
                 };
                 handler
@@ -253,8 +253,8 @@ impl HDDlog {
             prog:           Mutex::new(prog),
             update_handler: handler,
             db:             Some(db),
-            deltadb:        deltadb,
-            print_err:      print_err,
+            deltadb,
+            print_err,
             replay_file:    None}
     }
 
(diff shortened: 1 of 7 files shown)


rust/template/update_handler.rs
```diff
@@ -325,11 +325,11 @@ impl UpdateHandler<Value> for DeltaUpdateHandler
         let handler = self.clone();
         Box::new(move |relid, v, w|{
             let guard_ptr = handler.locked.get();
-            if guard_ptr != ptr::null_mut() {
+            if !guard_ptr.is_null() {
                 let mut guard: Box<MutexGuard<'_, Option<DeltaMap>>> = unsafe {
                     Box::from_raw(guard_ptr as *mut MutexGuard<'_, Option<DeltaMap>>)
                 };
-                (*guard).as_mut().map(|db|db.update(relid, v, w));
+                if let Some(db) = (*guard).as_mut() { db.update(relid, v, w) };
```
ryzhyk: Do you know why Clippy prefers these?

rust/template/valmap.rs
```diff
@@ -40,34 +41,40 @@ impl ValMap {
     }
 
     pub fn get_rel(&mut self, relid: RelId) -> &BTreeSet<Value> {
-        self.map.entry(relid).or_insert(BTreeSet::default())
+        self.map.entry(relid).or_insert_with(BTreeSet::default)
```
ryzhyk: Good catch! This might have been an expensive bug.

rust/template/update_handler.rs
remysucre: Mostly for readability: https://rust-lang.github.io/rust-clippy/master/index.html#option_map_unit_fn IMO each one looks fine

rust/template/api.rs
```diff
@@ -629,7 +631,7 @@ pub unsafe extern "C" fn ddlog_apply_updates(prog: *const HDDlog, upds: *const *
     let mut cmds_vec: Vec<Box<record::UpdCmd>> = Vec::with_capacity(n as usize);
 
     for i in 0..n {
-        cmds_vec.push(Box::from_raw(*upds.offset(i as isize)));
+        cmds_vec.push(Box::from_raw(*upds.add(i)));
```
ryzhyk: Do you know why Clippy insists on this?

rust/template/lib.rs
```diff
@@ -1,4 +1,4 @@
-#![allow(non_snake_case, non_camel_case_types, non_upper_case_globals, unused_parens, non_shorthand_field_patterns, dead_code, overflowing_literals)]
+#![allow(unused_imports, non_snake_case, non_camel_case_types, non_upper_case_globals, unused_parens, non_shorthand_field_patterns, dead_code, overflowing_literals, clippy::ptr_arg)]
```
ryzhyk: How come you didn't have to add clippy in Cargo.toml for this to work?

rust/template/api.rs
remysucre: To avoid casts: https://rust-lang.github.io/rust-clippy/master/index.html#option_map_unit_fn but I'm not exactly sure why casts are bad / how bad they are.

rust/template/lib.rs
remysucre: I think this was a compiler warning, not from clippy.

rust/template/main.rs
```diff
@@ -35,6 +32,7 @@ use std::io::Write;
 //extern crate cpuprofiler;
 //use cpuprofiler::PROFILER;
 
+#[allow(clippy::let_and_return)]
```
ryzhyk: In what line does clippy about let-and-return?

rust/template/lib.rs
ryzhyk: I am wondering about `clippy::ptr_arg`.

rust/template/main.rs
remysucre: 56. We need that let if the profiling code is uncommented

rust/template/lib.rs
remysucre: Hmm maybe because I installed clippy globally? Does it not work for you?

ryzhyk: It does, actually. I guess clippy is not sufficiently integrated with rustc?  Or perhaps rust just knows to ignore these annotations.

remysucre: Yes, turns out rust ignores it if clippy isn't installed: https://doc.rust-lang.org/reference/attributes/diagnostics.html#tool-lint-attributes

PR #291, opened 2019-07-07
Logging API

ryzhyk:
This PR introduces a logging library that provides additional visibility
into DDlog rule evaluation. The logging API has two facets.  First, it provides
the `log()` function that can be invoked from a DDlog rule to write a message
into a log whenever the rule is evaluated.  In addition to the usual log level
and log message arguments, this function takes a `module` argument, which
identifies the module or subsystem that this log message is associated with.
The API is defined in the new `lib/log.dl` library:

```
typedef module_t = signed<32>
typedef log_level_t = signed<32>
extern function log(module: module_t, level: log_level_t, msg: string): bool
```

Convenience wrappers for using the logging API with log4j are in
`lib/log4j.dl`:

```
function trace(module: module_t, msg: string): bool = log(module, 600/*TRACE*/, msg)
function debug(module: module_t, msg: string): bool = log(module, 500/*DEBUG*/, msg)
function info (module: module_t, msg: string): bool = log(module, 400/*INFO*/,  msg)
function warn (module: module_t, msg: string): bool = log(module, 300/*WARN*/,  msg)
function error(module: module_t, msg: string): bool = log(module, 200/*ERROR*/, msg)
function fatal(module: module_t, msg: string): bool = log(module, 100/*FATAL*/, msg)
```

The second facet of this API is a logging configuration function (in Rust) that can
only be invoked by the "host" program to set log levels and logging callbacks for
individual modules.  This way, different modules can produce separate log streams or
even use different logging mechanisms. It is declared in `lib/log.rs`:

```
type log_callback_t = Box<dyn Fn(log_log_level_t, &str) + Send + Sync>;
pub fn log_set_callback(module: log_module_t, cb: Option<log_callback_t>, max_level: log_log_level_t)
```

Matching C bindings are in `lib/ddlog_log.h`; Java bindings are in
`DDlogAPI.java` (see `DDlogAPI.log_set_callback()`).

The current implementation does some caveats:

- DDlog guarantees that the log callback function will be invoked at least
  once for each evaluated rule prefix.  But this is the only guarantee it
  provides.  It might invoke the callback multiple times per each evaluated rule,
  and it can invoke callbacks in arbitrary order.  It may further invoke the
  callback both during insertion and deletion operations (i.e., rule inputs can
  have positive or negative polarity), and there is currently no way
  to tell which one it is.

  The last limitation could be addressed by using the differential
  `inspect` operator for logging.  However, this would introduce
  some additional overhead.  In addition, doing this cleanly requires
  first adding support for lambdas in DDlog.

- Internally, this library maintains a global module-to-log_callback
  map **shared by all DDlog instances running in the same address space**.  The host
  program is responsible for maintaining this mapping via the Rust or Java API.

- Logging overhead: Even when logging is disabled for a particular
  module, log statements still have a small cost.  It is therefore recommended
  to comment out unused log statements when possible.  Specifically, depending
  on the context, the log statement may end up being compiled into a separate
  dataflow operator.  It is a light-weight operator that does not cost much
  to evaluate, but it nevertheless increases the size of the control flow
  graph.  In addition, the log function evaluates its argument eagerly even
  when not invoking the callback.  In particular, string operations can
  be expensive.

=== Example ===

In DDlog:
```
// Declare module id's for logging purposes
function mod_SPAN_UUID(): log.module_t = 100

// Use log() function in a rule
Span(entity, tn) :- Binding(entity, tn),
                    log(mod_SPAN_UUID(), 5, "Span(entity,tn) :- Binding(.entity=${deref(entity)},.tn=${deref(tn)}))").
```

In Java:
```
// Module id declarations must match DDlog values:
static int MOD_SPAN_UUID1 = 100;
// Set logging callback and log level from Java
DDlogAPI.log_set_callback(
    MOD_SPAN_UUID1,
    (msg, level) -> System.err.println("Log msg from module1 (" + level + "): " + msg),
    5);
```
===============

### DIFF
--- a/java/ddlogapi.c
+++ b/java/ddlogapi.c
@@ -75,8 +75,6 @@ static struct CallbackInfo* createCallback(JNIEnv* env, jobject obj, jstring met
     return cbinfo;
 }
 
-
-
 static void deleteCallback(struct CallbackInfo* cbinfo) {
     if (cbinfo == NULL)
         return;
@@ -282,6 +280,9 @@ void log_callback(uintptr_t callbackInfo, int level, const char *msg) {
 
 /* This function sets up a new logging callback for the given module and
  * then deallocates the old `CallbackInfo`, if any.
+ *
+ * Returns 0 if either `callback` is NULL or `callback` is not NULL and
+ * installing the new callback fails.
  */
 JNIEXPORT jlong JNICALL Java_ddlogapi_DDlogAPI_ddlog_1log_1replace_1callback(
     JNIEnv *env, jobject obj, jint module, jlong old_cbinfo, jobject callback, jint max_level) {
--- a/java/ddlogapi.c
+++ b/java/ddlogapi.c
@@ -11,6 +11,7 @@
 
 #include "ddlogapi_DDlogAPI.h"
 #include "ddlog.h"
+#include "ddlog_log.h"
 
 /* The _1 in all the function names below
    is the JNI translation of the _ character from Java */
@@ -45,7 +46,7 @@ static void printClass(JNIEnv* env, jobject obj) {
     (*env)->ReleaseStringUTFChars(env, strObj, str);
 }
 
-static struct CallbackInfo* createCallback(JNIEnv* env, jobject obj, jstring method, const char* signature) {
+static struct CallbackInfo* createCallbackByName(JNIEnv* env, jobject obj, const char* method, const char* signature) {
     if (method == NULL)
         return NULL;
     struct CallbackInfo* cbinfo = malloc(sizeof(struct CallbackInfo));
@@ -56,15 +57,25 @@ static struct CallbackInfo* createCallback(JNIEnv* env, jobject obj, jstring met
     cbinfo->obj = (*env)->NewGlobalRef(env, obj);
     jclass thisClass = (*env)->GetObjectClass(env, cbinfo->obj);
     cbinfo->cls = (jclass)(*env)->NewGlobalRef(env, thisClass);
-    const char* methodstr = (*env)->GetStringUTFChars(env, method, NULL);
-    jmethodID methodId = (*env)->GetMethodID(env, cbinfo->cls, methodstr, signature);
-    (*env)->ReleaseStringUTFChars(env, method, methodstr);
+    jmethodID methodId = (*env)->GetMethodID(env, cbinfo->cls, method, signature);
 
     if (methodId == NULL)
         return NULL;
     cbinfo->method = methodId;
     return cbinfo;
 }
+
+static struct CallbackInfo* createCallback(JNIEnv* env, jobject obj, jstring method, const char* signature) {
+    if (method == NULL)
+        return NULL;
+
+    const char* methodstr = (*env)->GetStringUTFChars(env, method, NULL);
+    struct CallbackInfo *cbinfo = createCallbackByName(env, obj, methodstr, signature);
+    (*env)->ReleaseStringUTFChars(env, method, methodstr);
+    return cbinfo;
+}
+
+
 
 static void deleteCallback(struct CallbackInfo* cbinfo) {
     if (cbinfo == NULL)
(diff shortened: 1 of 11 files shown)


ryzhyk:
@mbudiu-vmw, can you please look at the Java/JNI side of this? I am worried about safety of all this JNI magic.
@remysucre, this introduces a new Rust/C API method in `lib/log.rs`

java/ddlogapi.c
```diff
@@ -56,16 +57,26 @@ static struct CallbackInfo* createCallback(JNIEnv* env, jobject obj, jstring met
...
+static struct CallbackInfo* createCallback(JNIEnv* env, jobject obj, jstring method, const char* signature) {
+    if (method == NULL)
+        return NULL;
+
+    const char* methodstr = (*env)->GetStringUTFChars(env, method, NULL);
+    struct CallbackInfo *cbinfo = createCallbackByName(env, obj, methodstr, signature);
+    (*env)->ReleaseStringUTFChars(env, method, methodstr);
+    return cbinfo;
+}
+
```
mbudiu-vmw: three empty lines is often too much

java/ddlogapi.c
```diff
@@ -260,6 +271,40 @@ JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1enable_1cpu_1profiling(
...
+    struct CallbackInfo* cbi = (struct CallbackInfo*)callbackInfo;
+    JNIEnv* env;
+    (*cbi->jvm)->AttachCurrentThreadAsDaemon(cbi->jvm, (void**)&env, NULL);
+    assert(env);
+    jstring jmsg = (*env)->NewStringUTF(env, msg);
+    (*env)->CallVoidMethod(env, cbi->obj, cbi->method, jmsg, (jint) level);
+}
+
+/* This function sets up a new logging callback for the given module and
+ * then deallocates the old `CallbackInfo`, if any.
```
mbudiu-vmw: You should document the return value; it's a handle to a callbackinfo data structure.
A return of 0 seems to be possible even in case of success - when you just wanted to delete the callback.

java/test/SpanTest.java
```diff
@@ -115,7 +115,23 @@ public ContainerSpan() {}
         private int containerSpanTableId;
         private boolean localTables = false;
 
+        /* LOGGING: Module id's for logging purposes.  Must match declarations in
```
mbudiu-vmw: It would be nice if the compiler generated these.
We can probably achieve this using an annotation, e.g., @moduleid.

java/test/run.sh
```diff
@@ -29,10 +29,10 @@ rm -f ../../test/datalog_tests/span_uuid_ddlog/target/release/libspan_uuid_ddlog
 # Compile SpanTest.java
 javac -cp .. SpanTest.java
 # Create a shared library containing all the native code: ddlogapi.c, libspan_uuid_ddlog.a
-${CC} -shared -fPIC -I${JAVA_HOME}/include -I${JAVA_HOME}/include/${JDK_OS} -I../../rust/template ../ddlogapi.c -L../../test/datalog_tests/span_uuid_ddlog/target/release/ -lspan_uuid_ddlog -o libddlogapi.${SHLIBEXT}
+${CC} -shared -fPIC -I${JAVA_HOME}/include -I${JAVA_HOME}/include/${JDK_OS} -I../../rust/template -I../../lib ../ddlogapi.c -L../../test/datalog_tests/span_uuid_ddlog/target/release/ -lspan_uuid_ddlog -o libddlogapi.${SHLIBEXT}
 # Run the java program pointing to the created shared library
 #java -Xcheck:jni -Djava.library.path=. -jar span.jar ../../test/datalog_tests/span_uuid.dat >span_uuid.java.dump
-java -Djava.library.path=. -cp ../ddlogapi.jar:. SpanTest ../../test/datalog_tests/span_uuid.dat > span_uuid.java.dump
+java -Djava.library.path=. -cp ../ddlogapi.jar:. SpanTest ../../test/datalog_tests/span_uuid.dat > span_uuid.java.dump 2> span_uuid.log
```
mbudiu-vmw: You are not checking the contents of span_uuid.log.

test/datalog_tests/span_uuid.dl
```diff
@@ -25,9 +36,13 @@ input relation Source(parent: entid_t, child: entid_t)
 relation Span(entity: uuid_t, tn: tnid_t)
 
 // base case
-Span(entity, tn) :- Binding(entity, tn).
+Span(entity, tn) :- Binding(entity, tn),
+                    trace(mod_SPAN_UUID1(), "Span(entity,tn) :- Binding(.entity=${deref(entity)},.tn=${deref(tn)}))").
```
mbudiu-vmw: this is very nice

java/test/SpanTest.java
ryzhyk: This whole thing is just a library; compiler knows nothing about it.  One could declare those functions `extern` and define them in Rust and with some effort even pull them from Java. But this is probably not worth the effort.

mbudiu-vmw: But the compiler could know about the annotation. We define it.

java/test/run.sh
ryzhyk: I agree it would be good to do this, but the contents of logs is non-deterministic with multiple worker threads, but potentially also other DD-internal evaluation order.

lib/log.rs
```diff
@@ -0,0 +1,74 @@
...
+ * Logging API exposed to the DDlog program.
+ */
+pub fn log_log(module: &log_module_t, level: &log_log_level_t, msg: &String) -> bool
+{
+    if let Some((cb, current_level)) = LOG_CONFIG.read().unwrap().get(&module) {
+        if *level <= *current_level {
+            cb(*level, msg.as_str());
+        }
+    };
+    true
```
remysucre: What is the purpose of returning `bool`? (document this?)

lib/log.rs
```diff
@@ -0,0 +1,74 @@
...
+
+type log_callback_t = Box<dyn Fn(log_log_level_t, &str) + Send + Sync>;
+
+lazy_static! {
+    /* Logger configuration for each module consists of the maximal enabled
+     * log level (messages above this level are ignored) and callback.
+     */
+    static ref LOG_CONFIG: sync::RwLock<collections::HashMap<log_module_t, (log_callback_t, log_log_level_t)>> = {
+        sync::RwLock::new(collections::HashMap::new())
+    };
```
remysucre: FYI it compiles fine w/o the braces. I feel it's slightly more readable that way

lib/log.rs
```diff
@@ -0,0 +1,74 @@
...
+ */
+pub fn log_set_callback(module: log_module_t, cb: Option<log_callback_t>, max_level: log_log_level_t)
+{
+    match cb {
+        Some(cb) => {
+            LOG_CONFIG.write().unwrap().insert(module, (cb, max_level));
+        },
+        None => {
+            LOG_CONFIG.write().unwrap().remove(&module);
+        }
```
remysucre: So passing in a `None` callback "unsets" logging? (document this?)

lib/log.rs
ryzhyk: Thanks! I actually like braces better, but I did make other changes you suggested.

PR #292, opened 2019-07-09
start channel api

remysucre:


remysucre:
> Shouldn't we use the Observer/Observable APIs?

We should! Let me change it.

mbudiu-vmw:
In that case you need at least two APIs, one for the observer, and one for the observable.
A channel may actually implement both, one at each endpoint.

mbudiu-vmw:
http://reactivex.io/documentation/observable.html

d-e-s-o:
Would you mind adding a summary what this Channel API is meant to be used for exactly and why it looks the way it looks (i.e., the design decision made and alternatives discarded)?

It is very hard for an interested bystander to understand its purpose otherwise or follow development in general.

mbudiu-vmw:
The channel will be a mechanism that can connect multiple DDlog computations that run on independent machines. A channel will transport relation updates between these computations. This is still very early experimental design.

remysucre:
Closing to refactor the API.

Issue #287, opened 2019-07-02
`HDDlog::dump_table` and `HDDlog::commit_dump_changes` must return iterators


ryzhyk:

ryzhyk:
The way to do `dump_table()` is probably to return an iterator whose lifetime is bound to that of `HDDlog`.

remysucre:
I tried this and the lifetime seems to be bound to that of `HDDlog.db` and not `HDDlog`. Is there a way to hook those up?

remysucre:
To handle this properly one may need to implement a special iterator for the database, which may complicate the API, thus closing.

PR #297, opened 2019-07-10
Change `HDDlog::commit_dump_changes ` to return the delta DB

remysucre:
This is for #287 . However, we can't return iterators since the lifetime won't work out. I tried refactoring `HDDlog::dump_table`, but since we don't want to move out the DB I don't see a good solution without cloning a lot. Since `dump_table` is not on the critical path now I'm leaving it alone.

### DIFF
--- a/rust/template/main.rs
+++ b/rust/template/main.rs
@@ -49,62 +49,29 @@ fn handle_cmd(hddlog: &HDDlog,
         },
         Command::Commit(record_delta) => {
             // uncomment to enable profiling
-            //PROFILER.lock().unwrap().start("./prof.profile").expect("Couldn't start profiling");
-            if record_delta {
-                match hddlog.transaction_commit_dump_changes() {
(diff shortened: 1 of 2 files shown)


rust/template/main.rs
```diff
@@ -49,21 +49,24 @@ fn handle_cmd(hddlog: &HDDlog,
...
-                    })
-                } else {
-                    None
-                };
-                hddlog.transaction_commit_dump_changes(cb)
+                hddlog.transaction_commit_dump_changes().map(|changes| if print_deltas {
+                    for (table_id, table_data) in changes.as_ref().iter() {
+                        for (val, weight) in table_data.iter() {
+                            debug_assert!(*weight == 1 || *weight == -1);
+                            if current_table != Some(*table_id) {
```
ryzhyk: This code can be simplified since you explicitly iterate over tables now (just print table name on each iteration of the outer loop)

PR #298, opened 2019-07-11
Eliminate triple-buffering in the DDlog API

ryzhyk:
The `HDDlog::apply_updates()` method used to convert array of `*UpdCmd`
to `Vec<Box<UpdCmd>>` and then to `Vec<Update<Value>>` before passing
commands down to `RunningProgram::apply_updates()`.  This increased peak
memory consumption on the `redist` test by 30%.

This commit eliminates the tripple buffering by instead using a
streaming interface based on iterators in both `HDDlog` and `RunningProgram`.
As part of this transformation, we introduce a version of
`HDDlog::apply_updates`, called `HDDlog::apply_valupdates` that takes an
iterator over `Update<Value>`.   It is invoked internally by `apply_updates`
and is also used directly by `main.rs`.  This is a more efficient
interface and should be preferred going forward.  In particular, we will
use it in Java bindings once we generate FlatBuffer
serialization/deserialization directly to/from the `Value` type.

### DIFF
--- a/rust/template/api.rs
+++ b/rust/template/api.rs
@@ -31,7 +31,6 @@ pub struct HDDlog {
 
 /* Public API */
 impl HDDlog {
-    
     pub fn print_err(f: Option<extern "C" fn(msg: *const raw::c_char)>, msg: &str) {
         match f {
             None    => eprintln!("{}", msg),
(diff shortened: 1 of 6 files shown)


ryzhyk:
@remysucre, if you cannot review it on Thu, I'll go ahead and merge it, and you can still review it later.

PR #303, opened 2019-07-16
Optionally disable input caching

ryzhyk:
Do not merge this PR until we've fixed OVN to work with it.

-------------------

Until now, DDlog stored copies of all input relations.  These are
primarily used to filter out duplicate insertions and deletions, as well
as to implement `ddlog_clear_relation()` and
`ddlog_dump_input_snapshot()` APIs.  In applications where these
features are not required, this behavior introduces unnecessary memory
overhead.

This commit introduces a new flag (available in Rust, Java, and C APIs and
in the CLI) that controls caching of input relations.  This is similar
to how we've had a flag to control caching of output tables for a while.
We now have a pair of flags: `store_inputs` and `store_outputs`.

In the CLI, we now have the `--no-store-inputs` flag, which disables
input relation caching and `--no-store-outputs`, which is equivalent to
the old `--no-store` flag.

While we're at it, change CLI behavior so that printing of updates is
disabled by default (no more need for the `--no-print` switch). Here is
a summary of new DDlog CLI flags:

```
DDlog CLI interface.

Options:
  --no-store-outputs    Do not store output relation state.
  --no-store-inputs     Do not store input relation state.
  --no-store-delta      Do not record changes.
  --print               Print changes during commit.
  -w, --workers         The number of worker threads. (default: 4)
  -h, --help            Show this help message.
```

### DIFF
--- a/java/ddlogapi/DDlogAPI.java
+++ b/java/ddlogapi/DDlogAPI.java
@@ -15,7 +15,7 @@ public class DDlogAPI {
     /**
      * The C ddlog API
      */
-    native long ddlog_run(boolean storeData, int workers, String callbackName);
+    native long ddlog_run(boolean storeInputs, boolean storeOutputs, int workers, String callbackName);
     static native int ddlog_record_commands(long hprog, String filename, boolean append);
     static native int ddlog_stop_recording(long hprog, int fd);
     static native int ddlog_dump_input_snapshot(long hprog, String filename, boolean append);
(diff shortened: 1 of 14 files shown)


java/ddlogapi/DDlogAPI.java
```diff
@@ -115,11 +115,11 @@
      *                  many times, on potentially different threads, when the "commit"
      *                  API function is called.
      */
-    public DDlogAPI(int workers, Consumer<DDlogCommand> callback, boolean storeData) {
+    public DDlogAPI(int workers, Consumer<DDlogCommand> callback, boolean storeInputs, boolean storeOutputs) {
```
mbudiu-vmw: You may want to provide another version of this constructor with suitable default values for these arguments.

mbudiu-vmw: It looks like "true, false" is the most common combination.

rust/template/ddlog.h
```diff
@@ -88,7 +103,7 @@ extern table_id ddlog_get_table_id(const char* tname);
  *	- `table`    - table being modified
  *	- `rec`	     - record that has been inserted or deleted
  *	- `polarity` - `true` - record has been added
- *		       `false` - record has been deleted
+ *		           `false` - record has been deleted
```
mbudiu-vmw: indentation seems wrong

ryzhyk:
Closing this for now, as we don't plan to merge anytime soon.

PR #306, opened 2019-07-20
Bug in handling of updates from differential.

ryzhyk:
I wrongly assumed that, because output collections are distinct,
the `inspect()` operator attached to them only sees unit weights.
However, while these collections logically represent distinct records,
they can emerge as updates with varying counts that add up to 0, +1, or
-1.  Thus the entire logic based on representing updates as Booleans was
wrong.

The fix is to use the `consolidate()` operator to get DD to
consolidate all updates to the same record to at most 1 update with
multiplicity of +1 or -1.

This commit also introduces a degree of future-proofness by refactoring
the callback API to take `isize` instead of `bool` weights.

### DIFF
--- a/java/ddlogapi/DDlogAPI.java
+++ b/java/ddlogapi/DDlogAPI.java
@@ -20,7 +20,7 @@ public class DDlogAPI {
     static native int ddlog_stop_recording(long hprog, int fd);
     static native int ddlog_dump_input_snapshot(long hprog, String filename, boolean append);
     native int dump_table(long hprog, int table, String callbackMethod);
-    static native int  ddlog_stop(long hprog, long callbackHandle);
+    static native int ddlog_stop(long hprog, long callbackHandle);
     static native int ddlog_transaction_start(long hprog);
     static native int ddlog_transaction_commit(long hprog);
     native int ddlog_transaction_commit_dump_changes(long hprog, String callbackName);
@@ -123,12 +123,14 @@ public class DDlogAPI {
     }
 
     /// Callback invoked from commit.
-    void onCommit(int tableid, long handle, boolean polarity) {
+    void onCommit(int tableid, long handle, long w) {
         if (this.commitCallback != null) {
-            DDlogCommand.Kind kind = polarity ? DDlogCommand.Kind.Insert : DDlogCommand.Kind.DeleteVal;
-            DDlogRecord record = DDlogRecord.fromSharedHandle(handle);
-            DDlogCommand command = new DDlogCommand(kind, tableid, record);
-            this.commitCallback.accept(command);
+            DDlogCommand.Kind kind = w > 0 ? DDlogCommand.Kind.Insert : DDlogCommand.Kind.DeleteVal;
+            for (long i = 0; i < java.lang.Math.abs(w); i++) {
+                DDlogRecord record = DDlogRecord.fromSharedHandle(handle);
+                DDlogCommand command = new DDlogCommand(kind, tableid, record);
+                this.commitCallback.accept(command);
+            }
         }
     }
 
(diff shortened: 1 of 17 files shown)


remysucre:
This must have been a nasty bug to fix...

src/Language/DifferentialDatalog/Expr.hs
```diff
@@ -336,6 +338,30 @@ exprIsVarOrField' (EField _ e _)   = e
...
+exprIsInjective :: DatalogProgram -> S.Set String -> Expr -> Bool
+exprIsInjective d vs e =
+    S.fromList (exprVars e) == vs &&
+    exprFold (exprIsInjective' d) e
+
+-- No clever analysis here; just the obvious cases.
+exprIsInjective' :: DatalogProgram -> ExprNode Bool -> Bool
+exprIsInjective' _ EVar{}        = True
+exprIsInjective' d EApply{..}    =
+    and exprArgs && (maybe False (exprIsInjective d (S.fromList $ map name funcArgs)) $ funcDef)
```
remysucre: Is this check also applied to the function body in a function definition? If so this case is redundant, and may trigger non-termination if DDlog is to support recursive functions in the future.

rust/template/api.rs
```diff
@@ -625,7 +626,7 @@ pub unsafe extern "C" fn ddlog_transaction_commit_dump_changes(
             if let Some(f) = f {
                 for (table_id, table_data) in changes.as_ref().iter() {
                     for (val, weight) in table_data.iter() {
-                        debug_assert!(*weight == 1 || *weight == -1);
+                        assert!(*weight == 1 || *weight == -1);
```
remysucre: Is it still possible for the weight to be 0? If not we should clarify it in the c wrapper in `ddlog.h`

java/ddlogapi/DDlogAPI.java
```diff
@@ -123,12 +123,14 @@ public DDlogAPI(int workers, Consumer<DDlogCommand> callback, boolean storeData)
...
     /// Callback invoked from commit.
-    void onCommit(int tableid, long handle, boolean polarity) {
+    void onCommit(int tableid, long handle, long w) {
         if (this.commitCallback != null) {
-            DDlogCommand.Kind kind = polarity ? DDlogCommand.Kind.Insert : DDlogCommand.Kind.DeleteVal;
-            DDlogRecord record = DDlogRecord.fromSharedHandle(handle);
-            DDlogCommand command = new DDlogCommand(kind, tableid, record);
-            this.commitCallback.accept(command);
+            DDlogCommand.Kind kind = w > 0 ? DDlogCommand.Kind.Insert : DDlogCommand.Kind.DeleteVal;
+            for (long i = 0; i < java.lang.Math.abs(w); i++) {
```
remysucre: Do you intend this loop to execute only once when `w` is 1 or -1? If so I'd find an if statement to be more readable than a for loop.

rust/template/api.rs
remysucre: Ah I see the C wrapper takes a bool. In general, when are the weights allow to be 0? Any time before `consolidate` is called?

ryzhyk: This particular code runs after all changes have been accumulated, at which point weights must be 1 or -1 since output relations are distinct (regardless of consolidate). This is why the `commit_dump_changes` callback takes `bool`.

src/Language/DifferentialDatalog/Expr.hs
ryzhyk: This is a good point. We would have to track visited functions to terminate the recursion.

java/ddlogapi/DDlogAPI.java
ryzhyk: This is designed to work even without `consolidate` or `distinct`, in which case `w` can be an arbitrary value; hence the loop...

src/Language/DifferentialDatalog/Expr.hs
ryzhyk: In fact, reasoning about recursive functions is more subtle. I will just add a comment saying that, once we allow recursion (and have the infrastructure to check that the function is recursive), `exprIsInjective` should return `False` for such functions.

PR #313, opened 2019-08-08
Script to install dependencies (except Java)

mbudiu-vmw:
I also changed Spec.hs to preserve the PATH, otherwise it won't find javac if it is installed in a non-system directory.


install-dependencies.sh
```diff
@@ -0,0 +1,31 @@
...
+echo "Installing dependencies for DDlog"
+echo "This script should be invoked with '. ./install-dependencies.sh' to set up the environment properly"
+
+case "$OSTYPE" in
+    linux*) ;;
+    *) echo "Unhandled operating system $OSTYPE"; exit 1;;
+esac
+
+
+echo "Installing Haskell"
```
ryzhyk: There is a script to install Haskell in `.travis/install-stack.sh`, and it works on both Linux and MacOS.  Shall we call that instead?

install-dependencies.sh
```diff
@@ -0,0 +1,31 @@
...
+curl -sSL https://get.haskellstack.org/ | sh
+
+echo "Installing Rust"
+curl https://sh.rustup.rs -sSf | sh -s - -y
+export PATH=$HOME/.cargo/bin:$PATH
+
+echo "Installing Flatbuf"
```
ryzhyk: There is a script to install FlatBuffers in `.travis/install-flatbuf.sh`.  It works for both MacOS and Linux. Only problem is, it installs in the home directory (which makes sense on Travis, but not necessarily on the developer's machine). Your script, on the other hand, installs in the `differential-datalog` directory. I can probably change the Travis script to do the same, so we only have one script to install Travis.

test/Spec.hs
```diff
@@ -192,9 +192,11 @@ compilerTest progress java fname cli_args crate_types = do
         compileFlatBufferBindings prog specname (dir </> rustProjectDir specname)
     -- compile generated Java code
     classpath <- (maybe "" (":" ++ )) <$> lookupEnv "CLASSPATH"
+    path <- (maybe "" (":" ++ )) <$> lookupEnv "PATH"
```
ryzhyk: Can be just `path <- (maybe "" id) <$> lookupEnv "PATH"`, as we are not appending anything to it.

install-dependencies.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# Script that installs the dependencies needed to build and run Differential Dataflow
+
+echo "Installing dependencies for DDlog"
+echo "This script should be invoked with '. ./install-dependencies.sh' to set up the environment properly"
+
+case "$OSTYPE" in
```
ryzhyk: We can totally auto-install dependencies on MacOS (see below)

README.md
```diff
@@ -67,7 +67,11 @@ Datalog 2.0, Philadelphia, PA, June 4-5, 2019.
 
 ## Installing from sources
 
+To install the required dependencies for buidling DDlog run `. ./install-dependencies.sh`
```
ryzhyk: The comment should mention that manual installation is required if the user has Haskell or Rust or FlatBuffers already installed and want to keep their existing versions. Alternatively we need more complex logic to handle existing compilers.

install-dependencies.sh
mbudiu-vmw: The travis script depends on a lot of Travis-related stuff. 
This command-line is also very short, much shorter than your script.
I should investigate how to do it for the Mac, but I can't really test.

install-dependencies.sh
mbudiu-vmw: There you are worrying about travis retrying too.
It's short enough that some duplication should be OK.

mbudiu-vmw: Your scripts for Travis are very different from the manual installation instructions you are giving in the README.

install-dependencies.sh
ryzhyk: There's nothing travis-related in the script it literally just checks OS type and runs `curl`. Despite its name, `travis_retry` is declared in the same file and simply retries the command several times.

tools/install-flatbuf.sh
```diff
@@ -2,25 +2,26 @@
...
+    echo "Installing Flatbuf"
+    retry fetch_flatbuf
     cd flatbuffers
     cmake -G "Unix Makefiles"
     make
+    cd ..
 fi
 
-ln -f -s $HOME/flatbuffers/flatc ~/.local/bin/flatc
+ln -f -s ./flatbuffers/flatc ~/.local/bin/flatc
```
ryzhyk: You don't want to use relative paths in symlinks; should probably be:
```
ln -f -s `pwd`/flatbuffers/flatc ~/.local/bin/flatc
```

.travis.yml
```diff
@@ -61,7 +61,7 @@ before_install:
 - export PATH=$HOME/.local/bin:$PATH
 - ./tools/install-stack.sh
 - ./tools/install-flatbuf.sh
-- export CLASSPATH=$CLASSPATH:$HOME/flatbuffers/java
+- export CLASSPATH=$CLASSPATH:./flatbuffers/java
```
ryzhyk: Likewise, we should not use relative `$CLASSPATH`:
```
export CLASSPATH=$CLASSPATH:`pwd`/flatbuffers/java
```

ryzhyk:
Also there's a build failure:
```
/Users/travis/build/vmware/differential-datalog/test/Spec.hs:196:5: error: [-Wname-shadowing, -Werror=name-shadowing]

    This binding for ‘path’ shadows the existing binding
      imported from ‘Test.HUnit’ at test/Spec.hs:34:1-17
      (and originally defined in ‘Test.HUnit.Base’)
    |
196 |     path <- (maybe "" id) <$> lookupEnv "PATH"
    |     ^^^^
```

.travis.yml
ryzhyk: Do you want to fix this one too (otherwise, I can do it; just want to be sure you're ok with the suggestion)?

mbudiu-vmw: I actually fixed this one too, but I mistyped the file name when I committed...
Sorry about that. Pushed it now.

PR #334, opened 2019-09-04
Remove executable bit from files in doc/datalog2.0-workshop/

d-e-s-o:
The files in the doc/datalog2.0-workshop/ directory are data files only
and not meant to be executed -- executing them makes no sense. So remove
the executable bit currently set on almost all of them.


mbudiu-vmw:
This happened because the files were shared on a Windows filesystem mounted in a Linux vm.

PR #336, opened 2019-09-04
Add javac as a prerequisite

d-e-s-o:
It seems that at least for running the tests (and following the tutorial, which
includes a test) `javac` is a requirement, but it is not listed anywhere as
such.
With this change we drop a note about that into the section about prerequisites.
I am not aware of any particular JDK being required, so I kept the text generic.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,9 @@ curl https://sh.rustup.rs -sSf | sh
 to `~/.profile`, so that it becomes effective at the next login attempt.  To configure your current
 shell run `source $HOME/.cargo/env`.
 
+If you intend to run the test suite or want to generate [Java bindings](doc/java_api.md), you will
+also need a Java installation containing `javac`.
+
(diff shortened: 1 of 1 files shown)


d-e-s-o:
The risk then could be that we silently ignore tests just because `javac` not installed, though. So I am not really sure if that's an improvement. In my opinion that should be explicit one way or another.

d-e-s-o:
It's also not just java tests that require `javac`. The compile tests as executed as part of each (?) test seemingly do:
```bash
$ CLASSPATH="${CLASSPATH:-}:/home/danielmuller/local/opt/flatbuffers-1.11.0/java" stack test --ta '-p playpen'
differential-datalog-0.1.0.0: test (suite: differential-datalog-test, args: -p playpen)

ddlog tests
  parser tests
    playpen: OK (0.04s)
  compiler tests
    playpen: FAIL (0.10s)
      javac failed with exit code ExitFailure 1
```

ryzhyk:
We could print a warning about Java tests not running.  In any case, your change correctly documents current setup.

PR #343, opened 2019-09-05
Run CI/CD pipeline with minimum Rust version we claim to support

d-e-s-o:
Travis by default tests the project with the most recent stable version of
Rust. That is not optimal because at any point in time may we inadvertently
use a feature provided by a newer Rust version and, hence, break the contract
we guarantee by virtue of stating what versions we support.
To fix this problem, let's run our pipeline with the minimum Rust version we
claim to support, in order to identify issues. Newer versions of Rust are
supposed to be backwards compatible and tangentially developers are more
likely to use those when running the tests locally.
As a bonus, also add a badge for the minimum Rust version to the README.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,12 @@ osx_image: xcode9.3
 
 # Do not choose a language; we provide our own build tools.
 language: rust
+rust:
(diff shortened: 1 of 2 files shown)


ryzhyk:
Makes sense!

PR #344, opened 2019-09-06
Fix cmd_parser tests

d-e-s-o:
The unit tests for the cmd_parser crate are neither compiling nor (once
the compile problems are fixed) passing. This change fixes them.

### DIFF
--- a/rust/template/cmd_parser/parse.rs
+++ b/rust/template/cmd_parser/parse.rs
@@ -65,7 +65,10 @@ named!(pub parse_command<&[u8], Command>,
                             apply!(sym,";")         >>
                             (Command::Commit(delta.is_some())))                                 |
                   do_parse!(apply!(sym,"timestamp") >> apply!(sym,";") >> (Command::Timestamp)) |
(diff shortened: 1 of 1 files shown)


rust/template/cmd_parser/parse.rs
```diff
@@ -65,7 +65,10 @@ named!(pub parse_command<&[u8], Command>,
                             apply!(sym,";")         >>
                             (Command::Commit(delta.is_some())))                                 |
                   do_parse!(apply!(sym,"timestamp") >> apply!(sym,";") >> (Command::Timestamp)) |
-                  do_parse!(apply!(sym,"#")         >> take_until!("\n") >> (Command::Comment)) |
+                  do_parse!(apply!(sym,"#")         >>
+                            take_until!("\n")       >>
+                            apply!(sym,"\n")        >>
```
ryzhyk: Will this work if the comment is on the last line of the command file?

d-e-s-o: Fair point. Based on my experimentation that is not a problem. I had a comment on the last line (and a bunch of actual commands before) and then tested with and without a trailing newline symbol. Both cases behaved identically.

PR #346, opened 2019-09-09
run-souffle-tests.py: Return a list on all paths

d-e-s-o:
run-souffle-tests.py returns an implicit None on two error (?) paths which
callers simply don't expect (see [job 582855296][job-582855296] for an
instance of this problem). While the intention of returning from those
paths is not 100% clear to me (wouldn't an exception be more
appropriate?), let's at least fix the mixed return type issue.

[job-582855296]: https://travis-ci.org/d-e-s-o/differential-datalog/jobs/582855296

### DIFF
--- a/test/run-souffle-tests.py
+++ b/test/run-souffle-tests.py
@@ -177,10 +177,10 @@ def run_remote_tests():
     for tg in testgroups:
         code, _ = run_command(["mkdir", "-p", tg])
         if code != 0:
-            return
+            return []
         code, dirs = run_command(["svn", "ls", url + tg])
         if code != 0:
-            return
+            return []
         for test in dirs.split("/\n"):
             directory = tg + "/" + test
             if not should_run(test):


ryzhyk:
Requested @mbudiu-vmw 's review.

The `run-souffle-tests` script fails occasionally as it pulls test cases directly from the `souffle` svn.

@mbudiu-vmw , is there a way to suppy a list of tests to run (instead of a range) to the script, so that we can have a more stable Travis configuration?

mbudiu-vmw:
Whatever arguments you supply without flags will be used as the test names to run.
But I think you can't specify directory names... and some file names are canonicalized

ryzhyk:
@mbudiu-vmw , any chance you could replace `./run-souffle-tests.py -s 0 -r 17` in `travis.yml` to be in that format? Thanks!

mbudiu-vmw:
https://github.com/vmware/differential-datalog/pull/349

PR #347, opened 2019-09-09
Topic/rustfmt

d-e-s-o:
This pull request formats the Rust crates we have in the repository using `rustfmt` and enforces automated format checking by virtue of running it as part of the Travis CI/CD pipeline.
Right now we use the default `rustfmt` configuration, which is convenient in many ways. But if you feel that something looks off, we may be able to override the default to adjust the result.

----

Apologies for the large change. I don't think that it is necessary to check every formatting change in detail, and except for an introduction of a private typedef, no manual changes have been made.


ryzhyk:
Awesome!

d-e-s-o:
Sorry, I did not mean to update the pull request with `clippy` stuff. That is still in the works. Let me not reuse that branch.

PR #357, opened 2019-09-13
Topic/interactive

d-e-s-o:
When using the CLI that is provided with each generated ddlog program,
some errors are effectively ignored (in the sense that they don't
terminate the program; arguably sensible behavior), while others cause
program exit. E.g.,
```
  >> list;
  Invalid input: list;,
  >> hello;
  Invalid input: hello;,
```
  (we can just continue)

But:
```
  >> dump Left;
  Error: Unknown output relation Left
  $
```
  (we exit on this path)

The latter behavior can be very frustrating, as it can mean that an
interactive session into which data was fed manually can accidentally be
terminated just by virtue of misspelling a relation, for example.

This pull request addresses the problem.

### DIFF
--- a/rust/template/cmd_parser/lib.rs
+++ b/rust/template/cmd_parser/lib.rs
@@ -32,17 +32,13 @@ where
 {
     let mut buf: Vec<u8> = Vec::new();
 
-    let mut input = if unsafe { libc::isatty(libc::STDIN_FILENO as i32) } != 0 {
+    let istty = unsafe { libc::isatty(libc::STDIN_FILENO as i32) } != 0;
+    let mut input = if istty {
         let mut rl = Editor::<()>::new();
         let _ = rl.load_history(HISTORY_FILE);
         Input::TTY(rl)
     } else {
         Input::Pipe(BufReader::new(io::stdin()))
-    };
-
-    let istty = match &input {
-        Input::TTY(_) => true,
-        _ => false,
     };
 
     loop {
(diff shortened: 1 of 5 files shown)


mbudiu-vmw:
we could consider adding some options to stop on error as well. useful for scripting.
similar to bash `set -e`

ryzhyk:
Thanks!  And there are even tests!  Can you please double-check that feeding a dat file with an error will cause it to fail immediately?  I can see that the code takes care of this case, but just to be sure....

@mbudiu-vmw, the behavior @d-e-s-o fixed here only occurs when running interactively (i.e., `stdin` is connected to a TTY). In batch mode, it should exit immediately.  I agree that making this behavior  configurable could be useful, but this seems like a reasonable default.

d-e-s-o:
> Thanks! And there are even tests! Can you please double-check that feeding a dat file with an error will cause it to fail immediately? I can see that the code takes care of this case, but just to be sure....

Yep, still works:
```bash
$ /tmp/rust-build-artifacts/home/danielmuller/local/src/differential-datalog/test/datalog_tests/playpen_ddlog/target/debug/playpen_cli < test.dat
FIRST TEST
Error: Unknown output relation InvalidRelation
$ cat test.dat
echo FIRST TEST;
dump InvalidRelation;
echo SECOND TEST;
```

> @mbudiu-vmw, the behavior @d-e-s-o fixed here only occurs when running interactively (i.e., `stdin` is connected to a TTY). In batch mode, it should exit immediately. I agree that making this behavior configurable could be useful, but this seems like a reasonable default.

If we make it configurable it would probably make sense to restructure the code a bit. Then we could test it much more easily (the fact that we currently read from a file descriptor and can't just work with a trait implementation makes for a bit awkward tests).

PR #359, opened 2019-09-13
 Deserialization side of the FlatBuffers API 

ryzhyk:
Details in the commit message.


java/ddlogapi.c
```diff
@@ -220,6 +220,52 @@ JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1transaction_1commit_1dump_1
...
+JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1transaction_1commit_1dump_1changes_1to_1flatbuf(
+    JNIEnv * env, jobject obj, jlong handle, jobject fbdescr)
+{
+    unsigned char *buf_addr = NULL;
+    size_t buf_size         = 0;
+    size_t buf_capacity     = 0;
+    size_t buf_offset       = 0;
+
+    jclass cls = (*env)->FindClass(env, "ddlogapi/DDlogAPI$FlatBufDescr");
+    if(cls == NULL) {
```
mbudiu-vmw: missing a space

java/ddlogapi.c
```diff
@@ -220,6 +220,52 @@ JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1transaction_1commit_1dump_1
...
+            (ddlog_prog)handle,
+            &buf_addr,
+            &buf_size,
+            &buf_capacity,
+            &buf_offset);
+
+    if (res < 0) {
+        return res;
+    }
+
```
mbudiu-vmw: why so many empty lines

java/ddlogapi.c
```diff
@@ -220,6 +220,52 @@ JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1transaction_1commit_1dump_1
     return (jint)result;
 }
 
+JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1transaction_1commit_1dump_1changes_1to_1flatbuf(
+    JNIEnv * env, jobject obj, jlong handle, jobject fbdescr)
+{
```
mbudiu-vmw: the rest of the file is indented with the curly brace at the end of the previous line

java/ddlogapi.c
```diff
@@ -220,6 +220,52 @@ JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1transaction_1commit_1dump_1
...
+            direct_buf, (jlong)buf_size, (jlong)buf_offset);
+
+    return 0;
+}
+
+JNIEXPORT void JNICALL Java_ddlogapi_DDlogAPI_ddlog_1flatbuf_1free(
+    JNIEnv * env, jobject obj, jobject buf, jlong size, jlong offset)
+{
+    ddlog_flatbuf_free(
+            (unsigned char*)(((size_t)(*env)->GetDirectBufferAddress(env, buf)) - offset),
```
mbudiu-vmw: I would do this in three statements, checking the return values for the JNI functions as well.
E.g., GetDirectBufferAddress can return null.

java/ddlogapi/DDlogAPI.java
```diff
@@ -190,27 +192,54 @@ public int commit() {
...
+    // meant to be invoked directly by user code; they are only used by the
+    // autogenerated FlatBuffer API code.
+    public static class FlatBufDescr {
+        public FlatBufDescr() {
+            this.buf = null;
+            this.size = 0;
+            this.offset = 0;
+        }
+        public void set(ByteBuffer buf, long size, long offset) {
+            this.buf = buf;
```
mbudiu-vmw: Can you make some sanity checks, e.g., capacity is large enough?

java/ddlogapi/DDlogCommand.java
```diff
@@ -1,59 +1,47 @@
...
+ * </p>
+ *
+ * <p>
+ * This interface is used to represent both updates to DDlog input relations
+ * issued by the client and output updates generated by DDlog on transaction commit.
+ * </p>
+ *
+ * <p>
+ * The interface is parameterized by record type, which can be either DDlogRecord,
+ * used to represent any DDlog datatype in a generic way, or Object, when
```
mbudiu-vmw: I don't understand this; value() already returns a value of type R.

java/ddlogapi/DDlogCommand.java
```diff
@@ -1,59 +1,47 @@
...
+ * by the record's primary key.
+ * </p>
+ *
+ * <p>
+ * This interface is used to represent both updates to DDlog input relations
+ * issued by the client and output updates generated by DDlog on transaction commit.
+ * </p>
+ *
+ * <p>
+ * The interface is parameterized by record type, which can be either DDlogRecord,
```
mbudiu-vmw: You should write two phrases. This represents a command. The command is parameterized with a type, which can be...

java/test_flatbuf1/Test.java
```diff
@@ -16,16 +17,246 @@
...
+            throw new IllegalArgumentException("Invalid Many value " + m.toString());
+        }
+    }
+
+    void onCommit(DDlogCommand<Object> command) {
+        int relid = command.relid();
+        switch (relid) {
+            // output relation BO(b: bool)
+            case flatbufTestRelation.BO: {
+                BOReader v = (BOReader)command.value();
```
mbudiu-vmw: Alternatively, maybe (if type erasure allows) you could define a method in DDlogCommand: 
`<T> T as() { return (DDlogCommand<T>)this; }`
and then write `command.as<BOReader>().value();`

java/test_flatbuf1/Test.java
```diff
@@ -16,16 +17,246 @@
...
+            case flatbufTestRelation.LO: {
+                Tuple3__bool__bit_8___stringReader v = (Tuple3__bool__bit_8___stringReader)command.value();
+                System.out.println("From " + relid + " " + command.kind() + " " + printTuple(v));
+                break;
+            }
+
+            // output relation L0O(a: bool, b: bit<8>, s: string)
+            case flatbufTestRelation.L0O: {
+                L0OReader v = (L0OReader)command.value();
+                System.out.println("From " + relid + " " + command.kind() + " L0O{" + v.a() + ", " + v.b() + ", \"" + v.s() + "\"}");
```
mbudiu-vmw: Is this correct if s has quotes inside? Or do you need to escape some characters?

mbudiu-vmw: In fact, the same bug exists in DDlogRecord.java already.
We have to find out the rules for printing strings from Rust.

java/test_flatbuf1/Test.java
```diff
@@ -16,16 +17,246 @@
...
+            return "B{" + ((ManyReader.B)m).b() + "}";
+        } else if (m instanceof ManyReader.D) {
+            return "D{" + printTuple(((ManyReader.D)m).t()) + "}";
+        } else {
+            throw new IllegalArgumentException("Invalid Many value " + m.toString());
+        }
+    }
+
+    void onCommit(DDlogCommand<Object> command) {
+        int relid = command.relid();
```
mbudiu-vmw: One other way we can also test this is to print each record twice: once using DDlogRecord.toString(), and once using your method. Then we should check that all outputs appear twice.

java/test_flatbuf1/Test.java
```diff
@@ -48,32 +279,35 @@ void run() {
...
-            builder.insert_MI(v);
+            Boolean[] v = new Boolean[] {true, false, true};
+            builder.insert_MI(Arrays.asList(v));
         }
         {
-            Tuple3__bool__bit_8___string[] vec = {
-                builder.create_Tuple3__bool__bit_8___string(true, (byte)-1, "check"),
-                builder.create_Tuple3__bool__bit_8___string(false, (byte)1, "fails")
-            };
+            ArrayList<Tuple3__bool__bit_8___stringWriter> vec = new ArrayList<Tuple3__bool__bit_8___stringWriter>(2);
```
mbudiu-vmw: This API is more wasteful: it grows the ArrayList dynamically.
I think you should probably support both List and [].

java/test_flatbuf1/test.dump.expected
```diff
@@ -15,24 +15,24 @@ From 13 Insert HO{773153186440202938}
...
 From 37 Insert SO{C{"s"}}
 From 39 Insert TO{std_Some{10}}
 From 39 Insert TO{std_None{}}
 From 41 Insert A{"a"}
 From 41 Insert B{false}
 From 41 Insert D{(false, 2, "zz")}
-From 43 Insert VO{false,A{"aa"}}
-From 43 Insert VO{false,D{(false, 2, "string")}}
-From 43 Insert VO{true,B{false}}
+From 43 Insert VO{false, A{"aa"}
```
mbudiu-vmw: missing some closed braces

java/test_flatbuf1/test.dump.expected
```diff
@@ -15,24 +15,24 @@ From 13 Insert HO{773153186440202938}
 From 15 Insert IO{7}
 From 17 Insert JO{(true, 10, "string")}
 From 19 Insert KO{(false, 9, "text")}
-From 21 Insert L0O{false,254,"else"}
-From 23 Insert (true, 255, "something")
+From 21 Insert L0O{false, -2, "else"}
```
mbudiu-vmw: L0O has a `bit<8>` value in the second field. 
Do you expect it to be printed as a negative number?
If this is the intended behavior we should document it. In particular we have to document how each DDlog type is mapped to a Java type.

rust/template/ddlog.h
```diff
@@ -232,12 +233,38 @@ extern int ddlog_transaction_commit(ddlog_prog hprog);
...
+                   bool polarity),
+        uintptr_t cb_arg);
+
+/*
+ * Same as `ddlog_transaction_commit_dump_changes`, but serializes changes to a
+ * FlatBuffer.  On success, returns pointer to FlatBuffer, size and capacity of
+ * the buffer, and offset where valid data starts inside the buffer.
+ */
+extern int
+ddlog_transaction_commit_dump_changes_to_flatbuf(
```
mbudiu-vmw: Why would the offset be not zero?
You should probably say that the buffer has to be deallocated by ddlog_flatbuf_free.

src/Language/DifferentialDatalog/FlatBuffer.hs
```diff
@@ -494,19 +517,38 @@ jFBType nested x =
...
                            -> if nested
                                  then "int"
                                  else "int []"
-        t'                 -> error $ "FlatBuffer.jConvType: unsupported type " ++ show t'
+        t'                 -> error $ "FlatBuffer.jFBWriteType: unsupported type " ++ show t'
+
+-- Java types used in the FlafBuffers-generated _de-serialization_ API
+jFBReadType :: (WithType a, ?d::DatalogProgram, ?prog_name::String) => a -> Doc
+jFBReadType x =
+    case typeNormalizeForFlatBuf x of
```
mbudiu-vmw: This should be documented somewhere.

src/Language/DifferentialDatalog/FlatBuffer.hs
```diff
@@ -525,19 +567,32 @@ jConvType x =
...
+                           -> fbTupleName typeTupArgs <> "Reader"
+                   | rw == Write
+                           -> fbTupleName typeTupArgs <> "Writer"
+        TUser{..}  | rw == Read
+                           -> fbStructName typeName typeArgs <> "Reader"
+                   | rw == Write
+                           -> fbStructName typeName typeArgs <> "Writer"
         TOpaque{typeArgs = [elemType], ..} | elem typeName sET_TYPES
-                           -> jConvType elemType <> "[]"
+                           -> "java.util.List<" <> jConvObjType rw elemType <> ">"
```
mbudiu-vmw: Can you keep the other one as well?

src/Language/DifferentialDatalog/FlatBuffer.hs
```diff
@@ -678,21 +738,38 @@ jConvCreateTable t cons =
...
+
 -- Create builder class with methods to create FlatBuffer-backed instances of DDlog
 -- types.
-mkJavaBuilder :: (?d::DatalogProgram, ?prog_name::String) => String -> Doc
-mkJavaBuilder prog_name =
+mkJavaRelEnum :: (?d::DatalogProgram, ?prog_name::String) => (FilePath, Doc)
+mkJavaRelEnum = ("ddlog" </> ?prog_name </> relEnum <.> "java",
     "// Automatically generated by the DDlog compiler."                 $$
-    "package ddlog." <> pp prog_name <> ";"                             $$
+    "package ddlog." <> pp ?prog_name <> ";"                            $$
```
mbudiu-vmw: Maybe you can append the source program name and the data of the generation?

mbudiu-vmw: Maybe the date is a bad idea, since it will cause unnecessary rebuilds

src/Language/DifferentialDatalog/FlatBuffer.hs
```diff
@@ -799,51 +871,282 @@ mkJavaBuilder prog_name =
...
+                     -> do_read e
+                     | otherwise
+                     -> jReadField (nesting+1) (FBField (typeTableName t) "v") e t
+         -- Anything that's not a union or array can be stored in an array; unions and arrays must be
+         -- wrapped in a table first.
+         FBArray     | typeHasUniqueConstructor t && not (typeIsVector t)
+                     -> do_read e
+                     | otherwise
+                     -> jReadField (nesting+1) (FBField (typeTableName t) "v") e t
+         -- Vectors are representedFlatBuffers has a special method for embedding vectors in a buffer
```
mbudiu-vmw: typo in comment

src/Language/DifferentialDatalog/FlatBuffer.hs
```diff
@@ -799,51 +871,282 @@ mkJavaBuilder prog_name =
...
+                      TInt{}             -> bigint
+                      TString{}          -> e'
+                      TBit{..} | typeWidth <= 64
+                                         -> (parens $ jConvTypeR t) <> e'
+                      TBit{..}           -> biguint
+                      TSigned{..} | typeWidth <= 64
+                                         -> (parens $ jConvTypeR t) <> e'
+                      TSigned{..}        -> bigint
+                      TTuple{..}         -> "new" <+> jConvTypeR t <> parens e'
+                      TUser{..} |typeHasUniqueConstructor t
```
mbudiu-vmw: missing a space

src/Language/DifferentialDatalog/FlatBuffer.hs
```diff
@@ -799,51 +871,282 @@ mkJavaBuilder prog_name =
...
+                                         -> "new" <+> jConvTypeR t <> parens e'
+                                | otherwise
+                                         -> jConvTypeR t <> ".init" <> parens e'
+                      TOpaque{typeArgs = [elemType], ..} | elem typeName sET_TYPES ->
+                          let elem_type = jConvObjTypeR elemType
+                              ltype = "java.util.List<" <> elem_type <> ">"
+                              -- Add nesting depth to local variables inside lambda so that they don't
+                              -- clash with variables from containing scopes.
+                              vlen = "len" <> pp nesting
```
mbudiu-vmw: this is a new scope, it shouldn't be a problem I would think - unless you want to refer to outside variables.

java/ddlogapi/DDlogAPI.java
ryzhyk: This class only exists because you cannot return a tuple in Java.  It does not implement any useful abstraction and is not meant to be used directly by the client (it does have to be public, unfortunately, because auto-generated Java code that uses this API lives in a different package). I don't think it needs to be complicated with sanity checks.

java/ddlogapi/DDlogCommand.java
ryzhyk: Sorry I don't understand this comment. `R` can be either `DDlogRecord` or `Object`.

java/test_flatbuf1/Test.java
ryzhyk: I think this is fine for the purposes of this test, as we are not trying to generate something that parses as a DDlog or Java string literal. But yes, in `DDlogRecord.java` it is a bug. Also, strangely I couldn't find a standard library for printing strings with escaping in Java.

java/test_flatbuf1/Test.java
ryzhyk: Quick test seems to show that this works. But what's the advantage of this?  It does not seem to be any safer than just directly casting the record type.

java/test_flatbuf1/Test.java
ryzhyk: `ArrayList` can be pre-allocated, which is what I do in the above code. I had to switch from arrays to `List`'s because I did not find a function type in Java that returns an array (I need that internally to construct a closure that returns a vector).

rust/template/ddlog.h
ryzhyk: It's just how FlatBuffers work. The offset is almost never zero. Not sure why.

src/Language/DifferentialDatalog/FlatBuffer.hs
ryzhyk: Not easily, because Java stinks.

ryzhyk:
Closing this PR; will re-submit with fixes.

java/test_flatbuf1/Test.java
ryzhyk: done in the new PR

src/Language/DifferentialDatalog/FlatBuffer.hs
ryzhyk: yep, the next PR will include some documentation.

PR #363, opened 2019-09-16
Consume githash in version 0.1.3.3

### LINKED ISSUE #345: ddlog fails to build in git-worktree environment
Not an urgent issue, but ddlog fails to build when in a git-worktree(1) environment. The problem we hit is that we use `githash` for inferring the SHA-1 we are currently at but `githash` does not support git-worktree. I've filed [an upstream issue](https://github.com/snoyberg/githash/issues/13) to get this fixed, but opening one here as well to track the problem and fix if I can get to it.

d-e-s-o:
differential-datalog fails to build in a git-worktree(1) environment.
The reason is that githash, the package we use for inferring the commit
we built at in order to embed it into the binary, does not work with
git-worktree style repositories.
This functionality got [added upstream][githash-14] and version 0.1.3.3
now includes support for it. This change adds an explicit constraint to
ensure we consume this or a higher version. This change fixes #345.

[githash-14]: https://github.com/snoyberg/githash/pull/14

### DIFF
--- a/stack.yaml
+++ b/stack.yaml
@@ -41,6 +41,11 @@ packages:
 # (e.g., acme-missiles-0.3)
 # extra-deps: []
 
+extra-deps:
+# We need githash 0.1.3.3 in order to have git-worktree(1) support. It
+# needs to be whitelisted here until we upgrade our resolver.
+- githash-0.1.3.3
+
 # Override default flag values for local packages and extra-deps
 # flags: {}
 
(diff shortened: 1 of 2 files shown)


stack.yaml
```diff
@@ -41,6 +41,11 @@ packages:
 # (e.g., acme-missiles-0.3)
 # extra-deps: []
 
+extra-deps:
+# We need githash 0.1.3.3 in order to have git-worktree(1) support. It
```
ryzhyk: Happy to upgrade the resolver as soon as the fixed version makes it into an LTS.

PR #364, opened 2019-09-17
Java API cleanup

ryzhyk:
See commit messages


java/ddlogapi.c
```diff
@@ -13,6 +13,74 @@
 #include "ddlog.h"
 #include "ddlog_log.h"
 
+/* Error message returned by DDlog. */
+_Thread_local char* err_msg = NULL;
```
mbudiu-vmw: Is this standard? I could not find documentation for it on the web.

java/ddlogapi.c
```diff
@@ -13,6 +13,74 @@
...
+  return size;
+}
+
+/* 
+ * Callback invoked by DDlog on error.
+ *
+ * Record error message in a thread-local variable so that it can later be
+ * wrapped in an exception.
+ */
+static void eprintln(const char* msg) {
```
mbudiu-vmw: I hope this is race-free...

java/ddlogapi.c
```diff
@@ -13,6 +13,74 @@
...
+    va_list ap;
+    va_start(ap, fmt);
+    int nbytes = _vasprintf(&msg, fmt, ap);
+    va_end(ap);
+
+    if (nbytes >= 0) {
+        (*env)->ThrowNew(env, (*env)->FindClass(env, "java/io/IOException"), msg);
+        free(msg);
+    } else {
+        (*env)->ThrowNew(env, (*env)->FindClass(env, "java/io/IOException"), "Unable to format error message");
```
mbudiu-vmw: This should not happen, but if it does then the error will be very confusing.
Perhaps you want to say "exception occurred, but I could not show you the actual error message."

java/ddlogapi.c
```diff
@@ -203,24 +291,30 @@ void commit_dump_callback(void* callbackInfo, table_id tableid, const ddlog_reco
...
+        if (ddlog_transaction_commit_dump_changes((ddlog_prog)handle, NULL, 0) < 0) {
+            throwDDlogException(env);
+            return;
+        }
+    }
 
     struct CallbackInfo* cbinfo = createCallback(env, obj, callback, "(IJZ)V");
     if (cbinfo == NULL)
-        return -1;
+        return;
```
mbudiu-vmw: why not throw here?

java/ddlogapi.c
```diff
@@ -203,24 +291,30 @@ void commit_dump_callback(void* callbackInfo, table_id tableid, const ddlog_reco
...
 
-    int result = ddlog_transaction_commit_dump_changes((ddlog_prog)handle,
-                                                       commit_dump_callback,
-                                                       (uintptr_t)cbinfo);
+    if (ddlog_transaction_commit_dump_changes(
+                (ddlog_prog)handle,
+                commit_dump_callback,
+                (uintptr_t)cbinfo) < 0) {
+        throwDDlogException(env);
```
mbudiu-vmw: How about a helper function check(env, exitCode) which throws when the exitCode is negative?

java/ddlogapi.c
```diff
@@ -229,29 +323,29 @@ JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1transaction_1commit_1dump_1
 
     jclass cls = (*env)->FindClass(env, "ddlogapi/DDlogAPI$FlatBufDescr");
     if (cls == NULL) {
-        return -1;
+        return;
```
mbudiu-vmw: why not throw here?
this certainly looks like an error.

java/ddlogapi.c
```diff
@@ -229,29 +323,29 @@ JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1transaction_1commit_1dump_1
...
     }
     jmethodID setMethod = (*env)->GetMethodID(env, cls, "set", "(Ljava/nio/ByteBuffer;JJ)V");
     if(setMethod == NULL) {
-        return -1;
+        return;
```
mbudiu-vmw: same here: if you used to return -1 shouldn't you throw?

java/ddlogapi.c
```diff
@@ -229,29 +323,29 @@ JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1transaction_1commit_1dump_1
...
     }
 
     jobject direct_buf = (*env)->NewDirectByteBuffer(env, buf_addr + buf_offset, (jlong)(buf_capacity - buf_offset));
 
     (*env)->CallVoidMethod(env, fbdescr, setMethod,
             direct_buf, (jlong)buf_size, (jlong)buf_offset);
 
-    return 0;
+    return;
```
mbudiu-vmw: no longer required.

java/ddlogapi.c
```diff
@@ -268,37 +362,41 @@ JNIEXPORT void JNICALL Java_ddlogapi_DDlogAPI_ddlog_1flatbuf_1free(
...
         updates[i] = (ddlog_cmd*)a[i];
-    int result = ddlog_apply_updates(
-        (ddlog_prog)progHandle, updates, size);
+    if (ddlog_apply_updates((ddlog_prog)progHandle, updates, size) < 0) {
+        throwDDlogException(env);
+    }
     (*env)->ReleaseLongArrayElements(env, commandHandles, a, 0);
     free(updates);
-    return (jint)result;
+    return;
```
mbudiu-vmw: redundant

java/ddlogapi.c
```diff
@@ -268,37 +362,41 @@ JNIEXPORT void JNICALL Java_ddlogapi_DDlogAPI_ddlog_1flatbuf_1free(
...
+JNIEXPORT void JNICALL Java_ddlogapi_DDlogAPI_ddlog_1apply_1updates(
     JNIEnv *env, jclass obj, jlong progHandle, jlongArray commandHandles) {
     jlong *a = (*env)->GetLongArrayElements(env, commandHandles, NULL);
     size_t size = (*env)->GetArrayLength(env, commandHandles);
     ddlog_cmd** updates = malloc(sizeof(ddlog_cmd*) * size);
     for (size_t i = 0; i < size; i++)
         updates[i] = (ddlog_cmd*)a[i];
-    int result = ddlog_apply_updates(
-        (ddlog_prog)progHandle, updates, size);
+    if (ddlog_apply_updates((ddlog_prog)progHandle, updates, size) < 0) {
```
mbudiu-vmw: check if updates is not null too. I don't know if you also want to check the free.

java/ddlogapi.c
```diff
@@ -268,37 +362,41 @@ JNIEXPORT void JNICALL Java_ddlogapi_DDlogAPI_ddlog_1flatbuf_1free(
...
 
-JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1apply_1updates_1from_1flatbuf(
+JNIEXPORT void JNICALL Java_ddlogapi_DDlogAPI_ddlog_1apply_1updates_1from_1flatbuf(
     JNIEnv *env, jclass obj, jlong progHandle, jbyteArray bytes, jint position) {
     jbyte *buf = (*env)->GetByteArrayElements(env, bytes, NULL);
     size_t size = (*env)->GetArrayLength(env, bytes);
 
-    int result = ddlog_apply_updates_from_flatbuf(
-        (ddlog_prog)progHandle, ((const unsigned char *) buf) + position, size);
+    if (ddlog_apply_updates_from_flatbuf(
```
mbudiu-vmw: If you want to be very strict perhaps you can also check buf for being null.

java/ddlogapi.c
```diff
@@ -268,37 +362,41 @@ JNIEXPORT void JNICALL Java_ddlogapi_DDlogAPI_ddlog_1flatbuf_1free(
...
+        (ddlog_prog)progHandle, ((const unsigned char *) buf) + position, size) < 0) {
+        throwDDlogException(env);
+    };
 
     (*env)->ReleaseByteArrayElements(env, bytes, buf, JNI_ABORT);
 
-    return (jint)result;
+    return;
```
mbudiu-vmw: remove

java/ddlogapi.c
```diff
@@ -323,15 +421,16 @@ bool dump_callback(uintptr_t callbackInfo, const ddlog_record* rec) {
     return (bool)result;
 }
 
-JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_dump_1table(
+JNIEXPORT void JNICALL Java_ddlogapi_DDlogAPI_dump_1table(
     JNIEnv *env, jobject obj, jlong progHandle, jint table, jstring callback) {
     struct CallbackInfo* cbinfo = createCallback(env, obj, callback, "(J)Z");
     if (cbinfo == NULL)
-        return -1;
+        return;
```
mbudiu-vmw: throw

java/ddlogapi.c
ryzhyk: It's C11: https://en.cppreference.com/w/c/language/storage_duration

I checked that it works with gcc and clang.

java/ddlogapi.c
ryzhyk: good point; also the second one should have been `java.util.Exception`.  Thanks for spotting.

java/ddlogapi.c
ryzhyk: Well, it uses thread-local state, so hopefully should be.  The dodginess here is that we rely on DDlog to print an error on failure, which is an invariant we should be careful to observe.

java/ddlogapi.c
ryzhyk: If `createCallback` failed, this means it has already thrown an exception.  We should just return and the exception will show up in Java.

java/ddlogapi.c
ryzhyk: I had that initially, but it's a bit tricky, cause we still need to check the error code and act on it, whether we raise the exception or not.  It was not very ergonomic.

java/ddlogapi.c
ryzhyk: Likewise, `FindClass` generates an exception on error, which will trigger as soon as we're back in the Java world.

PR #365, opened 2019-09-17
Topic/debug

d-e-s-o:
Use debug build instead of release

I performed some comparison of the run times of the Rust tests that take
up the majority of our pipeline's time, focusing on the different build
types available. In the past the argument has come up that while release
builds take longer, they execute faster and make up for that, which is
also the main reason we use release builds in those tests to begin with.

I performed two runs, one using `cargo test --release`, the other using
`cargo test` (which defaults to debug build).

debug: https://travis-ci.org/d-e-s-o/differential-datalog/builds/585822177
release: https://travis-ci.org/d-e-s-o/differential-datalog/builds/585822065

The results are as follows. Shown is the run time of the compile time
test in seconds for Mac (mac) and Linux (lin) and the change in run time
when switching from release to debug builds.

| Test            | Runtime Release | Runtime Debug | Change   |
| --------------- | --------------- | ------------- | -------- |
| tutorial (mac)  | 1418.48s        | 456.24s       | -68%     |
| tutorial (lin)  | 1324.71s        | 319.96s       | -76%     |
| redist (mac)    | 1780.83s        | 1590.56s      | -11%     |
| redist (lin)    | 1501.80s        | 1213.07s      | -19%     |
| span_uuid (mac) | 1390.25s        | 557.00s       | -60%     |
| span_uuid (lin) | 1228.94s        | 386.98s       | -69%     |
| simple (mac)    | 1980.93s        | timeout       | N/A      |
| simple (lin)    | 1673.17s        | timeout       | N/A      |
| path (mac)      | 1325.78s        | 378.96s       | -71%     |
| path (lin)      | 1240.93s        | 301.83s       | -76%     |
| ovn (mac)       | timeout         | 587.29s       | N/A      |
| ovn (lin)       | 2025.68s        | 503.90s       | -75%     |
| modules (mac)   | 1434.49s        | 463.38s       | -68%     |
| modules (lin)   | 1308.62s        | 338.31s       | -74%     |

As can be seen from the results, the difference between the two variants
is rather big. The only test that regressed is `simple`, which
consistently timed out for debug builds, confirming that some tests'
runtime may in fact regress more than the reduction in compile time
decreases (at least that is a possible interpretation, I haven't look at
the reasons in that detail).
While a mere two runs may not be considered enough, the difference in
runtime is so severe that I don't think there is much doubt about the
effect overall (and we can always revert). I also performed some more
runs locally (although only with the `path` test) and the results are
similar in nature.

Given those results, this change proposes to switch to using debug
builds for testing by default, which will save us in the order of 2h per
test run (this time is calculated not from the above numbers but from
looking at the actual end-to-end test runtime in Travis, which includes
setup and more; given the stark decrease in individual test run time we
may even think about running fast tests on the same VM to cut down on
setup time, but I did not go down that route).
In order to accommodate for the `simple` test, I make the stack testing
framework honor an environment variable, STACK_CARGO_FLAGS, that is set
to "--release" for this test (in Travis).

--- >8 ------ >8 ------ >8 (cut here)

- I am clubbing two more minor clippy cleanup changes in this pull request.

- I am not very familiar with Haskell and the list function introduced in
`Spec.hs` seems very much like a kludge. I am sure there is a more
elegant syntax to accomplish the same. If you know, please let me know.

### DIFF
--- a/test/Spec.hs
+++ b/test/Spec.hs
@@ -60,12 +60,9 @@ main = do
     tests <- goldenTests progress
     defaultMain tests
 
-bUILD_TYPE :: String
---bUILD_TYPE = "debug"
-bUILD_TYPE = "release"
-
-cargo_build_flag :: [String]
-cargo_build_flag = if bUILD_TYPE == "release" then ["--release"] else []
+cargo_build_flags :: IO [[Char]]
+cargo_build_flags = do
+  (maybe [] (return)) <$> lookupEnv "STACK_CARGO_FLAGS"
 
 goldenTests :: Bool -> IO TestTree
 goldenTests progress = do
@@ -206,7 +203,8 @@ compilerTest progress java file cli_args crate_types = do
                                  "\n\nstdout:\n" ++ jstdo
 
     -- compile it with Cargo
-    let cargo_proc = (proc "cargo" (["build", "--features=flatbuf"] ++ cargo_build_flag)) {
+    cargo_flags <- cargo_build_flags
+    let cargo_proc = (proc "cargo" (["build", "--features=flatbuf"] ++ cargo_flags)) {
                           cwd = Just $ dir </> rustProjectDir specname
                      }
     (ccode, cstdo, cstde) <- withProgress progress $ readCreateProcessWithExitCode cargo_proc ""
@@ -214,15 +212,7 @@ compilerTest progress java file cli_args crate_types = do
         errorWithoutStackTrace $ "cargo build failed with exit code " ++ show ccode ++
                                  "\nstdout:\n" ++ cstde ++
                                  "\n\nstdout:\n" ++ cstdo
-    {-let cargo_proc = (proc "cargo" (["test"] ++ cargo_build_flag)) {
-                          cwd = Just $ dir </> specname
-                     }
 
-    (code, stdo, stde) <- withProgress progress $ readCreateProcessWithExitCode cargo_proc ""
-    when (code /= ExitSuccess) $ do
-        errorWithoutStackTrace $ "cargo test failed with exit code " ++ show code ++
-                                 "\nstderr:\n" ++ stde ++
-                                 "\n\nstdout:\n" ++ stdo -}
     cliTest progress fname specname dir cli_args
 
 progressThread :: IO ()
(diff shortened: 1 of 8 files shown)


ryzhyk:
The argument for release builds was never speed (all out tests only take few seconds once compiled), but rather the fact that these tests better approximate production use.  I suggest we run at leas one or two tests with release builds to catch regressions.

test/Spec.hs
```diff
@@ -60,12 +60,12 @@ main = do
...
---bUILD_TYPE = "debug"
-bUILD_TYPE = "release"
+list :: String -> [String]
+list x = [x]
 
-cargo_build_flag :: [String]
-cargo_build_flag = if bUILD_TYPE == "release" then ["--release"] else []
+cargo_build_flags :: IO [[Char]]
+cargo_build_flags = do
+  (maybe [] (list)) <$> lookupEnv "STACK_CARGO_FLAGS"
```
ryzhyk: Try replacing `list` -> `return` (cause lists are monads in Haskell)

d-e-s-o: Very nice, thank you!

d-e-s-o:
Thanks for the explanation. I must have misunderstood the intent then. But given the impact of the build type, do you still see value in switching over? We do have one test that runs in release mode.

My argument would be that we should not penalize each pipeline run this way. In my opinion the assumption should be that behavior-wise debug builds behave like release builds (plus additional assertions; which is generally a good thing to test with) and treat divergence as a compiler bug (meaning here: assuming the chances are low). I'd rather we have, say, a nightly, periodic, or pre-release run in release mode to catch those low-likelihood cases.

There is also another case that we do not cover: changes in behavior between Rust compiler versions.

ryzhyk:
> Thanks for the explanation. I must have misunderstood the intent then. But given the impact of the build type, do you still see value in switching over? We do have one test that runs in release mode.

Yes, I agree it is worth switching over.

> My argument would be that we should not penalize each pipeline run this way. In my opinion the assumption should be that behavior-wise debug builds behave like release builds (plus additional assertions; which is generally a good thing to test with) and treat divergence as a compiler bug (meaning here: assuming the chances are low).

Sure, but because we seriously stress the Rust compiler, it is worth checking if we trigger a compiler bug (or, more likely, an out-of-memory situation)

> I'd rather we have, say, a nightly, periodic, or pre-release run in release mode to catch those low-likelihood cases.

+1

PR #369, opened 2019-09-19
Fix java tests

### LINKED ISSUE #337: Some older Java tests no longer work
after changing the Java generation API.

mbudiu-vmw:
Fixes #337

### DIFF
--- a/java/ddlogapi/DDlogCommand.java
+++ b/java/ddlogapi/DDlogCommand.java
@@ -1,7 +1,7 @@
 package ddlogapi;
 
 /**
- * Update to an input or output relation. 
+ * Update to an input or output relation.
  *
  * <p>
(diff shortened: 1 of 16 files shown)


java/ddlogapi/DDlogCommand.java
```diff
@@ -27,11 +27,11 @@
  * </p>
  */
 public interface DDlogCommand<R> {
-    public enum Kind {
+    enum Kind {
```
ryzhyk: Why shouldn't it be public?

mbudiu-vmw: Intellij told me that it's redundant; I think that if it is in an interface it is always public.

mbudiu-vmw: But it is indeed clearer if you mark it public...

PR #374, opened 2019-09-20
Test failing transaction

mbudiu-vmw:
Also fixes DDlogException.


mbudiu-vmw:
This test currently fails.

PR #376, opened 2019-09-20
Fix dependencies for "unit tests"

d-e-s-o:
The recently introduced unit test feature for DDLog Rust projects has a
flaw: there is an implicit dependency on the corresponding *_ddlog
directory having been created first (which is done by the compiler
tests) and if that is not the case the test will just fail.
We need to make this dependency explicit such that running the unit
tests also causes generation of the *_ddlog project.
This change does exactly that. We factor out the Rust project generation
into a separate "test" and then make use of tasty's "after" feature in
order to make this test a dependency of the unit tests and the compiler
tests. The "after" feature got introduced with tasty 1.2, which
therefore is added as an extra dependency.
With this change, running of only the unit tests for server_api, for
example, could be accomplished using this mouthful:
  > $ stack test --ta
  >   '--pattern "$(NF) == \"generate server_api\" ||
  >              ($(NF-1) == \"unit tests\" && $(NF) == \"server_api\")"'

(we need to explicitly match the "generate server_api" test because of
 the way tasty implements dependency handling, see its [README][])

[README]: https://github.com/feuerbach/tasty/blob/master/README.md#dependencies

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -98,7 +98,7 @@ tests:
     - -Werror
     dependencies:
     - differential-datalog
-    - tasty
+    - tasty >= 1.2
     - tasty-hunit
     - tasty-golden
     - filepath
--- a/stack.yaml
+++ b/stack.yaml
@@ -42,9 +42,11 @@ packages:
 # extra-deps: []
 
 extra-deps:
-# We need githash 0.1.3.3 in order to have git-worktree(1) support. It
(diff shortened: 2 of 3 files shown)


d-e-s-o:
Hm, hold on with the review, I think the change breaks the souffle tests.

d-e-s-o:
Okay, everything should be fine now.

PR #388, opened 2019-09-25
Topic/debug and more

d-e-s-o:
This pull request is an accumulation of a couple things. In summary we:
- implement `Debug` for more type to improve ergonomics
- remove the `on_next` method from the `Observer` trait
- add another unit test, testing for actual deltas received

---

Pipeline run is to be found [here](https://travis-ci.org/d-e-s-o/differential-datalog/builds/589557611).


rust/template/differential_datalog/program.rs
```diff
@@ -839,6 +842,28 @@ pub struct RunningProgram<V: Val> {
     pub profile: Arc<Mutex<Profile>>,
 }
 
+// Right now this Debug implementation is more or less a short cut.
+// Ideally we would want to implement Debug for `RelationInstance`, but
```
mbudiu-vmw: I don't think this is a good comment to describe what this function does. Perhaps the comment should be somewhere else.

rust/template/observe/observer.rs
```diff
@@ -1,8 +1,9 @@
+use std::fmt::Debug;
 use std::sync::Arc;
 use std::sync::Mutex;
 
 /// A trait for objects that can observe an observable one.
-pub trait Observer<T, E>: Send
+pub trait Observer<T, E>: Debug + Send
```
mbudiu-vmw: We should rename this class to perhaps TransactionObserver.
The Observer interface must have an on_next method.

rust/template/differential_datalog/program.rs
d-e-s-o: Thanks for your review. It's not a comment describing what the function does as it's not a doc comment (`///`). I'd say the function is as self-explanatory as it gets. The comment is pointing out why we have a custom implementation of `Debug` and not an auto-derive, which is not obvious from the code from what I can tell.

rust/template/observe/observer.rs
d-e-s-o: Can you explain why we need an `on_next` method?

mbudiu-vmw: I didn't say we need an on_next method, I said that the Observer pattern must have an on_next method. See e.g., https://www.journaldev.com/22594/rxjava-observables-observers.
So this is not a standard Observer, so we should call it something else.

d-e-s-o: I see. Thanks for the link. That makes it a bit more clear. I tend to disagree, though. Just because some Java library implements an Observer which has an `on_next` method does neither mean that the name `Observer` is reserved nor that everybody has an association/expectation that just because it's called `Observer` it implements exactly this interface. My understanding is that observer is a [generic design pattern](https://en.wikipedia.org/wiki/Observer_pattern) with a wide variety of implementations and I don't see us claiming any relationship to or compatibility with `RxJava`.
So honestly, I'd like to keep the name this concise until we have multiple observers that require differentiation. Or am I missing something?

ryzhyk: It is true that `TransactionObserver` may be a more accurate name for this interface, not because of `on_next()` method not being part of it, but because the interface groups updates in batches punctuated by `on_start`, `on_commit`, or whatever those methods are called. But `TransactionObserver` is a bit verbose and so I prefer sticking to `Observer`.

PR #389, opened 2019-09-25
Allow for automatic consumption of dl.vim

d-e-s-o:
The Vim syntax highlighting file `dl.vim` does not lend itself to being
consumed by the various Vim Plugin managers out there, as most of them
expect a Vim configuration root style layout.
To allow for easy integration with this software, this change moves the
file into a newly created subfolder `tools/vim/syntax/`. It also drops a
note about integration with one Plugin manager, [Vundle](https://github.com/VundleVim/Vundle.vim), in the README
(most managers are very similar in configuration style and Vundle is one
of the first).


d-e-s-o:
I've cancelled the pipeline as I don't think it will check anything relevant for this pull request. Feel free to restart it if you see a need.

PR #392, opened 2019-09-26
Initial GitLab CI integration

ryzhyk:
GitLab CI script that replicates all existing Travis tests (on Linux only) + runs all Java tests and most tests from the Souffle repo. It completes in <30minutes using GitLab shared workers only, which is 4x faster than Travis, while running more tests.

### DIFF
--- a/java/test3/run.sh
+++ b/java/test3/run.sh
@@ -26,12 +26,13 @@ pushd ../../test/datalog_tests/${PROG}_ddlog
 cargo build --release --features=flatbuf
 popd
 
+# Build the Java library with the DDlog API
+make -C ..
+
 # Compile generated Java classes (the FlatBuffer Java package must be compiled first and must be in the
 # $CLASSPATH)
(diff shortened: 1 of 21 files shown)


java/test3/run.sh
```diff
@@ -26,12 +26,13 @@ pushd ../../test/datalog_tests/${PROG}_ddlog
 cargo build --release --features=flatbuf
 popd
 
+# Build the Java library with the DDlog API
+make -C ..
```
d-e-s-o: Add `--jobs` option to allow for concurrency?

ryzhyk:
Thanks, @d-e-s-o! Let me commit this as is. I'll add `--jobs` as part of a future PR (that make takes 1s anyway, so it's not urgent). I will also keep the `.gitlab` file at the top repo.  The main reason is that I expect we will setup such mirrors more than once in the future, e.g., you probably want one for your fork of DDlog. I bet we will forget to set the repo option every time, so fewer options is better :) Plus Travis is not going away any time soon; I just plan to reduce the number of Travis tests significantly.

ryzhyk:
@mbudiu-vmw, I will merge this so that we can start enjoying faster test runs, but if you have any comments, I will incorporate them in future PRs.

PR #397, opened 2019-09-26
Remove most tests from Travis

ryzhyk:
GitLab is proving to be a way more scalable CI platform than Travis.
So, in the interests of speeding up test runs we remove most tests,
which are now redundant, from Travis.

We still want to use Travis for:
- MacOS tests
- Release builds
- Tests using newer versions of rustc/clippy/rustfmt than GitLab
  (although there is no reason we couldn't test those in GitLab)

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -19,6 +19,8 @@ test-tutorial:
     script: stack --no-terminal test --ta "-p tutorial"
 
 test-simple:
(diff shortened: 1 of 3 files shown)


ryzhyk:
> Makes sense to me. Just out of curiosity, what happened when you did a release build of `simple` on a Gitlab worker? Did it hit some timeout?

It did not timeout, but it ran for ~50minutes.

PR #401, opened 2019-09-27
Fix handling of gzipped reference files

ryzhyk:


ryzhyk:
> Can't tell if there is a new reference output; git shows it as deleted.

Just double checked. Seems to be there.

PR #404, opened 2019-10-01
Use VMware-hosted runners to run more tests in GitLab CI .

ryzhyk:
- Support for project-specif CI runners hosted by VMware (`ddlog-ci-1`, `ddlog-ci-2`)
- Enabled all imported souffle tests (on `ddlog-ci-1`)
- Run the OVN test suite from branch `ddlog-dev-v2` of the `ovn-org` repo
- Unrelated: an ugly use-after-free error in `ddlogapi.c`
  - @mbudiu-vmw, do you think we can have a test case for this?

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -19,9 +19,9 @@ test-tutorial:
     script: stack --no-terminal test --ta "-p tutorial"
 
 test-simple:
-    # TODO: add release builds when we have more powerful workers
-    #script: STACK_CARGO_FLAGS='--release' stack --no-terminal test --ta "-p simple"
-    script: stack --no-terminal test --ta "-p simple"
+    tags:
+        - ddlog-ci-1
+    script: STACK_CARGO_FLAGS='--release' stack --no-terminal test --ta "-p simple"
 
 test-modules:
     script: stack --no-terminal test --ta "-p modules"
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -5,6 +5,8 @@ default:
 .install-ddlog:
     before_script:
         - stack install
+        - mkdir -p /root/.local/ddlog
+        - cp -r lib /root/.local/ddlog
```
mbudiu-vmw: why not just a symlink?

.gitlab-ci.yml
```diff
@@ -79,40 +81,66 @@ test-flatbuf:
...
+          ./configure  &&
+          make -j6)
+        # TODO: maintain and eventually remove the list of tests.
+        # TODO: use -j6 once build script is fixed
+        - (cd /root &&
+          git clone https://github.com/ovn-org/ovn.git &&
+          cd ovn && git checkout ddlog-dev-v2 &&
+          ./boot.sh &&
+          ./configure --with-ddlog=/root/.local/ddlog/lib --with-ovs-source=/root/ovs &&
+          make check -j1 TESTSUITEFLAGS="117-135 138-141 143-145 147-149 151-152 154-161 167-169 172 174-181 183-185 187-189 191 193-196 198-199 201-202")
```
mbudiu-vmw: this looks very brittle, but I don't know if there is any alternative

gitlab-ci/README.md
```diff
@@ -25,6 +25,58 @@
...
+concurrent = 8
+check_interval = 0
+
+[session_server]
+  session_timeout = 1800
+
+[[runners]]
+  name = "ddlog-ci-1"
+  url = "https://gitlab.com"
+  token = "Bjx-aJUGEH6Fp6rBEXkf"
```
mbudiu-vmw: is it ok to have this here in clear?

.gitlab-ci.yml
ryzhyk: It is. I plan to maintain this list until it becomes empty.

gitlab-ci/README.md
ryzhyk: As far as I can tell, anyone with this token can register their worker to run our tests. They're welcome to do so :) The worst a malicious worker should be able to do is report success when the test should actually fail.  But at least we will see in GitLab that external workers are being used and blacklist them if necessary.

.gitlab-ci.yml
ryzhyk: good point

PR #405, opened 2019-10-01
Debugging tool: dump intermediate relations

### LINKED ISSUE #391: Add a flag to label all non-input relations as `output`.
This is helpful when debugging a large program when the user wants to dump the contents of some relations without re-compiling the whole thing.

mbudiu-vmw:
Fixes #391
Also fixes the output of the `server_api.dl` test.
I have used the `--debug` flag, but perhaps we want a different one.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -47,7 +47,7 @@ data TOption = Help
              | Action String
              | LibDir String
              | Java
-             | Debug
+             | OutputInternal
              | DynLib
              | NoDynLib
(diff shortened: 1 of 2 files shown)


ryzhyk:
> I have used the `--debug` flag, but perhaps we want a different one.

Looks good, but yeah, `--debug` is a really unfortunate name. Pretty much anything else would be better :)

How about `--output-internal-relations`?

PR #407, opened 2019-10-01
Script to run tests; test for --output-internal-relations

mbudiu-vmw:
The script `test/datalog_tests/run-tests.sh` will run all the datalog tests that are not supposed to fail.
It cannot do the same things as Spec.hs, since it cannot dump the abstract syntax tree, but it can run the cli on the dat file it if exists.

Added a test for the new compiler flag.


test/datalog_tests/run-tests.sh
```diff
@@ -0,0 +1,51 @@
...
+
+    ddlog -i ${base}.dl -L../../lib ${extraflags[${base}]}
+    cd ${base}_ddlog
+    cargo build
+    cd ..
+    if [ -f ${base}.dat ]; then
+        ${base}_ddlog/target/debug/${base}_cli --no-print <${base}.dat >${base}.dump
+        if [ -f ${base}.dump.expected ]; then
+            diff -q ${base}.dump ${base}.dump.expected
+        elif [ -f ${base}.dump.expected.gz ]; then
```
ryzhyk: I we should check for `.gz` first and ignore the non-`.gz` file if it exists.

test/datalog_tests/run-tests.sh
```diff
@@ -0,0 +1,51 @@
...
+            gzip ${base}.dump
```
ryzhyk: I don't think this is necessary. `zdiff` can compare a compressed  file against an uncompressed one.

test/datalog_tests/run-tests.sh
```diff
@@ -0,0 +1,51 @@
...
+            zdiff -q ${base}.dump.gz ${base}.dump.expected.gz
+        fi
+    fi
+    rm -rf ${base}_ddlog ${base}.dump ${base}.dump.gz
```
ryzhyk: I prefer not to cleanup automatically. Oftentimes, I actually need to invoke the compiled executable.

test/datalog_tests/run-tests.sh
```diff
@@ -0,0 +1,51 @@
...
+# Always run in the directory where the script is installed
+cd ${mydir}
+
+function run_test {
+    local testname=$1;
+    local base=$(basename ${testname} .dl)
+
+    ddlog -i ${base}.dl -L../../lib ${extraflags[${base}]}
+    cd ${base}_ddlog
+    cargo build
```
ryzhyk: `cargo` needs extra flags too, e.g., `--release`,  `--lib`, and  `--features=flatbuf`.

test/datalog_tests/run-tests.sh
```diff
@@ -0,0 +1,51 @@
...
+    cd ..
+    if [ -f ${base}.dat ]; then
+        ${base}_ddlog/target/debug/${base}_cli --no-print <${base}.dat >${base}.dump
```
ryzhyk: `debug` must be replaced with `release` in case of release build.  Which suggests that the script should take a `build-target` argument.

test/datalog_tests/run-tests.sh
ryzhyk: Additionally, we should be able to set the `RUSTFLAGS` env variable, e.g., `RUSTFLAGS="-C opt-level=z" cargo...`.

mbudiu-vmw:
I have refactored the script into two scripts, but it's uglier now.
The reason is that when you need to pass values between the scripts they are not necessarily strings without spaces, so you cannot easily pass them on the command-line. So I am only using environment variables to pass information between the two scripts.

PR #408, opened 2019-10-03
Implemented flag --output-input-relations

mbudiu-vmw:
The only thing I am not sure about is whether I have to generate a fresh name for the variable; right now it's hardwired at "x".
This simplifies the flatbuf test significantly.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -48,6 +48,7 @@ data TOption = Help
              | LibDir String
              | Java
              | OutputInternal
+             | OutputInput String
              | DynLib
              | NoDynLib
              | StaticLib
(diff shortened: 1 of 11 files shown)


app/Main.hs
```diff
@@ -149,6 +154,28 @@ main = do
                               return ()
          ActionCompile -> compileProg config
 
+-- create an output relation for each input relation
+mirrorInputRelations :: DatalogProgram -> String -> DatalogProgram
```
ryzhyk: This function should probably be in `DatalogProgram.hs` and should be called `progMirrorInputRelations`
Likewise, `makeOutputRelations` should be `DatalogProgram.progOutputInternalRelations` or something similar.

PR #410, opened 2019-10-04
Topic/tcp channel

d-e-s-o:
This pull request introduces the concept of a TCP channel. It's basically Remy's change with follow up changes that fix certain problems or improve the API. I did not want to club everything into a single change as the individual commits serve as reference why things were done the way they were done.

----

Unfortunately even with this pull request we are not "there" yet. We need to rework the JSON based transport to make notification semantics sane and then run our existing tests for TCP channel based setups as well.

### DIFF
--- a/rust/template/observe/observable.rs
+++ b/rust/template/observe/observable.rs
@@ -1,7 +1,10 @@
 use std::fmt::Debug;
+use std::sync::Arc;
+use std::sync::Mutex;
 
-use super::Observer;
-use super::Subscription;
+use crate::observer::ObserverBox;
+use crate::subscription::Subscription;
+use crate::subscription::UpdatesSubscription;
 
 /// A trait for objects that can be observed.
 pub trait Observable<T, E>: Debug
@@ -12,5 +15,29 @@ where
     /// An Observer subscribes to an Observable to listen to data
     /// emitted by the latter. The Observer stops listening when
     /// the Subscription returned is canceled.
-    fn subscribe(&mut self, observer: Box<dyn Observer<T, E> + Send>) -> Box<dyn Subscription>;
+    fn subscribe(&mut self, observer: ObserverBox<T, E>) -> Box<dyn Subscription>;
+}
+
+/// A very simple observable that supports subscription of a single
+/// observer.
+#[derive(Debug)]
+pub struct UpdatesObservable<T, E> {
+    /// A reference to the `Observer` we subscribed to.
+    pub observer: Arc<Mutex<Option<ObserverBox<T, E>>>>,
+}
+
+impl<T, E> Observable<T, E> for UpdatesObservable<T, E>
+where
+    T: Debug + Send + 'static,
+    E: Debug + Send + 'static,
+{
+    /// An observer subscribes to the delta stream from an outlet.
+    fn subscribe(&mut self, observer: ObserverBox<T, E>) -> Box<dyn Subscription> {
+        // TODO: Should probably panic or report an error on attempt to
+        //       subscribe multiple observers.
+        *self.observer.lock().unwrap() = Some(observer);
+        Box::new(UpdatesSubscription {
+            observer: self.observer.clone(),
+        })
+    }
 }
--- a/rust/template/tcp_channel/lib.rs
+++ b/rust/template/tcp_channel/lib.rs
@@ -41,6 +41,11 @@ impl<T: DeserializeOwned + Send + Debug + 'static> TcpReceiver<T> {
     /// inactive on creation, so to receive data the user must call
     /// `connect` to start a connection; and `listen` to listen for
     /// incoming data.
+    ///
+    /// `addr` may have a port set (by setting it to 0). In such a case
+    /// the system will assign a port that is free. To retrieve this
+    /// assigned port (in the form of the full `SocketAddr`), use the
+    /// `addr` method.
     pub fn new(addr: SocketAddr) -> Self {
         Self {
             addr,
@@ -54,6 +59,14 @@ impl<T: DeserializeOwned + Send + Debug + 'static> TcpReceiver<T> {
     pub fn connect(&mut self) -> JoinHandle<()> {
         let listener = TcpListener::bind(self.addr).unwrap();
         let stream = self.stream.clone();
+
+        // We want to allow for auto-assigned ports, by letting the user
+        // specify a `SocketAddr` with port 0. In this case, after
+        // actually binding to an address, we need to update the port we
+        // got assigned in `addr`, but for simplicity we just copy the
+        // entire thing.
+        self.addr = listener.local_addr().unwrap();
+
         spawn(move || {
             let (s, _) = listener.accept().unwrap();
             let mut stream = stream.lock().unwrap();
@@ -98,6 +111,11 @@ impl<T: DeserializeOwned + Send + Debug + 'static> TcpReceiver<T> {
     pub fn disconnect(&mut self) {
         // Drop TCP stream to close connection
         let _ = self.stream.lock().unwrap().take();
+    }
+
+    /// Retrieve the address we are listening on.
+    pub fn addr(&self) -> &SocketAddr {
+        &self.addr
     }
 }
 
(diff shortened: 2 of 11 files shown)


src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -166,6 +166,8 @@ rustLibFiles specname =
         , (dir </> "ovsdb/Cargo.toml"                   , $(embedFile "rust/template/ovsdb/Cargo.toml"))
         , (dir </> "ovsdb/lib.rs"                       , $(embedFile "rust/template/ovsdb/lib.rs"))
         , (dir </> "ovsdb/test.rs"                      , $(embedFile "rust/template/ovsdb/test.rs"))
+        , (dir </> "tcp_channel/Cargo.toml"             , $(embedFile "rust/template/tcp_channel/Cargo.toml"))
```
mbudiu-vmw: do we need this in every project?
maybe we should add a --feature-like flag. Then we can probably remove ovsdb too.

d-e-s-o: We should not need it everywhere. The question is whether it's worth adding some form of gate for it. Let me open an issue to discuss that.

rust/template/tcp_channel/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+    fn accept(
+        stream: Arc<Mutex<Option<RawFd>>>,
+        listener: TcpListener,
+        observer: Arc<Mutex<Option<Box<dyn Observer<T, String> + Send>>>>,
+    ) -> JoinHandle<io::Result<()>> {
+        spawn(move || {
+            let (socket, _) = listener.accept()?;
+            *stream.lock().unwrap() = Some(socket.as_raw_fd());
+
+            let reader = BufReader::new(socket);
```
ryzhyk: Why are we using text-based encoding as opposed to binary encoding, which should be more compact?

Ok, looking at the description of the PR, this is wip and will eventually change. That's fine then.

rust/template/tcp_channel/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+            let mut lines = reader.lines();
+            loop {
+                let mut upds = lines
+                    .by_ref()
+                    .take_while(|result| match result {
+                        Ok(line) if line == "commit" => false,
```
ryzhyk: Should there be a "start" command as well or does "commit" implicitly start the next set of updates?

rust/template/tcp_channel/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+        listener: TcpListener,
+        observer: Arc<Mutex<Option<Box<dyn Observer<T, String> + Send>>>>,
+    ) -> JoinHandle<io::Result<()>> {
+        spawn(move || {
+            let (socket, _) = listener.accept()?;
+            *stream.lock().unwrap() = Some(socket.as_raw_fd());
+
+            let reader = BufReader::new(socket);
+            let mut lines = reader.lines();
+            loop {
```
ryzhyk: I don't have a better solution and this should not prevent this PR from being merged, but I wonder if there is a design that does not require `TCPReceiver` to buffer.

rust/template/tcp_channel/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+{
+    /// An observer subscribes to the receiving end of a TCP channel to
+    /// listen to incoming data.
+    fn subscribe(
+        &mut self,
+        observer: Box<dyn Observer<T, String> + Send>,
+    ) -> Box<dyn Subscription> {
+        let mut guard = self.observer.lock().unwrap();
+        match *guard {
+            Some(_) => panic!(
```
ryzhyk: This really should not panic. Option 1: quietly replace the existing observer; Option 2: return an error; Option 3: replace the existing observer and return the old one. Options 1 and 3 do not really work in the presence of the `Subscription` atrocity. I think Option 2 is probably the simplest one, but requires changing the API to allow errors. Whatever semantics we choose must be part of the contract for all observables.

Thoughts?

rust/template/tcp_channel/lib.rs
```diff
@@ -0,0 +1,262 @@
...
+        A: ToSocketAddrs,
+    {
+        Ok(Self {
+            stream: TcpStream::connect(addr)?,
+        })
+    }
+}
+
+impl<T: Serialize + Send> Observer<T, String> for TcpSender {
+    /// Connect to the specified TCP address.
```
ryzhyk: The method does not seem to do what the comment says it does.

rust/template/tcp_channel/lib.rs
d-e-s-o: What exactly is the problem with buffering? How would it be possible to ever go without buffering given that you don't know how much data you have to read in advance.

rust/template/tcp_channel/lib.rs
d-e-s-o: I'll make it return an error. This seems like the sanest approach to me, from a usability point of view.

rust/template/tcp_channel/lib.rs
d-e-s-o: I guess I can see if I can revert these changes, given that I'll switch over to using a binary encoding scheme anyway.

rust/template/observe/observable.rs
```diff
@@ -14,3 +18,27 @@ where
...
+    T: Debug + Send + 'static,
+    E: Debug + Send + 'static,
+{
+    /// An observer subscribes to the delta stream from an outlet.
+    fn subscribe(&mut self, observer: Box<dyn Observer<T, E> + Send>) -> Box<dyn Subscription> {
+        // TODO: Should probably panic or report an error on attempt to
+        //       subscribe multiple observers.
+        *self.observer.lock().unwrap() = Some(observer);
+        Box::new(UpdatesSubscription {
+            observer: self.observer.clone(),
```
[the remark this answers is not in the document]
d-e-s-o: Yeah. I kept it for now because my attempt at removing it wasn't too successful. But I know what went wrong...so...it will go eventually.

rust/template/observe/observable.rs
```diff
@@ -14,3 +18,27 @@ where
...
+    pub observer: Arc<Mutex<Option<ObserverBox<T, E>>>>,
+}
+
+impl<T, E> Observable<T, E> for UpdatesObservable<T, E>
+where
+    T: Debug + Send + 'static,
+    E: Debug + Send + 'static,
+{
+    /// An observer subscribes to the delta stream from an outlet.
+    fn subscribe(&mut self, observer: Box<dyn Observer<T, E> + Send>) -> Box<dyn Subscription> {
```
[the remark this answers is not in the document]
d-e-s-o: Yes, probably. Thanks!

rust/template/tcp_channel/lib.rs
ryzhyk: `TCPReceiver` could theoretically feed data directly to the observer as soon as it arrives. Problem is, `TCPReceiver` can have multiple connections, all sending data at the same time. Buffering is necessary to isolate these concurrent transactions.

One option is to spawn an observable per connection and pump data to this observable, which would allow policy decisions about buffering to be pushed downstream, e.g., to whoever aggregates data from all these observables. However, this will introduce complexity, which I'd rather avoid in this early prototype.

rust/template/tcp_channel/lib.rs
d-e-s-o: "commit" acts as the marker that we have read enough and an indication that we can now actually start processing. There is no explicit "start" send because "commit" basically subsumes that.
This will change once we switch to using a binary encoding, in which case we will have a 1:1 relationship between events (`on_***` method calls) and messages being sent over the wire, so I am not going to change anything here for this patch.

d-e-s-o:
I have addressed some of the comments, others will come in part two. Thanks for the reviews!

rust/template/tcp_channel/lib.rs
d-e-s-o: As we discussed in person, there is a slight misunderstanding in that a `TcpReceiver` only ever accepts a single connection (in an attempt to form a "channel"). We will always need some form of buffering in order to deserialize a message of unknown length, but we don't have to accumulate updates across `on_updates` calls. This will change shortly along with the switch of the encoder to use. Resolving.

Issue #381, opened 2019-09-24
Make sure the C API thread-safe


ryzhyk:
Potential races:
- The use of `Arc::get_mut()` in `ddlog_record_commands`
- Deallocation of `Arc<HDDlog>` in `stop`

@d-e-s-o, do you see any other issues?

d-e-s-o:
Thanks for filing this issue. I think these are the known issues at this point. Other than that (not really related to thread safety just sanity in general), we need to take care of not panicing in Rust code called from C (or at least not without stopping the unwind before crossing FFI boundaries). One instance (we discussed over Slack) will be fixed shortly, another is tracked by https://github.com/vmware/differential-datalog/issues/342 which I shall look into eventually. But those may not be all locations where we panic, not sure.

d-e-s-o:
Renaming this issue to reflect it also being about unwinding issues (and any other source of unsafety/undefined behavior, I suppose) after Slack discussion.

Perhaps it's best to keep a list of unwind issues as we stumble over them:
- [ ] [`expect`](https://github.com/vmware/differential-datalog/blob/934f876312c3466fd806cfee078843350a71da73/rust/template/src/api.rs#L985) in `ddlog_profile`
- [ ] [`unwrap`](https://github.com/vmware/differential-datalog/blob/934f876312c3466fd806cfee078843350a71da73/rust/template/differential_datalog/record.rs#L163) in `ddlog_dump_record`

PR #412, opened 2019-10-05
Implement get_table_name

### LINKED ISSUE #394: Add `ddlog_get_table_name` to the C API.
This is the inverse of `ddlog_get_table_id()`.

mbudiu-vmw:
Fixes #394

### DIFF
--- a/java/ddlogapi.c
+++ b/java/ddlogapi.c
@@ -436,8 +436,11 @@ JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1get_1table_1id(
 JNIEXPORT jstring JNICALL Java_ddlogapi_DDlogAPI_ddlog_1get_1table_1name(
     JNIEnv *env, jclass class, jint id) {
     const char* table = ddlog_get_table_name(id);
+    if (table == NULL) {
+        throwDDlogException(env, "Unknown table id");
+        return NULL;
+    };
     jstring result = (*env)->NewStringUTF(env, table);
-    ddlog_string_free(table);
     return result;
 }
 
--- a/java/ddlogapi.c
+++ b/java/ddlogapi.c
@@ -433,6 +433,14 @@ JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1get_1table_1id(
     return (jint)id;
 }
 
+JNIEXPORT jstring JNICALL Java_ddlogapi_DDlogAPI_ddlog_1get_1table_1name(
+    JNIEnv *env, jclass class, jint id) {
+    const char* table = ddlog_get_table_name(id);
+    jstring result = (*env)->NewStringUTF(env, table);
+    ddlog_string_free(table);
+    return result;
+}
+
 bool dump_callback(uintptr_t callbackInfo, const ddlog_record* rec) {
     struct CallbackInfo* cbi = (struct CallbackInfo*)callbackInfo;
     JNIEnv* env = cbi->env;
--- a/rust/template/ddlog.h
+++ b/rust/template/ddlog.h
@@ -61,7 +61,6 @@ extern table_id ddlog_get_table_id(const char* tname);
 /*
  * Get DDlog table name from id.
  *
- * The returned string must be deallocated using ddlog_string_free().
  * On error, returns NULL.
  */
 extern const char* ddlog_get_table_name(table_id id);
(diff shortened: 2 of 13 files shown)


rust/template/src/api.rs
```diff
@@ -563,6 +567,16 @@ pub unsafe extern "C" fn ddlog_get_table_id(tname: *const raw::c_char) -> libc::
     }
 }
 
+#[no_mangle]
+pub unsafe extern "C" fn ddlog_get_table_name(tid: libc::size_t) -> *const raw::c_char {
+    match HDDlog::get_table_name(tid) {
+        Ok(name) => ffi::CString::new(name)
+            .map(ffi::CString::into_raw)
+            .unwrap(),
```
d-e-s-o: We are trying to get rid of `unwrap`s in C APIs as unwinding across ffi boundaries is undefined behavior (https://github.com/vmware/differential-datalog/issues/381). Can we return `NULL` here gracefully instead?

mbudiu-vmw: I didn't read the chapter on unwrap yet, but this parallels the code for `profile` (minus the unwrap_or_else). Is the missing _else the problem?

d-e-s-o: Yes, exactly. `unwrap` may panic whereas `unwrap_or_else` would evaluate to the "or else" part.

java/ddlogapi.c
```diff
@@ -433,6 +433,14 @@ JNIEXPORT jint JNICALL Java_ddlogapi_DDlogAPI_ddlog_1get_1table_1id(
     return (jint)id;
 }
 
+JNIEXPORT jstring JNICALL Java_ddlogapi_DDlogAPI_ddlog_1get_1table_1name(
+    JNIEnv *env, jclass class, jint id) {
+    const char* table = ddlog_get_table_name(id);
```
ryzhyk: What happens if `table==NULL`? Will it return a Java `null`? Shouldn't it throw an exception in this case?

rust/template/ddlog.h
```diff
@@ -488,7 +496,7 @@ extern void ddlog_free(ddlog_record *rec);
  * (currently only applicable to the string returned by `ddlog_profile()` and
  * `ddlog_dump_record()`).
  */
-extern void ddlog_string_free(char *s);
+extern void ddlog_string_free(const char *s);
```
ryzhyk: It shouldn't be `const`, as it actually destroys the string.

mbudiu-vmw: I think it should be const, even though it destroys the string.
Otherwise I cannot deallocate any const pointer ever.
For example, the pointer that comes from `get_table_name`.

java/ddlogapi.c
mbudiu-vmw: There are two cases it could return null: one is that unwrap failed, which I am not sure what it means, and when the id is wrong (and ddlogt_get_table_name returns null), which would warrant an exception. I don't know how you can tell them apart, though. Ideally the unwrap_or_else would set up the error message, but I don't have a handle to the HDDlog in that function.

rust/template/ddlog.h
ryzhyk: A pointer that is supposed to be deallocated by the user must not be `const`. In particular, `get_table_name` should not return a `const` pointer. When we fix the API so that the user does not have to free the string, we will add the `const` qualifier to `get_table_name`, which will force users to stop calling `ddlog_string_free`.

java/ddlogapi.c
ryzhyk: I think unwrap will fail if the string cannot be converted to a C string, which should only be possible it its utf8 encoding contains `0` characters. Not sure if it's possible in practice, but in any case I think it's better to return a generic error message like "ddlog_get_table_name failed" than to crash the program.

mbudiu-vmw: how do I return this error message from rust without a handle to HDDlog?

ryzhyk: You can return it directly from `ddlogapi.c` by passing it as the second argument to `throwDDlogException`.

mbudiu-vmw: I can do that, but it would be nicer to distinguish between these two cases. The message should really be set in Rust, not in C, where it's too late.

rust/template/ddlog.h
mbudiu-vmw: Actually we do want to pass a const pointer to the user, because we don't want the user to modify that string. So we should cast it to const. This is kind of stupid really, const should not deal with the pointer lifetime, just with the contents of the memory pointed by. There should be a different mechanism for the lifetime, but we will not fix C today.

java/ddlogapi.c
ryzhyk: That's true. Ok, let me see if I can fix it. I'll push a commit to your branch...

mbudiu-vmw: This is not a high value thing. If you spend more than 5 minutes on this let's just forget it; we can file an issue if you really want to do this at some point.

PR #417, opened 2019-10-08
Topic/tcp channel tests

d-e-s-o:
This change concludes the TCP channel infrastructure. Most importantly we run the existing tests for the observable-observer infrastructure to also test TCP channel based observers. Also, run the test in the GitLab CI/CD pipeline.

### DIFF
--- a/rust/template/tcp_channel/receiver.rs
+++ b/rust/template/tcp_channel/receiver.rs
@@ -1,5 +1,6 @@
 use std::fmt::Debug;
 use std::io::BufReader;
+use std::io::Error;
 use std::net::SocketAddr;
 use std::net::TcpListener;
 use std::net::ToSocketAddrs;
@@ -14,9 +15,10 @@ use std::thread::JoinHandle;
 use bincode::deserialize_from;
 
 use libc::close;
+use libc::shutdown;
+use libc::SHUT_RDWR;
 
 use observe::Observable;
-use observe::Observer;
 use observe::ObserverBox;
 use observe::Subscription;
 use observe::UpdatesSubscription;
@@ -25,15 +27,54 @@ use serde::de::DeserializeOwned;
 
 use crate::message::Message;
 
+#[derive(Copy, Clone, Debug)]
+enum Fd {
+    /// We are still listening for an incoming connection and
+    /// this is the corresponding file descriptor.
+    Listening(RawFd),
+    /// We have accepted a connection and read data from it.
+    Accepted(RawFd),
+    /// The listener/accepted connection has been closed.
+    Closed,
+}
+
+impl Fd {
+    fn close(&mut self) -> Result<(), Error> {
+        match *self {
+            Fd::Listening(fd) | Fd::Accepted(fd) => {
+                let rc = unsafe { shutdown(fd, SHUT_RDWR) };
+                if rc != 0 {
+                    return Err(Error::last_os_error());
+                }
+
+                // Bad luck if we fail the close. There is not much we
+                // can do about that.
+                *self = Fd::Closed;
+
+                let rc = unsafe { close(fd) };
+                if rc != 0 {
+                    return Err(Error::last_os_error());
+                }
+                Ok(())
+            }
+            Fd::Closed => Ok(()),
+        }
+    }
+}
+
 /// The receiving end of a TCP channel has an address
 /// and streams data to an observer.
 #[derive(Debug)]
 pub struct TcpReceiver<T> {
+    /// The address we are listening on.
     addr: SocketAddr,
-    listener: RawFd,
-    stream: Arc<Mutex<Option<RawFd>>>,
-    thread: JoinHandle<Result<(), String>>,
-    observer: Arc<Mutex<Option<Box<dyn Observer<T, String> + Send>>>>,
+    /// Our listener/connection file descriptor state; shared with the
+    /// thread accepting connections and reading streamed data.
+    fd: Arc<Mutex<Fd>>,
+    /// Handle to the thread accepting a connection and processing data.
+    thread: Option<JoinHandle<Result<(), String>>>,
+    /// The connected observer, if any.
+    observer: Arc<Mutex<Option<ObserverBox<T, String>>>>,
 }
 
 impl<T> TcpReceiver<T>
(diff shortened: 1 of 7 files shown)


rust/template/tcp_channel/receiver.rs
```diff
@@ -25,15 +27,55 @@ use serde::de::DeserializeOwned;
...
+                    return Err(Error::last_os_error());
+                }
+                Ok(())
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+struct Data {
```
ryzhyk: `Data` is not a great name for this data structure. `ConnectionState` maybe?
Also, can't you just add the third option to `Fd` for closed state and not need this data type at all?

d-e-s-o:
GitLab pipeline succeeded: https://gitlab.com/ddlog/differential-datalog-d-e-s-o/pipelines/87446935

rust/template/tcp_channel/receiver.rs
```diff
@@ -46,6 +88,10 @@ where
     /// the system will assign a port that is free. To retrieve this
     /// assigned port (in the form of the full `SocketAddr`), use the
     /// `addr` method.
+    ///
+    /// Note that the receiver may require some time after creation
```
ryzhyk: Why is that? You will have called `bind` by the time this method returns, so it should be ok for `TcpSenders` to connect. The connections will be processed once the listener thread calls `accept`. I might be wrong.

rust/template/tcp_channel/receiver.rs
```diff
@@ -122,14 +198,22 @@ where
...
+            // TODO: We probably want to just log failures.
+            guard.fd.close().unwrap();
         }
+
+        // TODO: We probably want to log any errors reported by the
+        //       thread being joined.
+        self.thread
+            .take()
+            .map(JoinHandle::join)
+            .unwrap()
```
ryzhyk: Won't this crash if the receiver thread returns an error?

test/datalog_tests/server_api/tests/deltas.rs
```diff
@@ -1,58 +1,113 @@
...
+
+macro_rules! DeltaTest {
+    ( $($setup_fn:tt)* ) => {{
+        // We need a global here (and cannot, say, close over some local
+        // from a lambda expression) because our callback has to implement
+        // Clone and that is not possible for a closure capturing its
+        // environment in a mutable way.
+        // We need an option because Vec::new is not const and so we can't
+        // use it to initialize a static.
+        static mut DELTAS: Option<Mutex<Vec<(usize, Record)>>> = None;
```
ryzhyk: would `lazy_static` help here?

rust/template/tcp_channel/receiver.rs
d-e-s-o: Sure, I'll change the name. Yes, a third state may be possible. I'll try it out. I iterated through so many variants that when I finally had something working I did not think too hard to reduce the state even more.

rust/template/tcp_channel/receiver.rs
d-e-s-o: Based on my observations my conclusion was that until `accept` is issued you'd still get a connection refused when trying to connect. The mere act of `bind`ing only prepares the socket and "reserves" a port. Or so that was my conclusion. I'll verify tomorrow.

rust/template/tcp_channel/receiver.rs
d-e-s-o: Yeah. For now I concluded that this is better than not having any signalling that something went wrong (which was the cause of a long debug session already). I'll change this to behave more friendly once I added logging.

test/datalog_tests/server_api/tests/deltas.rs
d-e-s-o: I tried it, but the compiler threw some weird errors and my conclusion was that it does not deal with `mut` state, it's merely for initializing constants. But let me retry, it's possible I screwed up.

rust/template/tcp_channel/receiver.rs
ryzhyk: From `TcpListener` documentation:

```
After creating a TcpListener by binding it to a socket address, it listens for incoming TCP connections. These can be accepted by calling accept or by iterating over the Incoming iterator returned by incoming.
```

What I think this means is that `TcpListener.bind()` internally calls `listen()` from `sys/socket.h`, which in turn means that the socket should already be queuing incoming connections.

BTW, I think there is nothing wrong with `TcpSender` retrying regardless of this; just trying to figure out what's going on...

rust/template/tcp_channel/receiver.rs
ryzhyk: As long as there is a big FIXME in the code :)

test/datalog_tests/server_api/tests/deltas.rs
ryzhyk: `DELTAS` does not have to be `mut`, does it? It's only `mut` here so you can initialize it. This is exactly what `lazy_static` is supposed to solve.

rust/template/tcp_channel/receiver.rs
d-e-s-o: Yeah, I tried it out and it works as you suspected it would. I agree that eventually we may need a `retry_connect` but I guess for now we don't.

test/datalog_tests/server_api/tests/deltas.rs
d-e-s-o: It does have to mutable, because we accumulate the deltas in this vector. It works as long as there is no `mut` involved, but once I change the declaration the compiler bails out:
```
error: no rules expected the token `mut`
 --> src/main.rs:6:12
  |
6 |     static mut ref DELTAS: Mutex<Vec<usize>> = {
  |            ^^^ no rules expected this token in macro call
```

```
error: no rules expected the token `DELTAS`
 --> src/main.rs:6:20
  |
6 |     static ref mut DELTAS: Mutex<Vec<usize>> = {
  |                    ^^^^^^ no rules expected this token in macro call
```

d-e-s-o: We clarified that in person. I'll change the code to use `lazy_static` :)

d-e-s-o:
I've addressed all review comment. New gitlab pipeline: https://gitlab.com/ddlog/differential-datalog-d-e-s-o/pipelines/87724508

Thanks for the review!

PR #419, opened 2019-10-09
Remove usage of unsafe_abomonate

### LINKED ISSUE #409: warning: use of deprecated item 'unsafe_abomonate': please use the abomonation_derive crate
We discussed that already over Slack but let me file an issue here to not forget and so whoever feels up to it may just take it.

Since Rust 1.38 we get the following deprecation notice:
```
warning: use of deprecated item 'unsafe_abomonate': please use the abomonation_derive crate
   --> src/lib.rs:112:1
    |
112 | unsafe_abomonate!(Value);
    | ^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default
```

We probably want to adhere to it.

d-e-s-o:
The unsafe_abomonate macro has been deprecated in version 0.5 of the
crate and so we should probably refrain from using it. The
recommendation is to use abomonate_derive but the way we used the macro
we should just be able to directly implement the Abomonation trait for
the types we used the macro for previously, as that is exactly what the
macro did. This change fixes #409.

### DIFF
--- a/rust/template/Cargo.toml
+++ b/rust/template/Cargo.toml
@@ -32,7 +32,7 @@ path = "./observe"
 path = "./tcp_channel"
 
 [dependencies]
-abomonation= "0.7"
+abomonation = "0.7"
 cpuprofiler = "0.0.3"
 differential-dataflow = "0.9"
 fnv="1.0.2"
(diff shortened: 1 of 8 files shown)


d-e-s-o:
Gitlab pipeline: https://gitlab.com/ddlog/differential-datalog-d-e-s-o/pipelines/87707095

PR #420, opened 2019-10-09
Remove unwrap from ddlog_dump_record C API

d-e-s-o:
This change replaces the unwrap in ddlog_dump_record with a NULL pointer
return. I don't think we could ever encounter a panic there, because we
convert the record to a string using escape_default internally, which
will escape NUL bytes appropriately. But better safe than sorry
(especially if the implementation ever changes) and doing it properly
prevents others from pointing out the issue over and over.


d-e-s-o:
https://gitlab.com/ddlog/differential-datalog-d-e-s-o/pipelines/87738756

PR #421, opened 2019-10-09
Support for tuple fields; fixes #13

### LINKED ISSUE #13: syntax to select tuple fields by number
x == (x,y).0
y == (x,y).1
z == (x,y,z).2
...

mbudiu-vmw:


doc/tutorial/tutorial.md
```diff
@@ -1134,12 +1133,21 @@ Note the assignment that binds variables `b2` and `b3` to individual fields of t
...
 ```
-output relation IntranetHost(addr: ip4_addr_t)
-
 IntranetHost(addr) :- KnownHost(addr),
                       (192, 168, _, _) = addr_to_tuple(addr).
 ```
 
+The fields of a tuple can be accessed using the record field access
+syntax but using numbers instead of field names: `.0` is the 0-th
+field of the tuple (reading from the left).  The previous relation can
```
ryzhyk: "relation" -> "rule"

src/Language/DifferentialDatalog/Syntax.hs
```diff
@@ -551,6 +552,7 @@ instance Show Rule where
 data ExprNode e = EVar          {exprPos :: Pos, exprVar :: String}
                 | EApply        {exprPos :: Pos, exprFunc :: String, exprArgs :: [e]}
                 | EField        {exprPos :: Pos, exprStruct :: e, exprField :: String}
+                | ETupField     {exprPos :: Pos, exprTuple :: e, exprTupField :: Integer}
```
ryzhyk: This can be `Int`, no need for an `Integer`

src/Language/DifferentialDatalog/Type.hs
```diff
@@ -275,6 +274,16 @@ exprNodeType' d _   (EField p (Just e) f) = do
                                  return $ fieldType fld
          _           -> err (pos e) $ "Expression is not a struct"
 
+
+exprNodeType' _ ctx (ETupField p Nothing _)  = eunknown p ctx
+exprNodeType' d _   (ETupField p (Just e) i) = do
+    case typ' d e of
+         t@TTuple{} ->
+             if ((toInteger $ length $ typeTupArgs t) <= i) || (0 > i)
```
ryzhyk: You should be able to use `check` instead of if-then-else here. It's shorter and more idiomatic

src/Language/DifferentialDatalog/Type.hs
```diff
@@ -275,6 +274,16 @@ exprNodeType' d _   (EField p (Just e) f) = do
...
+             then err p  $ "Tuple \"" ++ show t ++ "\" does not have of " ++ show i ++ " fields"
+             else return $ (typeTupArgs t) !! (fromIntegral i)
```
ryzhyk: `fromIntegral` won't be necessary if you use `Int` for `exprTupleField`

PR #426, opened 2019-10-11
Topic/logging

d-e-s-o:
This pull request adds the basic infrastructure for logging to the code base and converts a few unwraps into log messages.

### DIFF
--- a/rust/template/tcp_channel/message.rs
+++ b/rust/template/tcp_channel/message.rs
@@ -1,3 +1,7 @@
+use std::fmt::Display;
+use std::fmt::Formatter;
+use std::fmt::Result;
+
 use serde::Deserialize;
 use serde::Serialize;
 
@@ -9,4 +13,16 @@ pub enum Message<T> {
     Updates(Vec<T>),
     Commit,
     Complete,
+}
+
+impl<T> Display for Message<T> {
+    fn fmt(&self, formatter: &mut Formatter) -> Result {
+        let s = match self {
+            Message::Start => "on_start",
+            Message::Updates(_) => "on_updates",
+            Message::Commit => "on_commit",
+            Message::Complete => "on_completed",
+        };
+        formatter.write_str(s)
+    }
 }
--- a/rust/template/tcp_channel/receiver.rs
+++ b/rust/template/tcp_channel/receiver.rs
@@ -19,6 +19,8 @@ use libc::close;
 use libc::shutdown;
 use libc::SHUT_RDWR;
 
+use log::error;
+
 use observe::Observable;
 use observe::ObserverBox;
 use observe::Subscription;
@@ -164,11 +166,11 @@ where
         loop {
             let message = match deserialize_from(&mut reader) {
                 Ok(m) => m,
-                Err(_) => {
+                Err(e) => {
                     if let Fd::Closed = *fd.lock().unwrap() {
                         return Ok(());
                     }
-                    // TODO: Can/should we log the error?
+                    error!("failed to deserialize message: {}", e);
                     continue;
                 }
             };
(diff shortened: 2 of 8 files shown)


d-e-s-o:
https://gitlab.com/ddlog/differential-datalog-d-e-s-o/pipelines/88319861

rust/template/tcp_channel/message.rs
```diff
@@ -10,3 +14,15 @@ pub enum Message<T> {
     Commit,
     Complete,
 }
+
+impl<T> Display for Message<T> {
```
ryzhyk: I'm not sure this is a good `Display` implementation. In the future we may want a "proper" `Display` that shows all updates rather than just print the name of the command.

rust/template/tcp_channel/receiver.rs
```diff
@@ -39,22 +44,33 @@ enum Fd {
 }
 
 impl Fd {
+    fn shutdown(self, fd: RawFd) -> Result<(), Error> {
```
ryzhyk: Why does this method take `self`?

rust/template/tcp_channel/receiver.rs
```diff
@@ -39,22 +44,33 @@ enum Fd {
...
+        let rc = unsafe { shutdown(fd, SHUT_RDWR) };
+        if rc != 0 {
+            return Err(Error::last_os_error());
+        }
+        Ok(())
+    }
+
+    fn close_(self, fd: RawFd) -> Result<(), Error> {
```
ryzhyk: Same here

rust/template/tcp_channel/receiver.rs
```diff
@@ -190,16 +221,17 @@ where
...
-            .unwrap();
+        if let Err(e) = self.fd.lock().unwrap().close() {
+            error!("failed to close TcpReceiver file descriptor: {}", e);
+        }
+
+        if let Some(t) = self.thread.take() {
+            match t.join() {
+                Ok(Ok(())) => (),
+                Ok(Err(e)) => error!("TcpReceiver accept thread failed: {}", e),
+                Err(e) => error!("TcpReceiver thread has paniced: {:?}", e),
```
ryzhyk: Typo: should be "panicked"

rust/template/tcp_channel/receiver.rs
d-e-s-o: Because the type implements `Copy` and Clippy kindly asks us to pass 8-byte-or-less values by value and not by reference.

ryzhyk: But you never actually use `self`in this function

rust/template/tcp_channel/message.rs
d-e-s-o: I understand what you are saying, but please consider that this is solely an internal type (with no intention/reason of ever being exposed outside the crate) and we have full control over the few locations it is used in within this crate. I deliberately picked `Display` because `Debug` would be more appropriate for producing a dump that includes updates, should the need arise. In my opinion there is no wrong or right implementation of `Display` here. Yes, we could introduce a "to_type_str" method or something like that to avoid using `Display`, but at this point it feels as if that just adds boilerplate code.
Just my opinion, do let me know if you disagree and I'll change it, though.

rust/template/tcp_channel/receiver.rs
d-e-s-o: I'll change it. Thanks!

rust/template/tcp_channel/message.rs
ryzhyk: Ok, I did not realize this was internal to tcpchannel

rust/template/tcp_channel/receiver.rs
d-e-s-o: Well, it's a method so invocation happens through the object and it is hence implicitly available. The alternative is to declare it as `fn shutdown(fd: RawFd)` but that means invocation is awkward as `Self::shutdown(fd)` instead of `self.shutdown(fd)`. Or am I misunderstanding what you are saying?

d-e-s-o:
I've merged the pull request because the last question is really minor. But we can chat on Monday and I can always include a fix in one of the next commits.

PR #429, opened 2019-10-15
Read and write ovsdb2ddlog configuration files from json

### LINKED ISSUE #393: Control `ovsdb2ddlog` via a config file
This avoids huge command lines. In addition, the config file can be declared as a make dependency, making sure that the DDlog schema gets re-generated when configuration changes.

Here's the relevant code:
https://github.com/vmware/differential-datalog/blob/master/adapters/ovsdb/Main.hs

mbudiu-vmw:
Fixes #393 
I have tested this by generating json files for ovn and using them in the Makefiles.
For example, this is the nb file generated:
```json
{
  "proxyTables": [],
  "keys": {
    "Logical_Switch_Port": [
      "name"
    ]
  },
  "outputTables": [
    [
      "NB_Global",
      [
        "ssl",
        "connections",
        "external_ids",
        "nb_cfg"
      ]
    ],
    [
      "Logical_Switch_Port",
      []
    ]
  ],
  "ovsFile": "ovn-nb.ovsschema"
}
```

### DIFF
--- a/adapters/ovsdb/Main.hs
+++ b/adapters/ovsdb/Main.hs
@@ -34,7 +34,7 @@ import Data.Maybe
 import Data.List.Split
 import Control.Monad
 import qualified Data.Map as M
-import Data.Aeson (FromJSON, ToJSON, decode, encode)
+import Data.Aeson (FromJSON, ToJSON, eitherDecode, encode)
 import GHC.Generics (Generic)
 import qualified Data.ByteString.Lazy.Char8 as LZ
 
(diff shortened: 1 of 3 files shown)


mbudiu-vmw:
One annoying thing about JSON is that the format does not support comments, so it's not possible to comment a configuration file to help other people understand it.

mbudiu-vmw:
Travis has failed due to a timeout, but this seems to pass.

adapters/ovsdb/Main.hs
```diff
@@ -41,81 +46,93 @@ data TOption = OVSFile     String
...
                  $ errorWithoutStackTrace $ "Unknown output table name " ++ table
             let outtabs = map (\(t,ro) -> if t == table then (t, nub $ col:ro) else (t,ro))
-                              $ confOutputTables config
-            return $ config{confOutputTables = outtabs}
+                              $ outputTables config
+            return $ (a, config{outputTables = outtabs})
          _ -> errorWithoutStackTrace $ "Invalid column name " ++ c
+addOption (a, _) (ConfigJsonI c) = do
+    cdata <- readFile c
+    return $ (a, fromJust $ ((decode $ LZ.pack cdata) :: Maybe Config))
```
ryzhyk: This should use `eitherDecode` instead of `decode` and report an error using `errorWithoutStackTrace`. Otherwise, the program will just crash without a meaningful error message.

adapters/ovsdb/Main.hs
```diff
@@ -41,81 +46,93 @@ data TOption = OVSFile     String
...
+addOption (a, config) (ConfigJsonO c) = do
+    writeFile c $ LZ.unpack $ encode config
+    return (a, config)
 
-validateConfig :: Config -> IO ()
-validateConfig Config{..} = do
-    when (confOVSFile == "" && confAction == ActionCompile) $ errorWithoutStackTrace "Input file not specified"
+validateConfig :: (Action, Config) -> IO ()
+validateConfig (action, Config{..}) = do
```
ryzhyk: This needs to additionally validate that the user does not specify both `-c` and one of the other options that will end up being ignored.

adapters/ovsdb/Main.hs
```diff
@@ -41,81 +46,93 @@ data TOption = OVSFile     String
...
 
-data Config = Config { confAction       :: Action
-                     , confOVSFile      :: FilePath
-                     , confOutputTables :: [(String, [String])]
-                     , confProxyTables  :: [String]
-                     , confKeys         :: M.Map String [String]
+data Config = Config { ovsFile      :: FilePath
+                     , outputTables :: [(String, [String])]
+                     , proxyTables  :: [String]
+                     , keys         :: M.Map String [String]
```
ryzhyk: More user-friendly names: `ovsFile` -> `ovsschemaFile`, `keys` -> `keyColumns`.

adapters/ovsdb/Main.hs
mbudiu-vmw: This is done in addOption for ConfigJsonI.

ryzhyk: But that only helps if -c occurs after other options, bit before

mbudiu-vmw: I think this behavior is very sensible: essentially options are processed from left to right, and you can load a configuration file and override some of its settings.

PR #435, opened 2019-10-18
Topic/rework subscription

d-e-s-o:
This pull request changes our subscribe/unsubscribe approach for the observable/observer infrastructure. The main changes are:
- remove `Subscription` trait
- move `unsubscribe` functionality into `Observable` itself
- rework `subscribe` & `unsubscribe` APIs to make them more useable

As usual, please look at the individual commits to get a better picture.

PR #434, opened 2019-10-16
Fix race condition causing unrelated socket closures

d-e-s-o:
Our tests have found a race condition which manifests in broken pipe or
connection refused errors when running tests concurrently. The
explanation, in a nutshell, is that with the current code we may issue a
close(2) on one file descriptor twice and any multiple closes, given the
right timing, could actually close an unrelated file descriptor if the
number has been reused by the system.
To fix this problem, this change introduces a second mutex to the
TcpReceiver's file descriptor state that synchronizes system calls with
respect to closes.
I verified the fix by running our unit tests for half an hour or so.
Previously, such a race condition would be hit after less than 10s in
most cases.
This change fixes #427.

PR #436, opened 2019-10-18
Topic/rework subscription new

d-e-s-o:
This pull request changes our subscribe/unsubscribe approach for the observable/observer infrastructure. The main changes are:
- remove `Subscription` trait
- move `unsubscribe` functionality into `Observable` itself
- rework `subscribe` & `unsubscribe` APIs to make them more useable

As usual, please look at the individual commits to get a better picture.

---

This is a resubmission of https://github.com/vmware/differential-datalog/pull/435 which got closed when https://github.com/vmware/differential-datalog/pull/434 got merged because it was based on a branch that the latter used and that was deleted automatically. For some reason GitHub won't let me pick a different base branch on the same pull request so opening a new one instead.


rust/template/distributed_datalog/observe/src/observable.rs
```diff
@@ -34,17 +43,80 @@ where
...
-                    observer: self.observer.clone(),
-                }))
+                Ok(())
             }
         }
     }
+
+    fn unsubscribe(&mut self, _subscription: &Self::Subscription) -> Option<ObserverBox<T, E>> {
+        let mut guard = self.observer.lock().unwrap();
+        match *guard {
```
ryzhyk: Can you use `swap(guard, &None)` or something like that instead of the `match`?

d-e-s-o: Thanks for taking a look! As part of the changes I have piled up (which will go out later today or on Monday) the `match` will go and the `unsubscribe` will actually become a one-liner:
```rust
self.observer.lock().unwrap().take()
```
So I'd like to just leave that be right now, as it's merely a transitional state.

PR #437, opened 2019-10-18
Topic/txnmux

d-e-s-o:
This change implements a transaction multiplexer for observables. Basically the moment we have `n` observables pushing updates to a single observer (e.g., datalog server/program) we may run into problems because only a single transaction can be active on an `HDDlog`.
To handle those cases we need to ensure that transactions don't interleave, buffering updates as required and then pushing them atomically.

### DIFF
--- a/rust/template/distributed_datalog/observe/src/observable.rs
+++ b/rust/template/distributed_datalog/observe/src/observable.rs
@@ -1,8 +1,8 @@
 use std::fmt::Debug;
-use std::sync::Arc;
-use std::sync::Mutex;
 
 use crate::observer::ObserverBox;
+use crate::observer::OptionalObserver;
+use crate::observer::SharedObserver;
 
 /// A trait for objects that can be observed.
 pub trait Observable<T, E>: Debug
@@ -32,10 +32,10 @@ where
 
 /// A very simple observable that supports subscription of a single
 /// observer.
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct UpdatesObservable<T, E> {
     /// A reference to the `Observer` subscribed to us, if any.
-    pub observer: Arc<Mutex<Option<ObserverBox<T, E>>>>,
+    pub observer: SharedObserver<OptionalObserver<ObserverBox<T, E>>>,
 }
 
 impl<T, E> Observable<T, E> for UpdatesObservable<T, E>
@@ -50,21 +50,16 @@ where
         observer: ObserverBox<T, E>,
     ) -> Result<Self::Subscription, ObserverBox<T, E>> {
         let mut guard = self.observer.lock().unwrap();
-        match *guard {
-            Some(_) => Err(observer),
-            None => {
-                *guard = Some(observer);
-                Ok(())
-            }
+        if guard.is_some() {
+            Err(observer)
+        } else {
+            guard.replace(observer);
+            Ok(())
         }
     }
 
     fn unsubscribe(&mut self, _subscription: &Self::Subscription) -> Option<ObserverBox<T, E>> {
-        let mut guard = self.observer.lock().unwrap();
-        match *guard {
-            Some(_) => guard.take(),
-            None => None,
-        }
+        self.observer.lock().unwrap().take()
     }
 }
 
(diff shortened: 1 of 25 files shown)


rust/template/distributed_datalog/observe/src/observer.rs
```diff
@@ -64,3 +112,178 @@ where
...
+            self.observer.on_commit()?;
+        } else {
+            panic!("on_commit was not preceded by an on_start event")
+        }
+        Ok(())
+    }
+
+    fn on_updates<'a>(&mut self, updates: Box<dyn Iterator<Item = T> + 'a>) -> Result<(), E> {
+        if let Some(ref mut data) = self.data {
+            data.append(&mut updates.collect());
```
ryzhyk: I wonder if `LinkedList` would be a better choice than `Vec`, which requires reallocation as it grows.

rust/template/distributed_datalog/src/txnmux.rs
```diff
@@ -0,0 +1,97 @@
...
+    /// Create a new `TxnMux`, without any observables.
+    pub fn new() -> Self {
+        Self {
+            subscriptions: Vec::new(),
+            observer: SharedObserver::new(OptionalObserver::default()),
+        }
+    }
+
+    /// Add an `Observable` to track to the multiplexer.
+    pub fn add_observable(
```
ryzhyk: Interesting, this is a new, non-standard way to connect `Observable` to an object that does not implement the `Observer` interface. A potential problem here is how the control plane will "know" how to instantiate such a graph. We can discuss in person.

rust/template/distributed_datalog/observe/src/observable.rs
```diff
@@ -30,12 +34,54 @@ where
     fn unsubscribe(&mut self, subscription: &Self::Subscription) -> Option<ObserverBox<T, E>>;
 }
 
+/// A trait abstracting away from the concrete type of subscription used
```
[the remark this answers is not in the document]
d-e-s-o: Basically, the `TxnMux` accepts and stores arbitrary observables. However, observables now come with a subscription type in which they can differ, so we can't treat them equally unless we prescribe the subscription to use. I wanted to keep things generic and so instead abstracted away those differences behind an `Any`, which is what `ObservableAny` can subscribe/unsubscribe.

rust/template/distributed_datalog/observe/src/observer.rs
d-e-s-o: Sure.

rust/template/distributed_datalog/src/txnmux.rs
d-e-s-o: Right, I did not see a good way to fudge that with the observable/observer infrastructure. The control plane would need knowledge of this interface.

I think in the future we may want to have an "integrated" multiplexer. Basically, that could be a TCP endpoint that repeatedly instantiates an observable (`TcpReceiver`) for each accepted connection. So there would be less control plane involvement.

rust/template/distributed_datalog/observe/src/observable.rs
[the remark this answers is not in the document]
ryzhyk: Understood, thank you.

rust/template/distributed_datalog/src/txnmux.rs
ryzhyk: Another approach is to have the multiplexer expose multiple observer instances. But this would still require a non-standard interface to allocate a new observer, so it's not a clear win to me.

d-e-s-o: Let's discuss some more tomorrow.

rust/template/distributed_datalog/observe/src/observer.rs
d-e-s-o: As discussed, I went with `LinkedList<Vec<T>>`. The implementation of `Vec::from_iter` (which is used in `collect` we call) properly honors any size hint of the passed in iterator already. See https://doc.rust-lang.org/src/alloc/vec.rs.html#1909.

PR #439, opened 2019-10-25
Topic/txnmux

d-e-s-o:
When run frequently, the "deltas" test fails on occasion with an error
indicating that an attempt was made to start a new transaction when one
is already in progress.
The problem here is that the initial idea was to have the
CachingObserver be invoked while holding a mutex protecting the overall
observer. However, that turned out to be not possible and so we ended up
invoking methods on the inner observer that by themselves were protected
by a lock but not as a whole -- so interleaving of transactions was
still possible.
This pull request fixes the problem by making the CachingObserver aware that
it contains a SharedObserver inside and protecting the overall
transaction by its lock. Because it is less general now I decided to
make it private to the txnmux module.
It also includes changes to update all our template's crates to use Rust Edition 2018.

### DIFF
--- a/rust/template/cmd_parser/Cargo.toml
+++ b/rust/template/cmd_parser/Cargo.toml
@@ -1,15 +1,16 @@
 [package]
 name = "cmd_parser"
 version = "0.1.0"
+edition = "2018"
 
 [dependencies.differential_datalog]
 path = "../differential_datalog"
 
 [dependencies]
+libc = "0.2.42"
 nom = "4.0"
 num = "0.2"
 rustyline = "1.0.0"
-libc = "0.2.42"
 
 [lib]
 name = "cmd_parser"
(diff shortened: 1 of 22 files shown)


rust/template/cmd_parser/Cargo.toml
```diff
@@ -1,15 +1,16 @@
 [package]
 name = "cmd_parser"
 version = "0.1.0"
+edition = "2018"
 
 [dependencies.differential_datalog]
 path = "../differential_datalog"
 
 [dependencies]
+libc = "0.2.42"
```
ryzhyk: Do we need to specify the minor version of each package given that after 1.0 they should maintain compatibility across minor versions?

d-e-s-o: I don't think so. I was not sure about this case, though. Typically that is done if a new feature was introduced and we want to make sure that a version with this feature is consumed. I don't have the background to judge whether that was the case for `libc` here.
As for the others, I don't think it makes a difference here ([reference](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements)). Do you want me to shorten the numbers where possible?

PR #441, opened 2019-10-28
Compile.hs: Avoid name clash in aggregation.

ryzhyk:
We generated variable named `group` when compiling the `Aggregate`
operator, which would clash with a user-defined variable with the same
name in, e.g.,:

```
relation GroupConstraints(group: groupid_t, constraints: Vec<constraint_t>)

GroupConstraints(group, constraints) :-
    GroupConstraint(group, constraint),
    var constraints = Aggregate((group), group2vec(constraint)).
```

Rename the generated variable to `__group__` instead.

### DIFF
--- a/src/Language/DifferentialDatalog/Compile.hs
+++ b/src/Language/DifferentialDatalog/Compile.hs
@@ -97,7 +97,7 @@ bOX_VAR = "__box"
 
 -- Input argument to aggregation function
 gROUP_VAR :: Doc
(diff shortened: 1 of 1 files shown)


ryzhyk:
> Unless you reserve this name this could happen again. The only safe way is to generate a fresh name.

or disallow user-defined variables that start with `__`. I agree, but for now this will have to do...

PR #444, opened 2019-10-29
Introduce profile Cargo feature

d-e-s-o:
We have two ways to profile programs, that serve slightly different
purposes and work differently: the `enable_cpu_profiling` API enables
profiling of differential operators and the cpuprofiler based approach
allows for creating a detailed callstack-style profile eligible to be
analyzed with pprof.
While we want to keep the two, the cpuprofiler infrastructure has a few
disadvantages in the way it is hooked up:
1) It is hard to find because all references to it are a dependency to
   the cpuprofiler crate and a few comments in some source file
2) It is cumbersome to enable because various lines have to be commented
   out. Those lines are also not compiled, making them an easy target
   for bit rot.
3) The cpuprofiler crate (and its dependencies) are compiled
   unconditionally, for the case that a user would use this
   functionality.

This patch addresses those three problems (to varying degrees) by
introducing another Cargo feature, `profile`. With it, we can compile the
dependent upon crate conditionally, have a clearly visible and fairly
easily discoverable switch, and don't need to do anything else than
enable the feature to make this functionality available.
According to my testing this feature switch reduces the number of crates
we compile by default from 79 down to 68.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -38,7 +38,11 @@ test-server_api:
       done
 
 test-simple:
-    script: stack --no-terminal test --ta "-p simple"
+    script:
+    - stack --no-terminal test --ta "-p simple"
+    - apt-get update
+    - apt-get install --assume-yes libgoogle-perftools-dev
+    - cd test/datalog_tests/simple_ddlog/ && LIBRARY_PATH=/usr/lib/x86_64-linux-gnu/ cargo build --features='profile'
 
 test-modules:
     script: stack --no-terminal test --ta "-p modules"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -40,8 +40,6 @@ test-server_api:
 test-simple:
     script:
     - stack --no-terminal test --ta "-p simple"
-    - apt-get update
-    - apt-get install --assume-yes libgoogle-perftools-dev
     - cd test/datalog_tests/simple_ddlog/ && LIBRARY_PATH=/usr/lib/x86_64-linux-gnu/ cargo build --features='profile'
 
 test-modules:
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -38,7 +38,11 @@ test-server_api:
       done
 
 test-simple:
-    script: stack --no-terminal test --ta "-p simple"
+    script:
+    - stack --no-terminal test --ta "-p simple"
+    - apt-get update
+    - apt-get install --assume-yes libgoogle-perftools-dev
```
ryzhyk: Any objections if I add this to Dockerfile instead?

d-e-s-o: No objection. We can also leave it out entirely if you think it doesn't help much and isn't worth the effort.

ryzhyk: Pushed to you branch. Hope it's ok.

.gitlab-ci.yml
```diff
@@ -38,7 +38,11 @@ test-server_api:
...
+    - cd test/datalog_tests/simple_ddlog/ && LIBRARY_PATH=/usr/lib/x86_64-linux-gnu/ cargo build --features='profile'
```
ryzhyk: Let's try to also do something about this so we don't compile it twice. @mbudiu-vmw recently added the `test/datalog_tests/run-test.sh` script that can be used instead of `stack test`.

d-e-s-o: Hm, I don't think this script will work here. For one, I don't see how the claim that it honors `CARGOFLAGS`(as stated in its header) is true, as it overwrites that variable unconditionally. But even if it did, there is the problem that Cargo's `--features` argument cannot be supplied multiple times (in 1.34, it works in 1.38 on my system). See https://gitlab.com/ddlog/differential-datalog-d-e-s-o/-/jobs/336749170

We could introduce a dedicated variable just for passing in the features to use, but that seems ugly. I am tempted to say we can just live with it. What do you think?

ryzhyk: Agreed.

Issue #342, opened 2019-09-05
Better error handling in `Program::run()`


ryzhyk:
There are a bunch of `unwrap()`'s in `Program::run()`.  Handle the errors properly by returning `Result<RunningProgram>` instead of `RunningProgram`.

d-e-s-o:
@ryzhyk A bunch of unwraps match the following pattern:
```
error: called `.get().unwrap()` on a HashMap. Using `[]` is more clear and more concise
    --> differential_datalog/program.rs:1123:110
     |
1123 |                                                     ||arrangements.insert((rel.id, i), arr.build_arrangement(collections.get(&rel.id).unwrap())));
     |                                                                                                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try this: `&collections[&rel.id]`
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#get_unwrap
```

Shall we follow `clippy`'s suggestion or do you feel that we should handle them gracefully?

ryzhyk:
I'd keep the `unwrap`, so we can easily find places where errors need to be handled.

d-e-s-o:
Returning errors instead of unwrapping is one thing, but the code seemingly can't handle that very well. For example, I injected the following error:

```patch
@@ -1232,6 +1237,8 @@
                         epoch += 1;
                         Self::advance(&mut all_sessions, epoch);
                         Self::flush(&mut all_sessions, &probe, worker, &*progress_lock, &progress_barrier);
+                        println!("INSERTING TEST ERROR #7!!!");
+                        Err(format!("THIS IS A TEST!!!"))?;
                         flush_ack_send.send(()).map_err(|e| format!("failed to send ACK: {}", e))?;
                     }

```
and the program ended up just hanging. So it seems there are some logical problems in there as well.

ryzhyk:
I think this is an instance of #134

ryzhyk:
In fact we should probably merge the two issues.  Let me delete the other one and edit the description of this issues.

d-e-s-o:
Allow me just quote from #134 to have that context directly in here:

> At the moment a panic in a worker thread, e.g., caused by an assertion in a Rust function, crashes the thread, sending the entire computation in an infinite loop. We should instead handle this situation by either exiting the process or, ideally, returning the error to the caller, so that they can report the error gracefully or even restart the computation.

PR #445, opened 2019-10-30
Topic/unwraps

d-e-s-o:
This is a draft for what will hopefully one day become a fix for issue https://github.com/vmware/differential-datalog/issues/342. First two commits should be fine. Last one is what I'd like to get an opinion on.

@ryzhyk, do you have any comments?

### DIFF
--- a/rust/template/differential_datalog/program.rs
+++ b/rust/template/differential_datalog/program.rs
@@ -1,13 +1,13 @@
 //! Datalog program.
 //!
 //! The client constructs a `struct Program` that describes Datalog relations and rules and
-//! calls `Program::run()` to instantiate the program.  The method returns an instance of
-//! `RunningProgram` that can be used to interact with the program at runtime.  Interactions
-//! include starting, committing or rolling back a transaction and modifying input relations.
-//! The engine invokes user-provided callbacks as records are added or removed from relations.
-//! `RunningProgram::stop()` terminates the Datalog program destroying all its state. If not
-//! invoked manually (which allows for manual error handling), `RunningProgram::stop` will be
-//! called when the program object leaves scope.
+//! calls `Program::run()` to instantiate the program.  The method returns an error or an
+//! instance of `RunningProgram` that can be used to interact with the program at runtime.
+//! Interactions include starting, committing or rolling back a transaction and modifying input
+//! relations. The engine invokes user-provided callbacks as records are added or removed from
+//! relations. `RunningProgram::stop()` terminates the Datalog program destroying all its state.
+//! If not invoked manually (which allows for manual error handling), `RunningProgram::stop`
+//! will be called when the program object leaves scope.
 
 // TODO: namespace cleanup
 // TODO: single input relation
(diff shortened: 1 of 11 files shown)


rust/template/differential_datalog/program.rs
```diff
@@ -1471,8 +1504,24 @@ impl<V: Val> Program<V> {
     }
 
     /* Returns all input relations of the program */
-    fn input_relations(&self) -> Vec<RelId> {
+    fn input_relations(&self) -> Response<Vec<RelId>> {
```
ryzhyk: I think this one is a bit too much on the paranoid side. I suggest we use `assert!()` instead of returning the error and keep the return type of this function as `Vec<RelId>`.

rust/template/differential_datalog/program.rs
```diff
@@ -1853,7 +1896,7 @@ impl<V: Val> RunningProgram<V> {
     /// Rollback the transaction, undoing all changes.
     pub fn transaction_rollback(&mut self) -> Response<()> {
         if !self.transaction_in_progress {
-            return Err("transacion_rollback: no transaction in progress".to_string());
+            return Err("transaction_rollback: no transaction in progress".to_string());
```
ryzhyk: clearly, I cannot spell

rust/template/src/main.rs
```diff
@@ -205,8 +205,15 @@ pub fn main() {
...
+        Err(err) => {
+            eprintln!("Failed to run differential datalog: {}", err);
+            1
+        }
+    }
+}
 
-    let ret = run(hddlog, args.delta);
-    exit(ret);
+pub fn main() {
```
ryzhyk: Can we use this trick from Rust 2018?
```
// Rust 2018
use std::fs::File;

fn main() -> Result<(), std::io::Error> {
    let f = File::open("bar.txt")?;

    Ok(())
}
```

rust/template/differential_datalog/program.rs
```diff
@@ -1132,14 +1132,12 @@ impl<V: Val> Program<V> {
                                      * updated collections returned from the inner scope. */
                                     for r in rels.iter() {
                                         let (session, collection) = outer.new_collection::<V,Weight>();
-                                        if r.input {
-                                            panic!("input relation in nested scope: {}", r.name)
-                                        }
+                                        assert!(r.input, "input relation in nested scope: {}", r.name);
```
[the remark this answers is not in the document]
d-e-s-o: Ugh, yeah. Everything went lit up red.

rust/template/src/main.rs
d-e-s-o: Yeah, I suppose we can. I tried to stay away from it because based on my own experience (in at least three small projects) this API in its current form is insufficient in so many ways and I always ended up going back to printing an error and exiting manually for various reasons. But here it may be okay. I'll change it.

d-e-s-o: [I decided](https://github.com/vmware/differential-datalog/pull/445/commits/cdaf2a6b01fd61c142f0a17baf136b92f3b90d58) to loop through a `Result` everywhere instead of dealing with `i32` values. I believe that makes the logic somewhat nicer.

d-e-s-o:
I've included one more [small bug fix](https://github.com/vmware/differential-datalog/pull/445/commits/dcfdcbf5a8d1c61e33c56aea3176653e88a42a74) and updated the CLI error reporting rework slightly, to not print errors when we also return them.

As per our discussion, I am moving this pull request out of the draft status..

PR #446, opened 2019-11-01
Retry tcp_channel::receiver::never_accepted on successful connection

d-e-s-o:
GitLab has found a case (and @ryzhyk reported it) where the
never_accepted test fails because the connection succeeded. The only
explanation I can come up with that would cause this behavior is, if
right after the drop of the TcpReceiver object and the
TcpStream::connect attempt a test running in parallel opens a listener
socket and gets assigned the very same port by random chance.

To prevent this problem from causing instability in the pipeline, let's
retry up to ten times if we actually succeed to connect. The chances of
such an even occurring ten times in a row should be extremely small.


d-e-s-o:
Not really sure what is going on here with `rustfmt`. Investigating.

d-e-s-o:
Fixed now.

PR #451, opened 2019-11-06
DDlog 0.10.3: Eager compaction support.

ryzhyk:
After a couple of months, the eager compaction feature seems to be stable and ready to be merged into master and released.  The main change here is that we use the latest timely/differential master (with slightly different type signatures), which support this feature.  We also change the main loop behavior in our worker threads to continue calling `step_or_park` after all data for the last timestamp has been processed.  Worker 0 additionally checks for new API commands, and the API makes sure to unpark worker 0 after sending such a command.

This PR also includes several memory-saving features that require the new DD, such as 32-bit timestamps and optimized SCC computation.

@d-e-s-o, @mbudiu-vmw, this PR is heavily DD-specific and will be hard to follow, as it's mostly minor changes throughout the code base, but can you do a quick sanity check pass anyway (I want to release this later today)? Thanks!


.gitlab-ci.yml
```diff
@@ -150,21 +154,27 @@ test-imported-souffle-tests7:
...
         # TODO: maintain and eventually remove the list of tests.
         # TODO: use -j6 once build script is fixed
+        # TODO: use primary OVN repo
         - (cd /root &&
-          git clone https://github.com/ovn-org/ovn.git &&
+          git clone https://github.com/ryzhyk/ovn.git &&
           cd ovn && git checkout ddlog-dev-v2 &&
           ./boot.sh &&
-          ./configure --with-ddlog=/root/.local/ddlog/lib --with-ovs-source=/root/ovs &&
+          ./configure --with-ddlog=/root/.local/ddlog/lib --with-ovs-source=/root/ovs --enable-ddlog-northd-cli &&
```
mbudiu-vmw: why are we avoiding the other tests? If we should run all of them then add a TODO: run all tests.

lib/graph.rs
```diff
@@ -6,6 +6,8 @@ use differential_dataflow::collection::Collection;
 use differential_dataflow::operators::consolidate::Consolidate;
 use differential_dataflow::lattice::Lattice;
 use timely::dataflow::scopes::Scope;
+use timely::order::TotalOrder;
```
d-e-s-o: I understand we currently don't `rustfmt` `lib/`, but given that it's Rust code I do believe that we should. Mind running the program on the newly added code?

.gitlab-ci.yml
ryzhyk: Yes, we should eventually run all of them, once (if) OVN folks ever add the missing features to DDlog. There is a todo few lines above.

rust/template/Cargo.toml
```diff
@@ -32,15 +32,18 @@ path = "./ovsdb"
...
+fnv="1.0.2"
 lazy_static = "1.3"
 libc = "0.2"
 num-traits = "0.2"
 rustop = "1.0.2"
 serde = {version = "1.0", features = ["derive"]}
 time = "0.1"
-timely = "0.9"
+#timely = "0.9"
+#timely = { git = "https://github.com/TimelyDataflow/timely-dataflow" }
```
d-e-s-o: Can we clean this up or add a TODO to know why we kept it? Same comment applies to similar constellations as well.

tools/prepush.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 # This script is configured to be run by git pre-push
 
-cd rust/template/
-cargo fmt --all -- --check
-cargo check --all
-cd ..
+(cd rust/template/ && \
+cargo fmt --all -- --check && \
```
d-e-s-o: Thank you!
You may also want to include
```
cargo clippy --all -- -D warnings
```

ryzhyk: Can't run `cargo` in `lib`, as there is no `Cargo.toml` there.

rust/template/Cargo.toml
ryzhyk: This will all go away with the next DD release...

tools/prepush.sh
d-e-s-o: Sure, but my comment was about `rust/template` (I missed when you introduced the file). Anyway, probably no need to schedule yet another pipeline...

ryzhyk: Oh, I see. I'll add it to my next PR.

PR #452, opened 2019-11-06
Generic deltamap

d-e-s-o:
The changes contained in this pull request mark the next step towards making `DDlogServer` independent of the generated code.

#### Require Display to be implemented for Val

In order to remove dependencies between DDlogServer and the generated
code we want to make the struct generic over the value type to use. That
in turns requires making generic a bunch of other structs, most
importantly DeltaMap.
As it turns out DeltaMap, by virtue of hard coding Value, made the
implicit assumption that Val: Display. However, when we switch it over
to using any Val instead, that constraint is not fulfilled. This change
adds the requirement that everything claiming to implement Val also
needs to implement Display.

---
#### Make DeltaMap generic over the value type it uses

In order to disentangle DDlogServer from the generated code we need to
make it generic over the value type to use. Because DDlogServer
indirectly works with a DeltaMap, which assumes a fixed Value type, we
also need to make this type generic over the value type to use.
This change does exactly that.

### DIFF
--- a/rust/template/differential_datalog/program.rs
+++ b/rust/template/differential_datalog/program.rs
@@ -13,7 +13,7 @@
 // TODO: single input relation
 
 use std::collections::hash_map;
-use std::fmt::{self, Debug, Formatter};
+use std::fmt::{self, Debug, Display, Formatter};
 use std::hash::Hash;
 use std::ops::{Add, Deref, Mul};
 use std::result::Result;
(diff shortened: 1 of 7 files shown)


ryzhyk:
Looks good, but can we merge this change after v0.10.3 release (hopefully later today)?

d-e-s-o:
Sure.

PR #454, opened 2019-11-07
More tests

ryzhyk:
This PR consists of several extensions and improvements on the test suite

## redist_opt

This is an optimized and extended version of the `redist` test along with
a test script designed to torture DDlog by running it with different
number of workers, different settings of DIFFERENTIAL_EAGER_MERGE, and
by feeding the input data file multiple times.  The data for this test
is stored in a separate github repo:
https://github.com/ddlog-dev/redist_opt-test-data

Also contains a memory leak test that runs a simple transaction that
adds and removes a couple of records many times and makes sure that
memory footpring does not increase significantly with the number of
iterations.

## Improved OVN scale test

We used to run the OVN scale test without actually looking at its
results.  This is useful for checking for deadlocks and performance
issues, but can miss correctness bugs.

This commit relies on the modified `ovn-test-data` repo, where the test
script is modified to dump changes on each commit and that also contains
a reference output for the test.

We have to be careful in diffing the expected and actual dumps to ignore
the performance profiles part of the dump, which we want to only keep for
performance tracking.

## Minor improvements

- bug fix in `run-test.sh`
- run `rustfmt` in `lib/`

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -178,4 +178,8 @@ test-ovn-northd:
           ./configure --with-ddlog=/root/.local/ddlog/lib --with-ovs-source=/root/ovs --enable-ddlog-northd-cli &&
           make check -j1 TESTSUITEFLAGS="117-135 138-141 143-145 147-149 151-152 154-161 167-169 172 174-181 183-185 187-189 191 193-196 198-199 201-202")
         - (git clone https://github.com/ddlog-dev/ovn-test-data.git &&
-          /usr/bin/time /root/ovn/northd/ovn_northd_ddlog/target/release/ovn_northd_cli -w 2 --no-store --no-print --no-delta < ovn-test-data/ovn_scale_test_short.dat)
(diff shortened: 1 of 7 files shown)


.gitlab-ci.yml
```diff
@@ -72,7 +72,7 @@ test-redist:
     tags:
         - ddlog-ci-1
     script:
-        - STACK_CARGO_FLAGS='--release' stack --no-terminal test --ta "-p redist"
+        - STACK_CARGO_FLAGS='--release' stack test --ta '-p "$(NF) == \"generate redist\" || ($(NF-1) == \"compiler tests\" && $(NF) == \"redist\")"'
```
mbudiu-vmw: where is this NF coming from?
Maybe a comment explaining?

test/datalog_tests/redist_opt.dl
```diff
@@ -0,0 +1,142 @@
+import tinyset as TS
```
mbudiu-vmw: is it ok to make this public?

.gitlab-ci.yml
ryzhyk: done

test/datalog_tests/redist_opt.dl
ryzhyk: it's just reachability in a graph, so should be ok.

PR #455, opened 2019-11-07
Topic/d3log

d-e-s-o:
This is an early draft of the next pieces of `d3log`. It contains a preliminary and simplistic "schema" and logic to configure a node given a "description" (not quite done yet, as the `TcpReceiver` will need a tad of adjustments beforehand). It's by no means finished (and not meant to be submitted), but early comments are welcome.

### DIFF
--- a/rust/template/distributed_datalog/src/instantiate.rs
+++ b/rust/template/distributed_datalog/src/instantiate.rs
@@ -1,0 +1,330 @@
+use std::collections::BTreeMap;
+use std::collections::HashMap;
+use std::collections::HashSet;
+use std::fmt::Debug;
+use std::time::Duration;
+
+use differential_datalog::program::RelId;
+use differential_datalog::program::Update;
+use differential_datalog::record::Record;
+use differential_datalog::DDlog;
+
+use crate::observe::Observable;
+use crate::observe::UpdatesObservable;
+use crate::schema::Addr;
+use crate::schema::Member;
+use crate::schema::Node;
+use crate::schema::NodeCfg;
+use crate::schema::RelCfg;
+use crate::schema::SysCfg;
+use crate::tcp_channel::TcpReceiver;
+use crate::tcp_channel::TcpSender;
+use crate::DDlogServer;
+
+type Outputs = BTreeMap<Addr, HashSet<RelId>>;
+/// A mapping from abstract nodes to actual members in the system.
+type Assignment = BTreeMap<Node, Addr>;
+
+/// Deduce the output streaming relations we require.
+///
+/// In a nutshell, this function deduces a mapping from all relations on
+/// a node to other nodes that have relations that have this relation as
+/// input. Unfortunately doing so is rather costly, as we ultimately
+/// have to visit pretty much all relations in the assignment and check
+/// them.
+fn deduce_outputs(
+    addr: &Addr,
+    node_cfg: &NodeCfg,
+    sys_cfg: &SysCfg,
+    assignment: &Assignment,
+) -> Outputs {
+    node_cfg.keys().fold(Outputs::new(), |mut outputs, rel| {
+        sys_cfg
+            .iter()
+            .filter_map(|(uuid, node_cfg)| {
+                assignment.get(uuid).and_then(|other_addr| {
+                    if other_addr != addr {
+                        Some((other_addr, node_cfg))
+                    } else {
+                        None
+                    }
+                })
+            })
+            .for_each(|(addr, node_cfg)| {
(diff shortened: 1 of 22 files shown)


rust/template/distributed_datalog/src/schema/config.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Input is coming from a file.
+    File(PathBuf),
+    // /// An observable directly used as a source.
+    // ///
+    // /// This variant is only meant for testing purposes.
+    // #[cfg(test)]
+    // Observable(observable: Box<dyn Observable>),
+}
+
+impl Ord for Source {
```
ryzhyk: Does Rust refuse to derive those automatically?

rust/template/distributed_datalog/src/schema/config.rs
```diff
@@ -0,0 +1,140 @@
...
+}
+
+/// Configuration of input/output relationships for a single node.
+pub type InOut = BTreeMap<RelId, BTreeSet<RelInOut>>;
+
+/// The configuration of the system is comprised of input/output
+/// configurations for a certain set of nodes.
+///
+/// Each element in the set represents a single node.
+pub type Config = BTreeSet<InOut>;
```
ryzhyk: Shouldn't this be a map from node id's?

rust/template/distributed_datalog/src/schema/relations.rs
```diff
@@ -0,0 +1,21 @@
+use std::collections::BTreeSet;
```
ryzhyk: Do we really need a file per data structure? I expect this schema to change a lot initially (and eventually get transformed into a DDlog schema); you'll waste a lot of effort adding/removing/renaming these files.  This also makes it hard to see big picture compared to having all schema data structure in one place, declared in a bottom-up (or top-down) order.

rust/template/distributed_datalog/src/instantiate.rs
```diff
@@ -0,0 +1,250 @@
...
+use crate::DDlogServer;
+
+type Outputs = BTreeMap<Node, HashSet<RelId>>;
+type Assignment = BTreeMap<Node, InOut>;
+
+/// Deduce the output streaming relations we require.
+///
+/// In a nutshell, this function deduces a mapping from all relations on
+/// a node to other nodes that have relations that have this relation as
+/// input. Unfortunately doing so is rather costly, as we ultimately
```
ryzhyk: Which is why we want to do this computation incrementally in the future.  If only there was a language that would let you do that ;)

rust/template/distributed_datalog/src/schema/relations.rs
```diff
@@ -0,0 +1,21 @@
...
+
+use differential_datalog::program::RelId;
+
+/// A set of relations used in various contexts.
+pub type Relations = BTreeSet<RelId>;
```
ryzhyk: Since `RelId` may not be globally unique, what we might want instead is `BTreeSet<(Node, RelId)>`.

rust/template/distributed_datalog/src/schema/config.rs
d-e-s-o: Good point, I haven't actually tried it here. It would make sense if it just worked.

rust/template/distributed_datalog/src/schema/relations.rs
d-e-s-o: It seemed a bit too much for a single file. But I am not objecting. Will change it.

rust/template/distributed_datalog/src/instantiate.rs
d-e-s-o: Haha :D One step at a time.

rust/template/distributed_datalog/src/schema/relations.rs
d-e-s-o: Okay, that is interesting. In my world view the `RelId` *is* unique, as every node has the same view of the world based on the compiled program (which defines `RelId`s). Let's discuss that tomorrow.

rust/template/distributed_datalog/src/schema/config.rs
d-e-s-o: Potentially. Right now there is nothing that is specific about a node. All are made equal. As such, I wouldn't know what to use as key.
In my current thinking that can't be the node's address, because the assignment of a part of the computation to a member is done as part of some computation in this file (see `assign` function).

d-e-s-o:
Let me close this pull request. I'll send out the more or less finalized "schema" soon.

Issue #422, opened 2019-10-10
Optimize expression generator


ryzhyk:
The DDlog compiler currently implements a purely syntax-guided translation from
DDlog to Rust expressions.  This causes redundant cloning and ugly code +
warnings.  We need a smarter compiler that will use dataflow analysis to use
borrow instead of cloning wherever possible.

PR #456, opened 2019-11-08
Fix Rust code generation for iteration over maps.

ryzhyk:
For-loop that iterates over maps, e.g.:

```
for (x in m) { ... }
```

used to compile into:

```
for x in m.iter() { ... }
```

where `m.iter()` returns our custom `MapIter`, defined in `std.rs`,
which used to have `Iterator::Target=(&K, &V)`. As a result, type of the
Rust variable `x` is `(&K, &V)`.  Our current naive model of Rust cannot
represent such types, instead treating `x` as if it was `&(K,V)`, which
could lead to compiler errors when we manipulate `x` in the body of the
loop.

A long-term solution is to model Rust's dataflow more accurately
(https://github.com/vmware/differential-datalog/issues/422).  The
short-term fix I implemented is to re-define `MapIter` to have
`Iterator::Target = (K, V)`, which involves cloning `K` and `V`
internally, but at least it's correct.

### DIFF
--- a/lib/std.rs
+++ b/lib/std.rs
@@ -567,11 +567,11 @@ impl<'a, K: Ord, V> MapIter<'a, K, V> {
     }
 }
 
-impl<'a, K, V> Iterator for MapIter<'a, K, V> {
-    type Item = (&'a K, &'a V);
+impl<'a, K: Clone, V: Clone> Iterator for MapIter<'a, K, V> {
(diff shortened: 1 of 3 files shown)


mbudiu-vmw:
I don't know yet enough Rust to judge this, but it looks fine.

PR #459, opened 2019-11-09
FlatBuffer.hs: Replicate Rust module name generation logic of flatc.

### LINKED ISSUE #425: flatbuf: unresolved import `flatbuf_generated::ddlog::__server_api_p2`
I have the following file:
server_api_p2.dl:
```
input relation P2In[string]
output relation P2Out[string]

P2Out[s] :- P2In[s].
```
I create a Rust project as follows:
```bash
ddlog -j -i server_api_p2.dl -L../../lib/
```

And then I compile it with flatbuf support:
```bash
cargo build --features=flatbuf
```

And I get
```bash
$ cargo build --features=flatbuf
>    Compiling server_api_p2 v0.1.0 (differential-datalog/test/datalog_tests/server_api_p2_ddlog)
> error[E0432]: unresolved import `flatbuf_generated::ddlog::__server_api_p2`
>    --> src/flatbuf.rs:575:5
>     |
> 575 | use flatbuf_generated::ddlog::__server_api_p2  as fb;
>     |     ^^^^^^^^^^^^^^^^^^^^^^^^^^---------------^^^^^^^
>     |     |                         |
>     |     |                         help: a similar name exists in the module: `__server_api_p_2`
>     |     no `__server_api_p2` in `flatbuf_generated::ddlog`
> 
> error[E0432]: unresolved import `flatbuf_generated::ddlog::__server_api_p2`
>    --> src/flatbuf.rs:576:35
>     |
> 576 | pub use flatbuf_generated::ddlog::__server_api_p2 ::__String;
>     |                                   ^^^^^^^^^^^^^^^ could not find `__server_api_p2` in `ddlog`
> 
> error: aborting due to 2 previous errors
```

And indeed the module is named `__server_api_p_2` and not `__server_api_p2`.

I think the problem is a mismatch between what `flatc` emits and the code we generate. In particular, it seems that:
```hs
rustFBModule :: (?prog_name :: String) => Doc
rustFBModule = "__" <> (pp $ Casing.quietSnake ?prog_name)
```
which is used for generating the `use` statment has a slightly different interpretation of "snake case" compared to `flatc` when it comes to numbers in the name.

ryzhyk:
We used the `Casing.quietSnake` function to generate snake-case Rust
module name when importing the `flatc`-generated module.  This function
does not precisely match the algorithm used by `flatc`, causing "unresolved
import" errors.

Instead of using the library function, we precisely re-implement the
algorithm used by `flatc` in Haskell.

Resolves #425

### DIFF
--- a/test/datalog_tests/dcm1.dl
+++ b/test/datalog_tests/dcm1.dl
@@ -1,0 +1,77 @@
+import intern
+
+typedef int_t = signed<64>
+
+function sum(g: Group<(int_t)>): (int_t) =
+{
+    var sum: int_t = 0;
+    for (resource in g) {
+        sum = resource + sum
+    };
+    sum
+}
+
+input relation NODE(
+	name: IString, 
+	is_master: bool, 
+	unschedulable: bool, 
+	out_of_disk: bool, 
+	memory_pressure: bool, 
+	disk_pressure: bool,
+	pid_pressure: bool, 
+	ready: bool, 
+	network_unavailable: bool, 
+	cpu_capacity: int_t, 
+	memory_capacity: int_t,
+	ephemeral_storage_capacity: int_t, 
+	pods_capacity: int_t, 
+	cpu_allocatable:int_t, 
+	memory_allocatable: int_t, 
+	ephemeral_storage_allocatable:int_t, 
+	pods_allocatable: int_t) 
+
+input relation POD(
+	pod_name: IString, 
+	status: IString, 
+	controllable__node_name: IString,
+	namespace: IString, 
+	cpu_request: int_t,  
+	memory_request: int_t, 
+	ephemeral_storage_request: int_t,
+	pods_request:int_t,
+	owner_name: IString,
+	creation_timestamp: IString,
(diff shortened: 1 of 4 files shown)


d-e-s-o:
Thanks for the fix! I second Mihai's comment. I guess at the least we could rename one of the existing tests to match the pattern (as that's also how I triggered it). Have you checked whether we can remove `casing` from `package.yml` now?

ryzhyk:
> Thanks for the fix! I second Mihai's comment. I guess at the least we could rename one of the existing tests to match the pattern (as that's also how I triggered it). Have you checked whether we can remove `casing` from `package.yml` now?

Added a test.  And removed `casing`.  Thanks for spotting!

test/datalog_tests/dcm1.dl
```diff
@@ -0,0 +1,77 @@
+import intern
```
mbudiu-vmw: no reference file?

ryzhyk: Test data coming in a few days..

PR #460, opened 2019-11-12
Topic/multi receiver

d-e-s-o:
Make TcpReceiver multi-sender capable

So far any `TcpReceiver` was always in a 1-to-1 relationship with a
`TcpSender` (of course, there could also be no sender), mostly because the
initial design was about a TCP "channel".
That turns out to be not very suitable for our current design, where a
`TcpReceiver` and the one port it exposes is supposed to be the endpoint
for all incoming connections (which conceptually can originate from any
number `TcpSender` objects on any number of nodes).
To fix this problem, this patch adjusts the `TcpReceiver` to support
accepting of multiple incoming connections. We use the previously
written transaction multiplexer to serialize transactions from the
individual connections and multiplex them onto a single observer object.

### DIFF
--- a/rust/template/distributed_datalog/src/tcp_channel/receiver.rs
+++ b/rust/template/distributed_datalog/src/tcp_channel/receiver.rs
@@ -24,32 +24,88 @@ use serde::de::DeserializeOwned;
 use crate::observe::Observable;
 use crate::observe::Observer;
 use crate::observe::ObserverBox;
-use crate::observe::OptionalObserver;
 use crate::observe::SharedObserver;
 use crate::tcp_channel::message::Message;
+use crate::txnmux::TxnMux;
 
+/// A struct representing both an `Observer` and an `Observable` that
+/// just passes observable events through to the inner observer.
+#[derive(Debug)]
+struct Passthrough<T, E>(Option<ObserverBox<T, E>>);
+
+impl<T, E> Passthrough<T, E> {
+    fn new() -> Self {
+        Self(None)
+    }
+}
+
+impl<T, E> Observable<T, E> for Passthrough<T, E>
+where
+    T: Debug + Send,
+    E: Debug + Send,
+{
+    type Subscription = ();
+
+    fn subscribe(
+        &mut self,
+        observer: ObserverBox<T, E>,
+    ) -> Result<Self::Subscription, ObserverBox<T, E>> {
+        assert!(
+            !self.0.is_some(),
+            "multiple subscriptions attempted on a Passthrough"
+        );
+        self.0 = Some(observer);
+        Ok(())
+    }
+
+    fn unsubscribe(&mut self, _subscription: &Self::Subscription) -> Option<ObserverBox<T, E>> {
+        self.0.take()
+    }
+}
+
+impl<T, E> Observer<T, E> for Passthrough<T, E>
+where
+    T: Debug + Send,
+    E: Debug + Send,
+{
+    fn on_start(&mut self) -> Result<(), E> {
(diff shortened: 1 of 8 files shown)


d-e-s-o:
https://gitlab.com/ddlog/differential-datalog-d-e-s-o/pipelines/95405073

rust/template/distributed_datalog/src/tcp_channel/receiver.rs
```diff
@@ -222,22 +315,13 @@ where
...
         observer: ObserverBox<T, String>,
     ) -> Result<Self::Subscription, ObserverBox<T, String>> {
-        let mut guard = self.observer.lock().unwrap();
-        if guard.is_some() {
-            Err(observer)
-        } else {
-            guard.replace(observer);
-            Ok(())
-        }
+        self.txnmux.lock().unwrap().subscribe(observer)
```
ryzhyk: Should this be unsubscribe?

rust/template/distributed_datalog/src/tcp_channel/receiver.rs
```diff
@@ -102,44 +162,74 @@ where
...
+                let fd = Arc::new(Mutex::new(Fd::Active(socket.as_raw_fd())));
+                let copy = fd.clone();
+                let thread = spawn(move || Self::process(socket, copy, passthrough));
+                handles.push((thread, fd));
+            }
 
-            Self::process(socket, fd, &mut observer)
+            // We only exit above loop when the receiver is dropped and
+            // in this case we intend to stop and join all the
+            // processing threads we started.
```
ryzhyk: Do we need to tell the processing threads to stop? Otherwise, is there any guarantee that they will terminate sometime soon?

rust/template/distributed_datalog/src/tcp_channel/receiver.rs
d-e-s-o: Not following. Why would it be `unsubscribe`? This is the implementation of the `subscribe` method, after all, and it just forward the request to the `txnmux`.

rust/template/distributed_datalog/src/tcp_channel/receiver.rs
d-e-s-o: We do. There is a cascading shutdown happening. When the user drops the object the "acceptor" thread's `fd` will be shutdown, which will initiate an exit of said thread. That in turn will invoke shutdown on all the individual connection sockets, unblocking the threads along. The "acceptor" thread will then wait to join these threads before exiting. `sender_cleanup` is supposed to test this case.

ryzhyk: > When the user drops the object the "acceptor" thread's fd will be shutdown, which will initiate an exit of said thread

Is it part of the socket API that when the accepting socket is closed, all connections it created shutdown? Can you point me to a doc?

rust/template/distributed_datalog/src/tcp_channel/receiver.rs
ryzhyk: Nevermind.  I got confused by how the diff was displayed. My bad.

rust/template/distributed_datalog/src/tcp_channel/receiver.rs
d-e-s-o: We discussed the matter in person. To summarize: we maintain a vector of `(socket-fd, thread handle)` pairs and when we exit the "acceptor" loop we iterate over this vector shutting down all the sockets and joining the threads. Line 220 is where the loop starts.

PR #462, opened 2019-11-13
Topic/misc

d-e-s-o:
This pull request is a more or less random assortment of changes, all of which will be needed in the future.
As usual, the recommendation is to look at individual commits.

### DIFF
--- a/rust/template/src/lib.rs
+++ b/rust/template/src/lib.rs
@@ -8,8 +8,7 @@
     dead_code,
     overflowing_literals,
     unreachable_patterns,
-    unused_variables,
-    clippy::redundant_closure // Broken in 1.34
+    unused_variables
 )]
 
 use std::borrow;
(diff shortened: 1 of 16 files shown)


d-e-s-o:
CI passed now.

rust/template/src/lib.rs
```diff
@@ -79,11 +79,12 @@ pub fn string_append(mut s1: String, s2: &String) -> String {
...
 
-/// A default implementation of `ConvertRelId` that just forwards calls
+/// A default implementation of `DDlogConvert` that just forwards calls
 /// to generated functions of equal name.
-pub struct RelIdConverter {}
+#[derive(Debug)]
+pub struct DDlogConverter {}
 
-impl ConvertRelId for RelIdConverter {
+impl DDlogConvert for DDlogConverter {
```
ryzhyk: `record.rs` already defines `FromRecord` and `IntoRecord` traits that are similar in spirit in that they convert between program-specific and generic types. I feel that going forward this trait should become part of the same framework.

d-e-s-o: Sure!

PR #463, opened 2019-11-14
Introduce a file adapter

d-e-s-o:
In order to really make use of any distributed computation we need a way
to feed data into it.
This change proposes such infrastructure in the form of an adapter from
a file to an observable. The expected file format is our usual CLI/dump
format. The adapter keeps "watching" the file and will process newly
written data.

### DIFF
--- a/rust/template/cmd_parser/lib.rs
+++ b/rust/template/cmd_parser/lib.rs
@@ -120,7 +120,7 @@ where
     }
 }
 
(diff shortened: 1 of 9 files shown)


d-e-s-o:
Hm, I could swear I was able to stack pull requests in the past. Doesn't seem to work after the base has been changed. So we'll have to wait until https://github.com/vmware/differential-datalog/pull/462 is merged.

d-e-s-o:
Addressed all offline comments now.

PR #464, opened 2019-11-15
Handle sender demise more gracefully in TcpReceiver

d-e-s-o:
Looking at CI logs and test output we can see a bunch error messages of
the form:
> failed to deserialize message: io error: failed to fill whole buffer

The problem here is essentially that a read on a socket whose
counterpart has been closed returns immediately having read 0 bytes. For
the TcpReceiver that means that we fail to decode the message, causing
us to attempt to read more data, which will return 0 bytes again, and so
on.
To fix this problem, this change detects this condition and exits the
processing thread for the closed connection promptly.


ryzhyk:
Looks good!

PR #467, opened 2019-11-19
Topic/rr

d-e-s-o:
Factor out record replay functionality into separate trait

With an upcoming change we would like to introduce a sink that can
produce a dump that is recognized by our CLI. In order to do so it
would be nice to reuse some of the existing logic for persisting
transactions.
Unfortunately, this functionality is not well isolated and fairly spread
out through various functions/methods. In order to unify all of this
logic in one place, this change proposes the introduction of another
trait, RecordReplay, that provides functions for recording various
operations (such as transaction start, transaction commit, etc.).

### DIFF
--- a/rust/template/Cargo.toml
+++ b/rust/template/Cargo.toml
@@ -27,9 +27,9 @@ path = "./ovsdb"
 
 [dependencies]
 abomonation = "0.7"
-cpuprofiler = {version = "0.0.3", optional = true}
+cpuprofiler = {version = "0.0", optional = true}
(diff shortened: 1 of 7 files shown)


ryzhyk:
PS: We should probably take a look into reusing the new trait in the `dump_input_snapshot` method.

d-e-s-o:
> PS: We should probably look into reusing the new trait in the `dump_input_snapshot` method.

Thanks for the review! I looked at it in the past, but something was off or I was unsure about. Let me merge first and look into that next.

PR #469, opened 2019-11-19
Topic/file sink

d-e-s-o:
(depends on https://github.com/vmware/differential-datalog/pull/467)

A recent change introduced a file based source capable of reading our
"replay" format from a file and adapting it to the observer interface.
This change introduces a corresponding sink, which takes updates
received through the observer interface and dumps them into a file in
the form of our "replay" format.

### DIFF
--- a/rust/template/distributed_datalog/src/sinks/file.rs
+++ b/rust/template/distributed_datalog/src/sinks/file.rs
@@ -1,0 +1,151 @@
+use std::fmt::Debug;
+use std::fmt::Display;
+use std::fs::File as FsFile;
+use std::marker::PhantomData;
+
+use differential_datalog::program::Update;
+use differential_datalog::record_val_upds;
+use differential_datalog::DDlogConvert;
+use differential_datalog::RecordReplay;
+
+use crate::Observer;
+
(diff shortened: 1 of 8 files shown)


rust/template/distributed_datalog/src/sinks/file.rs
```diff
@@ -0,0 +1,171 @@
...
+    /// created.
+    pub fn open<P>(path: P) -> Result<Self, Error>
+    where
+        P: AsRef<Path>,
+    {
+        Ok(Self {
+            file: OpenOptions::new()
+                .create(true)
+                .write(true)
+                .append(true)
```
ryzhyk: It might be good to have a flag in the `new()` method to choose whether to set `append` to `true`.

d-e-s-o: Let me remove the method for now, so we don't have any such policy stuff in there to begin with.

d-e-s-o:
Merging after Rust tests are done (https://gitlab.com/ddlog/differential-datalog-d-e-s-o/pipelines/97380181). Nothing else changed.

PR #474, opened 2019-11-20
Topic/trait usage

d-e-s-o:
Let's try this again...(GitHub really is throwing branches between one's legs whenever it can :-|)

-----

As discussed in person, this is a proposal for how we could decouple "generic" parts of the code from the generated code some more, while also cleaning up a bit.

In a nutshell, the proposal is to use Rust's `TryFrom` trait and implement it for `Relations` to provide certain functionality, as opposed to having explicit conversion functions.

Here is a list of functions we ultimately may want to convert. Checked off are the ones covered by this proposal:

- [x] `relname2id` => `TryFrom<&str> for Relations`
- [x] `output_relname_to_id` => `Relations::is_output`
- [x] `input_relname_to_id` => `Relations::is_input`
- [x] `relid2rel` => `TryFrom<RelId> for Relations`
- [ ] `relval_from_record` => ???
- [ ] `relkey_from_record` => ???
- [ ] `relid2name` => ???
- [ ] `relid2cname` => ???

### DIFF
--- a/rust/template/src/api.rs
+++ b/rust/template/src/api.rs
@@ -412,8 +412,8 @@ impl HDDlog {
 pub fn updcmd2upd(c: &record::UpdCmd) -> Result<Update<Value>, String> {
     match c {
         record::UpdCmd::Insert(rident, rec) => {
-            let relid: Relations =
-                relident2id(rident).ok_or_else(|| format!("Unknown relation {}", rident))?;
+            let relid =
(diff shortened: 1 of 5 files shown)


d-e-s-o:
Changes are ready for review now :-) (I added change descriptions and changed the function names in `Compile.hs`, nothing else changed)

PR #478, opened 2019-11-25
Add some tracing information to our observables & observers

d-e-s-o:
Getting an overview of what is happening in a particular setup can be
challenging, even for small configurations. In an attempt to provide a
first version of (arguably limited) introspection capabilities, this
change proposes the addition of some tracing information. We use the log
crate's trace macro (which just logs at a different log level as opposed
to, say, error, which we already use) to print chosen details of a
certain method being invoked. We also "tag" each observer/observable
with a dedicated ID, allowing for some limited correlation of events.
Such IDs are unique in any single process. In a truly distributed setup
IDs will be repeated (they just start at 1), but because output is
emitted on a different node uniqueness is still maintained (but may
require some post-processing).

### DIFF
--- a/rust/template/distributed_datalog/Cargo.toml
+++ b/rust/template/distributed_datalog/Cargo.toml
@@ -22,6 +22,7 @@ libc = "0.2"
 log = "0.4"
 nom = "4.0"
 serde = {version = "1.0", features = ["derive"]}
+uid = "0.1"
 uuid = {version = "0.8", default-features = false, features = ["serde", "v4"]}
 waitfor = "0.1"
 
--- a/rust/template/distributed_datalog/src/server.rs
+++ b/rust/template/distributed_datalog/src/server.rs
@@ -2,6 +2,9 @@ use std::collections::{HashMap, HashSet};
 use std::fmt::Debug;
 use std::sync::Arc;
 
+use log::trace;
+use uid::Id;
+
 use differential_datalog::program::RelId;
 use differential_datalog::program::Update;
 use differential_datalog::program::Val;
@@ -30,6 +33,7 @@ pub struct DDlogServer<P>
 where
     P: DDlog,
 {
+    id: usize,
     prog: Option<P>,
     outlets: Vec<Outlet<P::Value>>,
     redirect: HashMap<RelId, RelId>,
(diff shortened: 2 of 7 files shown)


d-e-s-o:
Thanks for taking a look!

> Why not use descriptive strings instead of numeric id's?

Where would we get these strings from? I'd not want to pollute every call site with some "name" of the component being instantiated.

> Does it make sense to have per-component trace macros (server_trace, etc.?)

Can you elaborate a bit please? What would such a macro do that would justify its existence?

ryzhyk:
> Thanks for taking a look!
> 
> > Why not use descriptive strings instead of numeric id's?
> 
> Where would we get these strings from? I'd not want to pollute every call site with some "name" of the component being instantiated.

If anything, it would improve readability. Besides, you could still keep existing constructors that do not take a name, in which case the name could be generated automatically, e.g., from the name of the file in case of file source/sink. 
 
> > Does it make sense to have per-component trace macros (server_trace, etc.?)
> 
> Can you elaborate a bit please? What would such a macro do that would justify its existence?

You'd be able to write

```
trace_server!("add_stream");
```

instead of

```
trace!("DDlogServer({})::add_stream", self.id);
```

d-e-s-o:
> You'd be able to write
> 
> ```
> trace_server!("add_stream");
> ```
> 
> instead of
> 
> ```
> trace!("DDlogServer({})::add_stream", self.id);
> ```

Hm, I don't think that's possible. Rust macros are [hygienic](https://en.wikipedia.org/wiki/Hygienic_macro) and cannot reference `self` implicitly, I believe; you'd have to pass it in. So we'd end up with `trace_server!("add_stream", self.id);` and I am not sure how much that buys us, because now we need add an "overload" whenever we want to add additional information and each file needs to define its own version (again, with overloads) just to avoid having to mention the name.

d-e-s-o:
> > Thanks for taking a look!
> > > Why not use descriptive strings instead of numeric id's?
> > 
> > 
> > Where would we get these strings from? I'd not want to pollute every call site with some "name" of the component being instantiated.
> 
> If anything, it would improve readability. Besides, you could still keep existing constructors that do not take a name, in which case the name could be generated automatically, e.g., from the name of the file in case of file source/sink.

I agree it would be useful to have, but it also requires a major rework of all constructors and I am having a hard time getting this "right".
I am the first one to improve the functionality should there be a real need, but right now I'd like to not spent more time on this minor detail as I am having trouble seeing the proposal as an actual improvement (compared to merely over engineering). If you think that this infrastructure is insufficient then let's not merge it now and we can rework it should the need arise.

ryzhyk:
As far a I'm concerned, this is ready to be merged when you think is ready.

PR #479, opened 2019-11-25
Eliminate unnecessary start & commit events

d-e-s-o:
With the end-to-end test we noticed that an unnecessary start-commit
pair of events is emitted into the replay file. The reason is to be
found in the `DDlogServer`, which invokes `on_start` and `on_commit` on
observers even if no data is actually emitted. We probably don't need
such useless chatter, so rework this logic to only emit those events if
actual deltas are calculated.


d-e-s-o:
https://gitlab.com/ddlog/differential-datalog-d-e-s-o/pipelines/98452174

PR #480, opened 2019-11-26
Simplify RelCfg::Input variant

d-e-s-o:
The `RelCfg::Input` variant currently requires a set of input relations.
That is actually unnecessary, because a `NodeCfg` already contains a set
of `RelCfg` objects.
So let's simplify the logic by removing the inner set, flattening the resulting structure somewhat.


d-e-s-o:
GitLab CI run to be found @ https://gitlab.com/ddlog/differential-datalog-d-e-s-o/pipelines/98507535

PR #481, opened 2019-11-26
Require explicit specification of output relations

d-e-s-o:
So far we have followed an approach where we did not explicitly specify
output relations in the system and node configurations. Instead, those
were deduced by looking at the input relations of other nodes'
relations.
While reasonably nice and concise, there are a few problems with this
approach:
- it seems inconsistent that we specify sources, inputs, and sinks (also
  a form of output), but not outputs themselves
- calculating output relations requires knowledge of the complete system
  configuration
- the output deduction was a rather complex piece of code

This change introduces a new variant, `Output`, to the `RelCfg` enum that
indicates that a relation is an output relation and should feed into a given
node.
The downside with this approach is that we now can have an invalid
configuration, where inputs are specified but no corresponding outputs,
for example. The believe is that we can and should generate the
configuration in one way or another, which would alleviate this problem
in many cases.

### DIFF
--- a/rust/template/distributed_datalog/src/assign.rs
+++ b/rust/template/distributed_datalog/src/assign.rs
@@ -87,7 +87,11 @@ mod tests {
         let node1 = Addr::Ip("127.0.0.1:2".parse().unwrap());
 
         let config = btreemap! {
-            uuids[0] => btreemap! {},
+            uuids[0] => btreemap! {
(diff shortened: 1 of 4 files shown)


d-e-s-o:
https://gitlab.com/ddlog/differential-datalog-d-e-s-o/pipelines/98732935

PR #482, opened 2019-11-27
Topic/buffered tcp sender

d-e-s-o:
Make `TcpSender` buffer transactions until it connected

The code for instantiating a system or node configuration attempts to
create a `TcpSender` to feed output relations to a different node.
Currently the `TcpSender` creation is attempting to synchronously connect
to the desired address, which can be a long blocking operation. But even
if it returns quickly (e.g., in cases where the connection is refused
immediately), we retry for a certain duration -- again blocking the
caller.
Such blocking is undesired in any of these code paths, as it can easily
cause deadlocks (or at least a state where nobody can make progress
until a timeout is reached).

### DIFF
--- a/rust/template/distributed_datalog/src/tcp_channel/socket.rs
+++ b/rust/template/distributed_datalog/src/tcp_channel/socket.rs
@@ -1,3 +1,13 @@
+//! A module providing an asynchronously cancelable socket. Rust's
+//! `std::net::TcpStream` does not allow for interrupting a connection
+//! attempt midway, as there is simply no API for doing so. As a clumsy
+//! work around one could use `std::net::TcpStream::connect_timeout`
+//! with a short timeout, after which one could check for a cancellation
+//! request, but that requires an unnecessary amount of cycles.
+//!
+//! This module provides a way to cancel a connection request issued
+//! earlier by separating socket creation from connection.
+
 use std::convert::TryFrom;
 use std::convert::TryInto;
 use std::io::Error;
@@ -120,6 +130,11 @@ fn connect(socket: &Fd, addr: &SocketAddr) -> Result<(), Error> {
                     // readiness. If the connection got refused we start
                     // over attempting to connect again.
                     if pollfds[0].revents & libc::POLLHUP != 0 {
+                        // TODO: We could probably do better than busy
+                        //       waiting, but it is not clear how.
+                        //       Perhaps we just sleep or we have some
+                        //       form of exponential backoff on repeated
+                        //       connection failures?
                         continue 'connect;
                     }
                     return Ok(());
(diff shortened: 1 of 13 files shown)


d-e-s-o:
While the changes here can probably be looked at already, the second commit for some reason causes intermittent failures of the `server_api` test. I haven't gotten to the bottom of that yet.

d-e-s-o:
Should be ready for review now.

d-e-s-o:
I think we have it now and GitLab even agrees with me (!).

https://gitlab.com/ddlog/differential-datalog-d-e-s-o/pipelines/100687007

rust/template/distributed_datalog/src/tcp_channel/txnbuf.rs
```diff
@@ -0,0 +1,245 @@
...
+impl<W, T> TxnBuf<W, T>
+where
+    W: Debug + Send + Write,
+    T: Debug + Send + Serialize,
+{
+    /// Convert the `TxnBuf` into the `Writer` variant.
+    ///
+    /// An error return indicates a failure to flush all buffered
+    /// transactions. The objects is in an undefined state afterwards.
+    pub fn make_writer(&mut self, mut writer: W) -> Result<(), String> {
```
ryzhyk: This function name refers to internal structure of `TxnBuf`, not what it does from the user's perspective. Can we give it a more intuitive name, e.g., `set_mode_passthrough()`.

rust/template/distributed_datalog/src/tcp_channel/txnbuf.rs
```diff
@@ -0,0 +1,245 @@
+use std::collections::LinkedList;
```
ryzhyk: Please add a module-level comment.

rust/template/distributed_datalog/src/tcp_channel/socket.rs
```diff
@@ -0,0 +1,461 @@
+use std::convert::TryFrom;
```
ryzhyk: Please add a module-level comment.

rust/template/distributed_datalog/src/tcp_channel/txnbuf.rs
d-e-s-o: Thanks for your review! Well, it's an `enum` so the there really isn't anything internal about the structure. But the name you proposed is fine with me as well.

ryzhyk: As a user, it does not matter to me that it's an enum. It's just a filter that can operate in either buffered or passthrough mode. In reviewing the code, I came across a call to this function before reading the function itself, and I couldn't figure our what it was doing, hence my suggestion to rename.

d-e-s-o:
Gitlab succeeded here: https://gitlab.com/ddlog/differential-datalog-d-e-s-o/pipelines/101040599, merging.

PR #483, opened 2019-12-04
Convert a subset of SQL into DDlog

mbudiu-vmw:


sql/src/test/java/ddlog/RuntimeTest.java
```diff
@@ -0,0 +1,1064 @@
...
+    }
+
+    @Test
+    public void testJoinStarWNull() {
+        String query = "create view v1 as SELECT DISTINCT * FROM t1 JOIN t2 ON t1.column1 = t2.column1";
+        String program = this.header(true) +
+                "typedef Ttmp0 = Ttmp0{column1:Option<signed<64>>, column2:Option<string>, column3:Option<bool>, column10:Option<signed<64>>}\n" +
+                this.relations(true) +
+                "output relation Rv1[Ttmp0]\n" +
+                "Rv1[v3] :- Rt1[v0],Rt2[v1],unwrapBool(a_eq_NN(v0.column1, v1.column1)),true,var v2 = Ttmp0{.column1 = v0.column1,.column2 = v0.column2,.column3 = v0.column3,.column10 = v1.column1},var v3 = v2.";
```
lalithsuresh: I assume the ",true," and "val v2 = v3" clauses get optimized away by ddlog/rust?

sql/src/test/java/ddlog/RuntimeTest.java
```diff
@@ -0,0 +1,1064 @@
...
+                "64'sd1} else {\n" +
+                "if unwrapBool(a_lt_RN(64'sd1, v0.column1)){\n" +
+                "64'sd2} else {\n" +
+                "64'sd3}}},var v3 = v2.";
+        testTranslation(query, program, true);
+    }
+
+    @Test
+    public void testAlias() {
+        String query = "create view v1 as SELECT DISTINCT t2.column1 FROM t1 AS t2";
```
lalithsuresh: It looks the alias is ignored in the ddlog program because it is not required in this instance. That wouldn't be the case with things like select expressions right (they'd affect the names in the output relation)?

sql/src/test/java/ddlog/RuntimeTest.java
```diff
@@ -0,0 +1,1064 @@
...
+            "typedef Ttmp0 = Ttmp0{column1:Option<signed<64>>}\n" +
+            this.relations(true) +
+            "relation Rtmp0[Ttmp0]\n" +
+            "output relation Rv1[Ttmp0]\n" +
+            "Rv1[v2] :- Rt1[v0],var v1 = Ttmp0{.column1 = v0.column1},var v2 = v1.";
+        testTranslation(query, program, true);
+    }
+
+    @Test
+    public void testScope() {
```
lalithsuresh: This is the same output program as testAlias(). What's being tested here?

sql/src/test/java/ddlog/RuntimeTest.java
lalithsuresh: Ah I see them being tested elsewhere in testMultiJoin().

sql/src/test/java/ddlog/RuntimeTest.java
mbudiu-vmw: We should support as much as possible from the SQL syntax.

sql/src/test/java/ddlog/RuntimeTest.java
mbudiu-vmw: If they aren't they should. This is a legal program, even if it is not optimal.

mbudiu-vmw:
@ryzhyk : should we merge this PR before I add more code to it?
I thought I had rebased, I will investigate what's going on.

ryzhyk:
> @ryzhyk : should we merge this PR before I add more code to it?
> I thought I had rebased, I will investigate what's going on.

I am ok with merging, assuming it's cleanly rebased on top of master (right now it contains a bunch of @d-e-s-o 's commits that are already in master on top of some of the SQL patches).

mbudiu-vmw:
I tried to rebase, but it turns into a nightmare, which I don't really understand. I get lots of conflicts between my own edits.

mbudiu-vmw:
However, squash and merge would just work.

ryzhyk:
> However, squash and merge would just work.

Maybe, but please let's not do anything hasty like that.

mbudiu-vmw:
In the other projects I work on squash-and-merge is the preferred method and it works quite well. You will get exactly one commit with the actual changes between this branch and the master branch. What other alternative are you proposing?

lalithsuresh:
I managed to do a "rebase -i master" today morning just fine. Do you want me to create another branch on my repo that you could work off instead?

ryzhyk:
> In the other projects I work on squash-and-merge is the preferred method and it works quite well. You will get exactly one commit with the actual changes between this branch and the master branch. What other alternative are you proposing?

Squash and merge is ok with me (assuming you don't need to preserve your commit history), but it should only contain your changes, not a mix of commits from this branch and master. Perhaps the right thing to do is to remove master commits and cleanly rebase this branch on top of master first.

mbudiu-vmw:
I think I can just create a new branch and move everything in the sql directory on top of it. Let me try that. I don't really understand why rebase is so confused.

PR #485, opened 2019-12-07
Topic/server api cli w zookeeper

d-e-s-o:
The changes in this pull request do two things. First, they introduce a CLI application for the `server_api` program that can be used to manually instantiate a configuration. Second, they also introduce a first version of `ZooKeeper` support and hook that up with the CLI app.

The end result of all that is that we can instantiate a configuration based on data read from a `ZooKeeper` cluster, and when the data is changed we reconfigure the distributed computation.

Example:
Shell 1:
```bash
$ rm /tmp/output.dump
$ tail -F /tmp/output.dump
> tail: cannot open '/tmp/output.dump' for reading: No such file or directory
```

Shell 2 (one a ZooKeeper node):
```bash
$ zkCli.sh -server 127.0.0.1:2181
[zk: 127.0.0.1:2181(CONNECTED) 0] rmr /d3log
```

Shell 3:
```bash
$ RUST_LOG=debug,distributed_datalog=trace <...>/server_api_test zookeeper \
    --nodes=10.20.92.42:2181 --nodes=10.20.95.186:2181 --nodes=10.20.92.175:2181 \
    --member=127.0.0.1:5007
```

Shell 4:
```bash
$ RUST_LOG=debug,distributed_datalog=trace <...>/server_api_test zookeeper \
    --nodes=10.20.92.42:2181 --nodes=10.20.95.186:2181 --nodes=10.20.92.175:2181 \
    --member=127.0.0.1:5008
```

Shell 5:
```bash
$ RUST_LOG=debug,distributed_datalog=trace <...>/server_api_test zookeeper \
    --nodes=10.20.92.42:2181 --nodes=10.20.95.186:2181 --nodes=10.20.92.175:2181 \
    --member=127.0.0.1:5009
```

At this point all "nodes" are in a waiting state, because there is no
configuration to instantiate yet. So let's create that.

Shell 2 (continued):
```bash
[zk: 127.0.0.1:2181(CONNECTED) 2] create /d3log/members [...]
> Created /d3log/members
create /d3log/config {"7aba4dd8-80fc-43f0-bbf6-0dc56abaa001":...}
> Created /d3log/config
```

Shell 1 (continued):
```bash
> tail: '/tmp/output.dump' has appeared;  following new file
> start;
> insert server_api_3.P3Out["server_api_1:1server_api_2:1"];
> commit;
^C
```


rust/template/distributed_datalog/src/read.rs
```diff
@@ -0,0 +1,21 @@
...
+pub trait ReadMembers {
+    /// Retrieve the current set of members.
+    fn members(&self) -> Result<Members, String>;
+}
+
+/// A client to some form of membership service.
+///
+/// Right now only read-only access is supported.
+pub trait ReadConfig {
+    /// Retrieve the current set of members.
```
ryzhyk: copy-paste error; should be "Retrieve the current configuration."?

rust/template/distributed_datalog/src/read.rs
```diff
@@ -0,0 +1,21 @@
+//! A module containing traits that abstract away the retrieval of
```
ryzhyk: `read_config.rs` might be a better name for this file.

rust/template/distributed_datalog/src/zookeeper.rs
```diff
@@ -0,0 +1,142 @@
...
+    I: Iterator<Item = S>,
+    S: AsRef<str>,
+    W: Watcher + 'static,
+{
+    let first = servers
+        .next()
+        .ok_or_else(|| "failed to connect to ZooKeeper: no servers supplied")?
+        .as_ref()
+        .to_string();
+    let servers = servers.fold(first, |x, y| x + "," + y.as_ref());
```
ryzhyk: Would it be simple to use `join()` instead?
https://doc.rust-lang.org/beta/std/primitive.slice.html#examples-71

.gitlab-ci.yml
```diff
@@ -12,10 +21,18 @@ variables:
 
 # Test Rust template only.
 test-rust:
+    variables:
+      ZOOKEEPER_ENDPOINTS: "127.0.0.1:2181"
     script:
+    - apt-get update
+    - apt-get install --assume-yes zookeeper
```
ryzhyk: I just pushed a new docker image with zookeeper pre-installed, so it should be ok to remove these commands now

rust/template/distributed_datalog/src/read.rs
d-e-s-o: Yeah, I was struggling hard to name this file :-0

.gitlab-ci.yml
d-e-s-o: Awesome! Thank you!

rust/template/distributed_datalog/src/zookeeper.rs
d-e-s-o: The reason I decided against it is because `join` seemingly requires a slice, which we don't have. So we'd need to `collect` into a `Vec` first to make this work, which seems wasteful.

PR #488, opened 2019-12-12
Dynamic loading of ddlog program

mbudiu-vmw:
@lalithsuresh : this adds an API for dynamically compiling a ddlog program and loading it at runtime.
At this point only one program can be loaded, and there's no "unload" operation.


java/ddlogapi/DDlogAPI.java
```diff
@@ -129,6 +126,7 @@ private void checkHandle() throws DDlogException {
      */
     public DDlogAPI(int workers, Consumer<DDlogCommand<DDlogRecord>> callback, boolean storeData)
             throws DDlogException {
+        System.loadLibrary("ddlogapi");
```
ryzhyk: Will this work if DDlogAPI is instantiated multiple times with the same version of the library?

java/ddlogapi/DDlogAPI.java
```diff
@@ -412,4 +410,107 @@ static public synchronized void logSetCallback(int module, ObjIntConsumer<String
...
+
+    /**
+     * Compile and load a ddlog program stored in a file.
+     * @param ddlogFile  Pathname to the ddlog program.
+     * @param ddlogInstallationPath  Path to DDLog installation.
+     * @param ddlogLibraryPath  Additional list of paths for needed ddlog libraries.
+     * @return           A DDlogAPI which can be used to access the program.
+     *                   On error returns null and prints an error on stderr.
+     */
+    public static DDlogAPI compileAndLoad(
```
ryzhyk: Is there a reason we cannot script this? This probably won't work if one wants to set custom flags for one of the commands executed below.

java/ddlogapi/DDlogAPI.java
mbudiu-vmw: I don't really know. Do we have such a test case yet?

java/ddlogapi/DDlogAPI.java
mbudiu-vmw: What do you mean "script"? You mean "generate a script and run it?"

java/ddlogapi/DDlogAPI.java
ryzhyk: we do

java/ddlogapi/DDlogAPI.java
ryzhyk: I don't think it needs to be generated, it can be written manually and customized for a particular application such as Weave. Point is, it is probably best if each step in this workflow is handled by a separate program:

1. convert SQL to DDlog
2. compile DDlog code into a dynamic library
3. load and use the library from a Java program.

java/ddlogapi/DDlogAPI.java
ryzhyk: and I thought you also had a test like that at some point

mbudiu-vmw:
Same test failed on gitlab. I don't know why.

ryzhyk:
The test is failing because `SpanTest.java` invokes some static methods of `DDlogAPI` before creating a `DDlogAPI` instance.

mbudiu-vmw:
Do you happen to know what the static method is the one that is called?
I see only one obvious candidate: `logSetCallback`

mbudiu-vmw:
Also, would you like for stderr not be redirected at all?

ryzhyk:
>  I see only one obvious candidate: logSetCallback

Yep, it's `logSetCallback`. Here's the exception:

```
java -Djava.library.path=. -cp ../ddlogapi.jar:. SpanTest ../../test/datalog_tests/span_uuid.dat
Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8
Exception in thread "main" java.lang.UnsatisfiedLinkError: 'long ddlogapi.DDlogAPI.ddlog_log_replace_callback(int, long, java.util.function.ObjIntConsumer, int)'
	at ddlogapi.DDlogAPI.ddlog_log_replace_callback(Native Method)
	at ddlogapi.DDlogAPI.logSetCallback(DDlogAPI.java:412)
	at SpanTest$SpanParser.<init>(SpanTest.java:127)
	at SpanTest.main(SpanTest.java:427)
```

ryzhyk:
> Also, would you like for stderr not be redirected at all?

Do you remember why we're redirecting it in the first place.  Does this program print a lot of data to stderr? Ideally, we should not need to redirect it, as this makes it hard to debug CI failures.

mbudiu-vmw:
I will try to run it without to see what happens

mbudiu-vmw:
Yes, it has a lot of logging.
IT was testing the log integration.

mbudiu-vmw:
I will try to log directly to a file and leave stderr alone.

ryzhyk:
> I will try to log directly to a file and leave stderr alone.

Sounds good, thanks!

PR #501, opened 2020-01-13
Improve re-compilation time using dynamic dispatch.

ryzhyk:
Rationale:

Differential dataflow allows the user to assign an arbitrary user-defined type to
each collection.  It relies on Rust's static dispatch mechanism to specialize its internal
machinery for each user-defined type.  Unfortunately, beyond very simple programs this leads to
extremely long compilation times.  One workaround that we used to rely on is to declare a
single enum type with a variant per concrete type used in at least one relation.  This make
compilation feasible, but still very slow (~6 minutes for a simple DDlog program and ~10
minutes for complex programs).

Design:

Another alternative we implement here is to use a fixed value type that does not depend on
a concrete DDlog program and rely on dynamic dispatch to forward operations that DD expects
all values to implement (comparison, hashing, etc.) to their concrete implementations.  This
way the differetial-datalog crate can be compiled all the way to binary code separately
from the DDlog program using it and does not need to be re-compiled when the DDlog program
changes.  Thus, the only part that must be re-compiled on changes to the DDlog code is the
auto-generated crate that declares concrete value types and rules.  This is much faster than
re-compiling both crates together.

The next design decision is how to implement dynamic dispatch.  Rust trait objects is an
obvious choice, with value type being declared as `Box<dyn SomeTrait>`.  However, this proved
suboptimal in our experiments, as this design requires a dynamic memory allocation per value,
no matter how small.  Furthermore, cloning a value (which DD does a lot, e.g., during
compaction) requires another allocation.

We improve over this naive design in two ways.  First, we use `Arc` instead of `Box`, which
introduces extra space overhead to store the reference count, but avoids memory allocation due
to cloning and shares the same heap allocation across multiple copies of the value.  Second, we
store small objects <=`usize` bytes inline instead of wrapping them in an Arc to avoid dynamic
memory allocation for such objects altogether.  Unfortunately, Rust's dynamic dispatch
mechanism does not support this, so we roll our own instead, with the following `DDValue`
declaration:

```
pub struct DDValue {
   val: DDVal,
   vtable: &'static DDValMethods,
}
```

where `DDVal` is a `usize` that stores either an `Arc<T>` or `T` (where `T` is the actual type
of value stored in the DDlog relation), and `DDValMethods` is a virtual table of methods that
must be implemented for all DD values.

Results:

This design still requires a separate heap allocation for each value >8 bytes, which slows
things down quite a bit.  Nevertheless, it has the same performance as our earlier
implementation using static dispatch and at least in some benchmarks uses less memory.
(see `doc/benchmarking/dynamic-dispatch.gnumeric` for more detail)

While initial compilation time is approximately the same as before, re-compilation after changing
DDlog code is much faster, e.g., approximately one minute for `redist_opt.dl` and 5 minutes for `ovn-northd`.

Future work:

There's more to do here, most importantly:
- Move `differential-datalog` crate out of the template and into a shared location, e.g., $DDLOG_HOME, so we only need to compile it once and not for each DDlog program.  This way we should be able to run more tests.
- See if we can speed-up recompilation of large programs even more, e.g., by placing each value type in a separate module that will only be recompiled when that type changes.
- Smarter memory allocation strategy that would allocate multiple values at once using domain-specific information from DD.


rust/template/differential_datalog/program.rs
```diff
@@ -2525,7 +2479,7 @@ impl<V: Val> RunningProgram<V> {
     }
 
     /* Broadcast message to all worker threads. */
-    fn broadcast(&self, msg: Msg<V>) -> Response<()> {
+    fn broadcast(&self, msg: Msg) -> Response<()> {
```
mbudiu-vmw: This is actually very nice - it's almost search-and-replace.

rust/template/differential_datalog/test.rs
```diff
@@ -639,20 +376,16 @@ fn test_semijoin(nthreads: usize) {
             })))),
         }
     };
-    fn fmfun1(v: Value) -> Option<Value> {
-        match v {
-            Value::Tuple2(v1, _v2) => Some(*v1),
-            _ => None,
-        }
+    fn fmfun1(v: DDValue) -> Option<DDValue> {
+        let Tuple2(ref v1, ref _v2) = unsafe { Tuple2::<U64>::from_ddvalue(v) };
```
mbudiu-vmw: This does not seem to be the same thing as the original code - this can never return None.
Was the original code unable to statically typecheck? 
Where is the dynamic type check made?

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -794,51 +765,51 @@ mkValueFromRecord d@DatalogProgram{..} =
...
     "    match rel {"                                                                               $$
-    (nest' $ nest' $ vcat $ punctuate comma entries)                                                $$
+    (nest' $ nest' $ vcommaSep entries)                                                             $$
     "    }"                                                                                         $$
     "}"                                                                                             $$
-    "pub fn relkey_from_record(rel: Relations, _rec: &record::Record) -> Result<Value, String> {"   $$
+    "pub fn relkey_from_record(rel: Relations, _rec: &record::Record) -> result::Result<DDValue, String> {" $$
     "    match rel {"                                                                               $$
-    (nest' $ nest' $ vcat $ key_entries)                                                            $$
+    (nest' $ nest' $ vcommaSep key_entries)                                                         $$
```
mbudiu-vmw: was this a bug? these do not seem to be equivalent.

rust/template/differential_datalog/ddval.rs
```diff
@@ -0,0 +1,423 @@
...
+    fn into_ddvalue(self) -> DDValue;
+}
+
+#[typetag::serde]
+pub trait DDValSerialize {
+    fn ddvalue(&self) -> DDValue;
+}
+
+/// Macro to implement `DDValConvert` for type `t` that satisfies the following type bounds:
+///
```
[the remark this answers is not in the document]
ryzhyk: Hopefully :)

rust/template/differential_datalog/ddval.rs
```diff
@@ -0,0 +1,423 @@
...
+                    };
+                    $crate::ddval::DDVal { v }
+                } else {
+                    $crate::ddval::DDVal {
+                        v: std::sync::Arc::into_raw(std::sync::Arc::new(self)) as usize,
+                    }
+                }
+            }
+
+            fn into_ddvalue(self) -> $crate::ddval::DDValue {
```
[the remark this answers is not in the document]
ryzhyk: Yes, the const declaration should take care of this.

rust/template/differential_datalog/ddval.rs
```diff
@@ -0,0 +1,423 @@
...
+                        };
+                        __f
+                    },
+                    eq: {
+                        fn __f(this: &$crate::ddval::DDVal, other: &$crate::ddval::DDVal) -> bool {
+                            unsafe { <$t>::from_ddval_ref(this).eq(<$t>::from_ddval_ref(other)) }
+                        };
+                        __f
+                    },
+                    partial_cmp: {
```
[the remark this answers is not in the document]
ryzhyk: No. Values of different types should never be compared. We enforce this at runtime.

rust/template/differential_datalog/ddval.rs
```diff
@@ -0,0 +1,423 @@
...
+//! minutes for complex programs).
+//!
+//! Another alternative we implement here is to use a fixed value type that does not depend on
+//! a concrete DDlog program and rely on dynamic dispatch to forward operations that DD expects
+//! all values to implement (comparison, hashing, etc.) to their concrete implementations.  This
+//! way this crate (differetial-datalog) can be compiled all the way to binary code separately
+//! from the DDlog program using it and does not need to be re-compiled when the DDlog program
+//! changes.  Thus, the only part that must be re-compiled on changes to the DDlog code is the
+//! auto-generated crate that declares concrete value types and rules.  This is much faster than
+//! re-compiling both crates together.
```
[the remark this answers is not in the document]
ryzhyk: Not until we move the shared crate out of the template and in a separate location (even then I wonder what it will take to convince Rust not to re-compile it for each test).

src/Language/DifferentialDatalog/Compile.hs
ryzhyk: I think the comma used to be in `key_entries`.

rust/template/differential_datalog/test.rs
ryzhyk: The original code used an `enum Value` and had to always match on it even though we knew exactly what variant we handle in every function. Now we use unsafe conversion instead.

PR #502, opened 2020-01-13
Implementation of raw SQL aggregates (without groupby)

mbudiu-vmw:
Note that the CodeFormatter class is not currently used. This is WIP.
I have split the RuntimeTests file with junit tests into multiple files.
@lalithsuresh : if this works then the remaining important elements are group-by and having.


ryzhyk:
I think we should have at least minimal CI tests for this. E.g., we could use @lalithsuresh's DCM queries. This way we'll be able to catch DDlog changes that break the SQL client.

mbudiu-vmw:
It's on the TODO list. 
We should start by having the SQL client run any of the generated queries on some data.
I am currently only running the ddlog compiler, but not the result it produces.

ryzhyk:
> I am currently only running the ddlog compiler, but not the result it produces.

Then we should start with having tests for just this part in CI. For example here is a CI job that runs a souffle test:

```
test-souffle0:
    extends: .install-ddlog
    tags:
        - ddlog-ci-1
    script:
        - (cd test/souffle0 &&
          ../../tools/souffle_converter.py test.dl souffle --convert-dnf &&
          ../datalog_tests/run-test.sh souffle.dl release)
```

You probably want something similar, just using your SQL compiler instead of `souffle_converter.py`.

Anyhow, if you have a list of commands you can run to test your pipeline it is trivial to covert it to a CI job and is definitely worth doing.

sql/src/main/java/com/vmware/ddlog/translator/AggregateVisitor.java
```diff
@@ -0,0 +1,151 @@
...
+public class AggregateVisitor
+        extends AstVisitor<Ternary, TranslationContext> {
+    /**
+     * This class represents the decomposition of an expression that contains
+     * aggregates into multiple expressions.  Consider this example:
+     *
+     * select max(salaries) + min(abs(salaries)) from employees
+     *
+     * Consider the expression tree
+     * + ( functionCall(max, salaries), functionCall(min, functionCall(min, salaries)))
```
lalithsuresh: functionCall(min, functionCall(min, salaries))) -> functionCall(min, functionCall(abs, salaries)))

sql/src/main/java/com/vmware/ddlog/translator/SqlSemantics.java
```diff
@@ -0,0 +1,175 @@
...
+        this.stringFunctions.put("s_neq", DDlogEBinOp.BOp.Neq);
+
+        this.booleanFunctions.put("b_eq", DDlogEBinOp.BOp.Eq);
+        this.booleanFunctions.put("b_neq", DDlogEBinOp.BOp.Neq);
+        this.booleanFunctions.put("b_and", DDlogEBinOp.BOp.And);
+        this.booleanFunctions.put("b_or", DDlogEBinOp.BOp.Or);
+        this.booleanFunctions.put("b_impl", DDlogEBinOp.BOp.Impl);
+    }
+
+    public static SqlSemantics semantics = new SqlSemantics();
```
lalithsuresh: Why does this need an instance? As opposed to calls like isAggregateFunction() etc being static methods?

sql/src/main/java/com/vmware/ddlog/translator/Translator.java
```diff
@@ -83,10 +83,9 @@ public DDlogIRNode translateExpression(final String sql) {
     }
 
     public DDlogProgram generateLibrary() {
-        return this.translationContext.generateLibrary();
+        return SqlSemantics.semantics.generateLibrary();
```
lalithsuresh: Why is generateLibrary in SqlSemantics?

ryzhyk:
@mbudiu-vmw, the build is failing in CI, probably because `ddlogapi.jar` does not get compiled and added to `$CLASSPATH`. Is this something you need to add to the maven config or should I call `cd java; make` + set `$CLASSPATH`  directly from the CI script?

PS. I checked that the docker container used in CI already has all dependencies from `install-dependencies.sh` installed.

mbudiu-vmw:
We need to do two more things:
`cd java; make` to build the ddlogapi
`cd sql; ./install-ddlog-jar.sh` to install the ddlogapi in a place where mvn can find it

ryzhyk:
GitLab tests are now passing. Any idea what's happening in Travis?

mbudiu-vmw:
My guess is that we are pulling the latest version of parglare and that's broken. Perhaps we should pull a specific version of it.

mbudiu-vmw:
Yes, the problem is parglare. I could reproduce it on my machine.
The solution is to do this in .travis.yml: `pip install parglare==0.10.0`

ryzhyk:
All tests finally passes.  Merging.

PR #503, opened 2020-01-24
CI and build_java fixes (see commit comments).

ryzhyk:


java/build_java.sh
```diff
@@ -57,7 +57,7 @@ function compile {
     # Compile Java program
     javac -encoding utf8 -Xlint:unchecked ${JAVAPROG}
     # Create a shared library containing all the native code: ddlogapi.c, ${DLPROG}_ddlog.a
-    ${CC} -shared -fPIC -I${JAVA_HOME}/include -I${JAVA_HOME}/include/${JDK_OS} -I${DLDIR}/${DLPROG}_ddlog -I../../lib ../ddlogapi.c -L${DLDIR}/${DLPROG}_ddlog/target/${BUILD}/ -l${DLPROG}_ddlog -o libddlogapi.${SHLIBEXT}
+    ${CC} -std=gnu11 -shared -fPIC -I${JAVA_HOME}/include -I${JAVA_HOME}/include/${JDK_OS} -I${DLDIR}/${DLPROG}_ddlog -I../../lib ../ddlogapi.c -L${DLDIR}/${DLPROG}_ddlog/target/${BUILD}/ -l${DLPROG}_ddlog -o libddlogapi.${SHLIBEXT}
```
mbudiu-vmw: are we using gnu specific extensions?
why not just -std=c11?

ryzhyk: We use `strdup`, which, as it turns out, is not part of the standard C library and requires GNU extensions.

PR #506, opened 2020-02-05
Control flow tests.

ryzhyk:
I sneaked three new control flow constructs (`continue`, `break`, and `return`) in the last PR. This PR merely adds more tests and fixes a bug found in the process.

@mbudiu-vmw, can you please look at this commit https://github.com/vmware/differential-datalog/commit/75f5b9867eedb5116de3ddd6bf1a15380844c8da in addition to commits in this PR?

### DIFF
--- a/test/datalog_tests/simple.dl
+++ b/test/datalog_tests/simple.dl
@@ -1139,6 +1139,7 @@ function all(xs: Vec<bool>): bool =
     return true
 }
 
+// At least one element of xs is true.
 function any(xs: Vec<bool>): bool =
 {
     var res = false;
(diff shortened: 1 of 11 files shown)


test/datalog_tests/simple.dl
```diff
@@ -1151,6 +1152,16 @@ function any(xs: Vec<bool>): bool =
     return res
 }
 
+// A weird implementation of any using continue inside return.
+function weird_any(xs: Vec<bool>): bool =
+{
+    var res = false;
+    for (x in xs) {
+        return if (x == true) { true } else { continue }
```
mbudiu-vmw: I don't think I have seen this kind of stuff before in any language.
People may find it confusing, but if it works as expected for the "normal" cases perhaps it's fine.

ryzhyk: Rust is ok with this. We can of course restrict the use of constructs that interrupt normal control flow (`continue`, `break`, and `return`) to avoid weird cases, but "weird" is hard to define, and oftentimes proves useful in practice...

PR #508, opened 2020-02-06
Disable failing souffle tests.

ryzhyk:
Several newly added souffle tests do not compile.

@mbudiu-vmw, one of these tests uses floating point types. I haven't looked into what's going on with others.

### DIFF
--- a/test/run-souffle-tests.py
+++ b/test/run-souffle-tests.py
@@ -44,6 +44,7 @@ xfail = [
     "cellular_automata", # issue 198 - order of clauses
     "comp-override2", # nested component declaration
     "components1",    # improper component nesting
+    "float_operations",
     "functor_arity",  # min, max, cat with more than 2 arguments
     "grammar",        # funny unicode char in a comment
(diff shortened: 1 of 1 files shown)


test/run-souffle-tests.py
```diff
@@ -44,6 +44,7 @@
     "cellular_automata", # issue 198 - order of clauses
     "comp-override2", # nested component declaration
     "components1",    # improper component nesting
+    "float_operations",
```
mbudiu-vmw: I will try to add float and double to ddlog.
not sure yet about literals, but at least the types.

ryzhyk: That'd be awesome, thanks. I can help with Rust bindings. Here's a relevant thread:
https://internals.rust-lang.org/t/pre-rfc-dealing-with-broken-floating-point/2673

Sounds like different CPUs can produce different results for the same floating operations.  This is going to be a problem for D3log. It also sounds like this should not be an issue for most modern desktops/servers, so maybe this is nothing to worry about.

PR #509, opened 2020-02-07
Antrea todos

ryzhyk:
- `uuid.dl` library for parsing and formatting UUIDs
- `net` library, currently including functions and types for working with IPv4 and IPv6 addresses.
- used both libraries in the antrea example.


lib/net/ipv4.dl
```diff
@@ -0,0 +1,126 @@
...
+ * IPv4 addresses are defined as 32-bit integers in IETF RFC 791. They are
+ * usually represented as four octets.
+ *
+ * See IpAddr for a type encompassing both IPv4 and IPv6 addresses.
+ *
+ * The size of an Ipv4Addr struct may vary depending on the target operating
+ * system.
+ */
+
+import net.ipv6
```
mbudiu-vmw: this is strange.

lib/net/ipv4.dl
```diff
@@ -0,0 +1,126 @@
...
+ * - 198.51.100.0/24 (TEST-NET-2)
+ * - 203.0.113.0/24 (TEST-NET-3)
+ */
+extern function ipv4_is_documentation(addr: Ipv4Addr): bool
+
+/* Converts this address to an IPv4-compatible IPv6 address.
+ *
+ * a.b.c.d becomes ::a.b.c.d
+ */
+extern function ipv4_to_ipv6_compatible(addr: Ipv4Addr): Ipv6Addr
```
mbudiu-vmw: I guess that's why you need that module.

lib/net/ipv6.dl
```diff
@@ -0,0 +1,134 @@
...
+ * IPv6 addresses are defined as 128-bit integers in IETF RFC 4291. They are
+ * usually represented as eight 16-bit segments.
+ *
+ * See IpAddr for a type encompassing both IPv4 and IPv6 addresses.
+ *
+ * The size of an Ipv6Addr struct may vary depending on the target operating
+ * system.
+ */
+
+import net.ipv4
```
mbudiu-vmw: I didn't know you could have mutually recursive modules.

lib/net/ipv4.dl
ryzhyk: Yep, fortunately we allow circular dependencies between modules.

PR #511, opened 2020-02-08
Pass group-by variables to aggregation functions.

ryzhyk:
Aggregation functions previously took the contents of the group as the
    only argument. In some cases, e.g., in the SQL converter, it is helpful
    to pass the group key as well.  We add this capability without changing
    existing syntax.  The only change is that the `Group` type is now
    parameterized with both key and value types.  A program can iterate over
    group values as before; in addition, the `group_key(Group<'K,'V>): 'K`
    function extracts the group key.
    
    Example:
    
    ```
    // Custom aggregation function that group of (vendor, price) tuples and
    // uses item name as key, and returns a string containing item name, along
    // with the lowest price and matching vendor for this item.
    function best_vendor_string(g: Group<string, (string, bit<64>)>): string =
    {
        var min_vendor = "";
        var min_price: bit<64> = 'hffffffffffffffff;
        for (vendor_price in g) {
            if (vendor_price.1 < min_price) {
                min_vendor = vendor_price.0;
                min_price = vendor_price.1
            }
        };
        "Best deal for ${group_key(g)}: ${min_vendor}, $${min_price}"
    }
    
    output relation BestDeal(best: string)
    BestDeal(best) :-
        Price(item, vendor, price),
        var best = Aggregate((item), best_vendor_string((vendor, price))).
    ```

### DIFF
--- a/doc/tutorial/tutorial.md
+++ b/doc/tutorial/tutorial.md
@@ -919,7 +919,7 @@ describes constraints on the recursive programs accepted by DDlog.
 #### Aggregation
 
 The `Aggregate` operator groups records that have the same values of a subset of
-variables and applies an aggregation function to each group.
+variables (group-by variables) and applies an aggregation function to each group.
 The following program groups the `Price` relation by `item` and selects the minimal
 price for each item:
 
@@ -933,7 +933,19 @@ BestPrice(item, best_price) :-
 ```
 
 Here `group_min()` is one of several aggregation functions defined in
-`lib/std.rs`.  What if we wanted to return the name of the vendor along with the
+`lib/std.rs`.
+
+It is possible to group a relation by multiple fields.  For
+example, the following rule computes the lowest price that each vendor
+charged for each item:
+
+```
+BestPricePerVendor(item, vendor, best_price) :-
+    Price(.item = item, .vendor = vendor, .price = price),
+    var best_price = Aggregate((item, vendor), group_min(price)).
+```
+
+What if we wanted to return the name of the vendor along with the
 lowest price for each item?  The following naive attempt will not work:
 
 ```
@@ -958,7 +970,7 @@ aggregation functions:
 
 ```
 /* User-defined aggregate that picks a tuple with the smallest price */
-function best_vendor(g: Group<(string, bit<64>)>): (string, bit<64>) =
+function best_vendor(g: Group<'K, (string, bit<64>)>): (string, bit<64>) =
 {
     var min_vendor = "";
     var min_price: bit<64> = 'hffffffffffffffff;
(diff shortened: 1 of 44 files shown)


doc/tutorial/tutorial.md
```diff
@@ -987,6 +987,34 @@ BestPricePerVendor(item, vendor, best_price) :-
     var best_price = Aggregate((item, vendor), group_min(price)).
 ```
 
+The aggregation function can take group-by variables in addition to
+the group.  The following custom aggregation function computes the cheapest
+vendor for each item and returns a string containing item name, vendor,
+and price:
+
+```
+function best_vendor_string(item: string, g: Group<(string, bit<64>)>): string =
```
mbudiu-vmw: Frankly I find this quite confusing. The signature of the function is not the same in the declaration and in the call.
Is backwards compatibility so important? If you want to make changes to the language you should do them while it still has few users. You can provide a good diagnostic to help people migrate their code. 

I think it's cleaner if the Group object just has a key field. Group should be templated on two types, the key type and the value type. If you really want backwards compatibility you can have the for operate directly on the group. In LINQ this is done by the group implementing the `IEnumerable<Value>` interface.

mbudiu-vmw: Let's run this by Ben, he has very good taste in software design.

lib/std.dl
```diff
@@ -131,32 +131,37 @@ extern function hash128(x: 'X): bit<128>
 
 /* The `Group` type is used exclusively in aggregation operations.  It
  * represents a non-empty list of objects sorted in ascending order.
+ * `'K` is the type of group key, and `'V` is the type of value in the group.
  */
-extern type Group<'A>
+extern type Group<'K,'V>
```
mbudiu-vmw: this is actually very nice.

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -1502,9 +1502,12 @@ mkAggregate d filters input_val rl@Rule{..} idx = do
     -- - compute aggregate
     -- - return variables still in scope after this term
     let tmap = ruleAggregateTypeParams d rl idx
-    let tparams = commaSep $ map (\tvar -> mkType (tmap M.! tvar)) $ funcTypeVars $ getFunc d rhsAggFunc
+    let agg_func = getFunc d rhsAggFunc
+    -- Pass group-by variable to the aggregate function if it accepts them.
```
mbudiu-vmw: is this comment still accurate?

ryzhyk:
@mbudiu-vmw , I'm about to merge.  I had to fixup the SQL converter to pass the tests.  Please have a look if my changes make sense (I imagine you will end up overriding them anyway in your upcoming commits).

sql/src/test/java/ddlog/QueriesTest.java
```diff
@@ -126,7 +126,7 @@ public void testTwoAggregations() {
         String query = "create view v1 as SELECT MIN(column1) + MAX(column1) FROM t1";
         String program = this.header(false) +
                 "typedef Ttmp0 = Ttmp0{col4:signed<64>}\n" +
-                "function agg1(g1: Group<(Tt1)>):Ttmp0 =\n" +
+                "function agg1(g1: Group<'K, (Tt1)>):Ttmp0 =\n" +
```
mbudiu-vmw: We will probably generate only concrete types here instead of a type variable, but this should work for now.

PR #527, opened 2020-02-17
Introduce insert-or-update command in the C API.

### LINKED ISSUE #525: Implement insert_or_update API
Useful when the client does not know if the key exists in the input relation.

@antoninbas

ryzhyk:
Introduce insert-or-update command.

The new command is useful when working with DDlog input relations with
primary keys.  Operations available on such relations so far are:

- insert - fails if a record with the same key exists
- delete-by-value, delete-by-key - fail if a record with the specified
  key does not exist
- modify - fails if a record with the specified key does not exist.

The new command is similar to insert, except that, if a record with the
same key as the one being inserted already exists, it replaces this
record with the new one instead of failing.  This is convenient in
scenarios where the client does not know if the key exists or cannot
easily issue a delete command.

The insert-or-update command is only applicable to input relations with
a primary key.

The CLI syntax for the new command is the same as for the `insert`
command, but using the `insert_ord_update` keyword:

```
insert_or_update WithKey(1, "bar3");
```

C API:

```
extern ddlog_cmd* ddlog_insert_or_update_cmd(table_id table, ddlog_record *rec);
```

This resolves #525.

### DIFF
--- a/rust/template/ddlog.h
+++ b/rust/template/ddlog.h
@@ -332,7 +332,6 @@ extern int ddlog_dump_ovsdb_delta(ddlog_prog hprog, const char *module,
  */
 extern void ddlog_free_json(char *json);
 
-
 /*
  * Apply updates to DDlog tables.  See the ddlog_cmd API below.
  *
@@ -342,6 +341,13 @@ extern void ddlog_free_json(char *json);
  * Whether the function succeeds or fails, it consumes all commands in
  * the `upds` array (but not the array itself), so they can no longer be
  * accessed by the caller after the function returns.
+ *
+ * This function fails non-atomically: upon a failure, it may have applied
+ * a _subset_ of input commands (one exception is when `ddlog_apply_updates`
+ * is called with a single command, in which case it leaves the database
+ * unmodified in case of a failure).  Use `ddlog_transaction_rollback()` to
+ * bring the database back to a known state, specifically the state where it
+ * was before the start of the transaction.
  */
 extern int ddlog_apply_updates(ddlog_prog prog, ddlog_cmd **upds, size_t n);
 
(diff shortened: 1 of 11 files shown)


rust/template/ddlog.h
```diff
@@ -342,6 +341,13 @@ extern void ddlog_free_json(char *json);
  * Whether the function succeeds or fails, it consumes all commands in
  * the `upds` array (but not the array itself), so they can no longer be
  * accessed by the caller after the function returns.
+ *
+ * This function fails non-atomically: upon a failure, it may have applied
+ * a _subset_ of input commands (one exception is when `ddlog_apply_updates`
+ * is called with a single command, in which case it leaves the database
+ * unmodified in case of a failure).  Use `ddlog_transaction_rollback()` to
+ * bring the database back to a known state, specifically the state where it
+ * was before the start of the transaction.
```
mbudiu-vmw: I think we had an issue for this behavior not documented at the time

rust/template/ddlog.h
```diff
@@ -1053,35 +1059,73 @@ extern const ddlog_record* ddlog_get_struct_field(const ddlog_record* rec,
...
  *
  * Returns pointer to a new command, which can be sent to DDlog by calling
  * `ddlog_apply_updates()`.
+ *
+ * This function never fails; however the command it creates may fail to
+ * execute, causing `ddlog_apply_updates()` to return an error if:
+ * - `table` is not a valid input table id
+ * - `rec` does not match the record type of `table`
+ * - `table` has a primary key and record `rec` does not exist in `table`
+ *   (NOTE: for tables without a primary key, the command succeeds even if
```
mbudiu-vmw: This sounds wrong

ryzhyk: Agreed, but at least this documents how things are working right now.

PR #528, opened 2020-02-20
JSON support

ryzhyk:
Infrastructure to serialize/deserialize DDlog types from JSON.  Please see individual commit messages for details.

### DIFF
--- a/test/datalog_tests/json_test.dl
+++ b/test/datalog_tests/json_test.dl
@@ -1,0 +1,79 @@
+import json
+
+output relation JsonTest(description: string, value: string)
+
+
+typedef BoolStruct = BoolStruct {
+    b: bool
+}
+
+function boolStruct1(): string = [|{"b":true}|]
+function boolStruct2(): string = [|{"b":true, "foo":"bar"}|]
+function boolStruct3(): string = [|{"foo":"bar"}|]
+
+JsonTest(boolStruct1(),
+         result_unwrap_or_default(to_json_string(from_json_string(boolStruct1()): Result<BoolStruct, string>))).
+JsonTest(boolStruct2(),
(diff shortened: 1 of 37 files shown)


test/datalog_tests/json_test.dl
```diff
@@ -0,0 +1,107 @@
...
+typedef BoolStruct = BoolStruct {
+    b: bool
+}
+
+function boolStruct1(): string = [|{"b":true}|]
+function boolStruct2(): string = [|{"b":true, "foo":"bar"}|]
+function boolStruct3(): string = [|{"foo":"bar"}|]
+
+JsonTest(boolStruct1(),
+         result_unwrap_or_default(to_json_string(from_json_string(boolStruct1()): Result<BoolStruct, string>))).
```
mbudiu-vmw: This Look like it could use a little helper function.

test/datalog_tests/tutorial.ast.expected
```diff
@@ -270,6 +270,7 @@ function std.result_unwrap_or (res: std.Result<'V,'E>, def: 'V): 'V =
         std.Ok{.res=var v} -> v,
         std.Err{.err=_} -> def
     }
+extern function std.result_unwrap_or_default (res: std.Result<'V,'E>): 'V
```
mbudiu-vmw: I wonder whether we could somehow automatically prepend std to avoid always changing all reference outputs

test/datalog_tests/json_test.dl
ryzhyk: Unfortunately, a function won't work, as I need to sneak the type annotations in-between the two transformations (otherwise it's just a function from strings to strings and there is no way for the compiler to figure out what the deserialized type should be).

I guess I can at least use

``` function to_json_string_or_default(x: 'T): string```

PR #529, opened 2020-02-21
Add Go bindings for DDlog C API

antoninbas:
The code uses [cgo](https://golang.org/cmd/cgo/) to call the DDlog C
code.

Unlike the C API, the Go bindings introduce *some* type-safety. All
record types share a base interface (ddlog.Record). But some records,
and in particular compound types (vectors, tuples, etc.), benefit from a
more specific interface (RecordVector, RecordTuple, etc.) with
additional methods. For example, unlike with the C API, methods specific
to vector records (Push, At, Size) cannot be called on a string or
integer record without triggering a build error. This type-safety comes
at a small cost compared to simply wrapping the ddlog_record pointers
into a Go struct, since using Go interfaces introduce
indirection. However, when running a benchmark creating "complex"
records, there was less than 25% overhead. In a real system where DDlog
actually has to process transactions, the difference will be
neglectable.


antoninbas:
TODO:
 * [ ] add README to describe usage (how to link to the DDlog library)
 * [ ] add unit tests using a sample DDlog program
 * [ ] integrate with Travis CI

PR #534, opened 2020-02-25
Access fields of stucts & tuples wrapped in Ref's.

ryzhyk:
This commit addresses one of the awkward aspects of working with complex
types wrapped in `Ref<>`'s.  Ideally, we would like `Ref<>`'s to be as
transparent as possible, so that one could work with them as if they were
objects of the inner type wrapped in the `Ref`.  We're not quite there
yet.  One issue is that Rust does not allow matching on the contents of
a `Ref`.  The other issue is that accessing fields of structs or tuples
wrapped in a `Ref<>` requires either unwrapping it using `deref` or
using pattern matching syntax, as in:

```
relation StudentInfo(student: Ref<Student>, school: Ref<School>)

TopScore(school, top_score) :-
    StudentInfo(&Student{.sat_score = sat}, &School{.name = school}),
    // WE HAVE BOUND THE `.sat_score` FIELD OF Student TO `sat` AND CAN
    // USE IT BELOW.
    var top_score = Aggregate((school), group_max(sat)).
```

This commit extends the field access syntax `s.field` to perform automatic
dereferencing if `s` is a `Ref<>`, so the above rule can now be equivalently
re-written as:

```
TopScore(school, top_score) :-
    StudentInfo(student, &School{.name = school}),
    var top_score = Aggregate((school), group_max(student.sat_score)).
```

### DIFF
--- a/doc/tutorial/tutorial.md
+++ b/doc/tutorial/tutorial.md
@@ -1393,6 +1393,17 @@ TopScore(school, top_score) :-
 
 We again use `&` to pattern match values stored by reference.
 
+As another syntactic convenience, DDlog allows accessing fields of structs and
+tuples wrapped in references directly, without dereferencing them first.
+The following rule is equivalent to the one above, but instead of opening up the
+`student` record, we use `.`-notation to access the `sat_score` field:
+
+```
+TopScore(school, top_score) :-
+    StudentInfo(student, &School{.name = school}),
+    var top_score = Aggregate((school), group_max(student.sat_score)).
+```
+
 ## A more imperative syntax
 
 Some people dislike Datalog because the evaluation order of rules is not always obvious.  DDlog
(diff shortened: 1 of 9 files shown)


doc/tutorial/tutorial.md
```diff
@@ -1393,6 +1393,17 @@ TopScore(school, top_score) :-
 
 We again use `&` to pattern match values stored by reference.
 
+As another syntactic convenience, DDlog allows accessing fields of structs and
+tuples wrapped in references directly using, without dereferencing them first.
```
mbudiu-vmw: missing something here

ryzhyk:
> I wonder whether Ref could be made even more transparent, i.e., inheriting all operations of the argument type, not just dot. The question is: when do you want something to stay Ref?

Perhaps we should hide Ref's altogether, but they does have big performance implications (both positive and negative if abused). Plus the contents of Ref's is immutable.  Finally cyclic Ref's are not allowed (this is not checked automatically yet).  I don't know what the best design is.  At least this fix gets rid of some ugliness.

PR #543, opened 2020-03-02
Deserialize records on the fly.

ryzhyk:
Consider a scenario where input relations contain JSON strings.  These
strings are parsed using the recently added JSON lib and then DDlog goes
on to compute on the parsed data.  When working with large JSON
documents where only a handful of attributes are actually used by DDlog,
DDlog's memory footprint is dominated by unparsed JSON inputs.

To avoid this waste, we introduce a new record type that contains
serialized data represented as a string.  DDlog parses this string using
appropriate deserializer (currently, only json is supported) and drops
the string on the floor, keeping only the extracted data in memory.

Syntactically, the new type of record is written down as
`@<format><string_token>`, where `<string_token>` can be either a quoted
string literal or a reference to a file, e.g.:

```
insert json_test.Deserialized[@json"{\"@type\": \"t.V1\", \"b\": true}"]
```

### DIFF
--- a/rust/template/cmd_parser/parse.rs
+++ b/rust/template/cmd_parser/parse.rs
@@ -314,7 +314,7 @@ named!(rel_key<&[u8], (Name, Record)>,
 );
 
 named!(record<&[u8], Record>,
-    alt!(bool_val | string_val | string_val_from_file | tuple_val | array_val | struct_val | int_val )
+    alt!(bool_val | string_val | serialized_val | tuple_val | array_val | struct_val | int_val )
 );
 
 named!(named_record<&[u8], (Name, Record)>,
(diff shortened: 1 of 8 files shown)


rust/template/cmd_parser/parse.rs
```diff
@@ -406,6 +414,28 @@ fn test_string() {
     );
 }
 
+named!(serialized_val<&[u8], Record>,
+    do_parse!(
+        tag!("@")
```
mbudiu-vmw: Do you want to document this particular format?
How does one register a new deserializer?

mbudiu-vmw:
your PR comment looks wrong, the format seems to start with @

rust/template/cmd_parser/parse.rs
ryzhyk: yes, there are a couple of other recent undocumented extensions.  Let will submit a separate documentation PR.

ryzhyk:
> your PR comment looks wrong, the format seems to start with @

Right, I fixed the commit message, but not the PR....

ryzhyk:
> Looks a bit like magic. I wonder what other applications there may be.

The magic is really in the Rust `serde` crate.

> Should we make it possible to register serializers in DDlog? (like _2string functions)?
> I can't imagine writing deserializers in ddlog.

Not sure about per-type deserializers, but we can definitely make it possible to register de-serialization formats. I suggest we don't generalize too early though. Let's wait until we have at least one other format...

PR #545, opened 2020-03-05
Support for floating points

### LINKED ISSUE #448: Add support for floating point types
These could almost be extern types, but we probably want support for parsing double/float literals.

ryzhyk:
This is @mbudiu-vmw 's earlier PR with some changes.

Fixes #448


lib/json.dl
```diff
@@ -40,6 +40,6 @@ typedef JsonNum = // Integer number.  The value must be in the range between the
                   // inclusive.  Values outside of this range will be truncated
                   // during serialization.
                   JsonInt{i: s128}
-                | // Always finite.
-                  // TODO.
-                  JsonFloat
+                | // NaN and infinity are not valid JSON values.
```
mbudiu-vmw: I don't like this.
Perhaps JSON null would be a better choice.

rust/template/cmd_parser/parse.rs
```diff
@@ -563,6 +564,70 @@ fn test_struct() {
...
+named!(recognize_exponent<&[u8], ()>,
+       do_parse!(alt!(char!('e') | char!('E')) >>
+                 opt!(alt!(char!('+') | char!('-'))) >>
+                 digit >>
+                 (()))
+);
+
+// We do not have enough type information to distinguish between floats
+// and doubles during parsing, so we parse all floating point numbers as
+// double's and postpone casting them to floats until when they get converted
```
mbudiu-vmw: doubles, no apostrophe.
you may want a similar comment for bigint if it isn't there

rust/template/cmd_parser/parse.rs
```diff
@@ -563,6 +564,70 @@ fn test_struct() {
...
+// into concrete types.
+named!(float_val<&[u8], Record>,
+       do_parse!(val: flat_map!(recognize_float, parse_to!(f64)) >>
```
mbudiu-vmw: I wonder what happens on overflow

ryzhyk: It will be rounded up to infinity

lib/json.dl
ryzhyk: ok, null it is.

PR #524, opened 2020-02-15
group by implementation

mbudiu-vmw:
We still don't have 'having' yet.

PR #547, opened 2020-03-06
GroupBy and Having implementation

mbudiu-vmw:


mbudiu-vmw:
This includes the work in #524, so I will close that one.

mbudiu-vmw:
@lalithsuresh : you have reviewed part of this previously.
now this adds support for having, so we can compile two more of the weave queries.
if you want to review I would suggest reading the tests in QueryTests that show various translations. The weave queries are rather large. It may be that the work we do is a bit inefficient - we copy lots of fields on joins. We'll have to see whether this needs to be optimized somehow.

lalithsuresh:
Those indeed look large. I guess the windowed aggregates are what's missing now?

For having(), I suggest also adding tests where the having clause has more than one aggregate (count(*) > 1 and sum(blah) == 4)

PR #548, opened 2020-03-07
Support DDlog u128 records in the Go library

antoninbas:
Note that we only support unsigned 128-bit numbers for now.

### DIFF
--- a/go/pkg/uint128/uint128.go
+++ b/go/pkg/uint128/uint128.go
@@ -1,0 +1,71 @@
+package uint128
+
+import (
+	"encoding/binary"
+	"encoding/hex"
+
+	"github.com/google/uuid"
+)
+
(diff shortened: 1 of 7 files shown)


go/pkg/uint128/uint128.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"encoding/binary"
+	"encoding/hex"
+
+	"github.com/google/uuid"
+)
+
+// Uint128 stores a 128-bit unsigned integer as two 64-bit words. Two Uint128 values can be compared
+// with ==.
+type Uint128 struct {
```
ryzhyk: One thing I love about Rust is the `u128` built-in type  ;)

antoninbas: Maybe it will come to Go eventually: https://github.com/golang/go/issues/9455...

PR #549, opened 2020-03-08
Update README.md

ryzhyk:
A more user-friendly summary of DDlog focusing on its capabilities.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,16 +5,29 @@
 
 # Differential Datalog (DDlog)
 
-DDlog is a *bottom-up*, *incremental*, *in-memory*, *typed* Datalog engine. It is well suited for
-writing programs that incrementally update their output in response to input changes. With DDlog,
+DDlog is a programming language for *incremental computation*. It is well suited for
+writing programs that continuously update their output in response to input changes. With DDlog,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,16 +5,29 @@
...
 Instead they specify the desired input-output mapping in a declarative manner, using a dialect of Datalog.
 The DDlog compiler then synthesizes an efficient incremental implementation.
 DDlog is based on [Frank McSherry's](https://github.com/frankmcsherry/)
 excellent [differential dataflow](https://github.com/frankmcsherry/differential-dataflow) library.
 
-1. **Bottom-up**: DDlog starts from a set of *ground facts* (i.e., facts provided by the user) and
-computes *all* possible derived facts by following Datalog rules, in a bottom-up fashion.  In
+DDlog has the following key properties:
+
+1. **Relational**: A DDlog program transforms a set of input relations (or tables) into a set of output relations.
```
mbudiu-vmw: The fact that we can store rich objects and collections as cells in tables goes far behind the relational model. you don't have to do weird stuff like shredding json to access it efficiently. So I think this should be mentioned somewhere, not necessarily here.

PR #553, opened 2020-03-10
Added DDLOG_HOME environment variable

mbudiu-vmw:
This is useful, for example, because the standard library path will automatically include $DDLOG_HOME/lib.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -129,6 +129,7 @@ main :: IO ()
 main = do
     args <- getArgs
     prog <- getProgName
+    home <- lookupEnv "DDLOG_HOME"
     config <- case getOpt Permute options args of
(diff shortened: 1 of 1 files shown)


app/Main.hs
```diff
@@ -137,13 +138,16 @@ main = do
                                       (\e -> do putStrLn $ usageInfo ("Usage: " ++ prog ++ " [OPTION...]") options
                                                 throw (e::SomeException))
                    _ -> errorWithoutStackTrace $ usageInfo ("Usage: " ++ prog ++ " [OPTION...]") options
-    case confAction config of
+    config' <- case home of
+         Just(p) -> addOption config (LibDir $ p ++ "/lib")
```
ryzhyk: This happens to parse, but `Just p` would be more regular Haskell.

Issue #550, opened 2020-03-08
Make install-dependencies.sh more robust


ryzhyk:
The script needs to be more robust and user-friendly in several ways.  We advertise it as a recommended way to prepare the system for DDlog, but in practice it does not seem to do the right thing for most people.

- The installation consists of three sections:
  1. The minimal set of components needed to use DDlog.  These always get installed.
  1. Additional components needed to use DDlog from java or golang.
  1. (dev-only) Software needed to compile and test DDlog.
  The script should determine if it's running inside the repo and, if so, install everything.  Otherwise, the user is working with a binary distro and we should install group 1 and ask if they also want to install 2.
  For a complete list of prerequisites, see `tools/Dockerfile`, which has everything needed to configure a vanilla ubuntu system from scratch.

- Check for existing software:
  - Rust
    - `rustc --version`. If there is no Rust compiler installed, run `rustup`. If there is a compiler, check that it's `1.39.0+` and if not, run `rustup update && rustup default stable`.
  - Haskell stack (dev-only).
   - `stack --version`. If no stack, install stack. If stack is present, but old, run `stack upgrade`.
  - Java
  - flatc
  - golang.
  - python (dev-only)

- Help setup `$CLASSPATH` and `$PATH` (stack)
  - It is not particularly useful that the script only configures the current shell. Besides, most people will forget to source vs run the script. I don't know if there is a standard way to make permanent changes to user's environment, but at least we can print necessary environment variable changes in the end of the script.
  - `install-stack.sh` assumes that `~/.local/bin` is in the path. If we installed stack there and it's not in the path, make sure we add it to `$PATH` or remind the user to do so.
  
- The script should work when invoked from any directory (by determining the path to itself and adjusting the rest of its logic accordingly)

- The script should warn the user about some changes, e.g., before installing the `default-jdk` package it should check that this is what the user wants and give them the option to install their own JDK.

- The script should finish by compiling a test DDlog program. If this fails, the user will have an error report to send to us.

- We should test the script on a bunch of systems to make sure it handles various combinations of preexisting software reasonably well.

- Once ready, the script can be included in the binary distro.

@mbudiu-vmw

danbst:
I'm a random stranger here... If you don't target native Windows, then you may consider using [Nix](https://nixos.org/nix/) and `nix-shell` for development shell setup.

I can invest some time to provide a demo how would it look like.

ryzhyk:
[...]

PR #555, opened 2020-03-14
setup prerequisites and dependencies with Nix

danbst:
- ~~had to disable static binary compilation, due to some `gmp` problems with Nix infra~~ static binary build is supported
- didn't install Go into profile, cause don't know how to test it
- `stack test --ta '-p path'` works
- the mentioned in tutorial small script now lives as a shell function `ddlog-run`

Packaging `ddlog` as a "package" can be next step, but it will require some thinking. It will definitely require converting `package.yaml` and `Cargo.toml` into Nix files. But `nix-shell` approach is still fine for grab&try experience.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,90 @@
+{ ghc ? null
+, pkgs ? import <nixpkgs> { }
+} @ args:
+
+let
+
+  ghc = args.ghc or pkgs.haskell.compiler.ghc865;
+  java = pkgs.adoptopenjdk-bin;
+  rust-toolchain = "1.39.0";
+
+  flatbuffers-java = pkgs.runCommand "flatbuffers-java" {
+    buildInputs = [ java pkgs.findutils ];
+  } ''
+    mkdir -p $out
+    cp -r ${pkgs.flatbuffers.src}/java $out/
+    chmod -R +w $out/java
+    javac $out/java/com/google/flatbuffers/*.java
+    rm -rf $out/java/com/google/flatbuffers/*.java
+  '';
+
+  stack-shell = pkgs.haskell.lib.buildStackProject {
+    inherit ghc;
+    name = "stack-shell";
+
+    CLASSPATH = "${flatbuffers-java}/java";
+
+    shellHook = ''
+      export PATH="$HOME/.local/bin:$PATH"
+    '';
+
+    buildInputs = [
+      pkgs.flatbuffers
+      pkgs.zlib
+      pkgs.adoptopenjdk-bin
+      pkgs.gitMinimal
+      pkgs.rustup
+      pkgs.cacert
+    ];
+  };
+
+  dev-shell = stack-shell.overrideAttrs (old: {
+    buildInputs = old.buildInputs ++ [
+      pkgs.stack
+    ];
+
+    shellHook = (old.shellHook or "") + ''
+
+      unset STACK_IN_NIX_SHELL
+
+      function initial-setup {
+        rustup default ${rust-toolchain}
+        stack build
+        stack install
+      }
+
+      function ddlog-run {
+        local program="$1"
+        local program_noext=''${program%.*}
+
+        if [ ! -f "$program" ]; then
+          >&2 echo "You are trying to run $program, but it does not exist"
+          return
+        fi
+
+        local input_file="$2"
+        if [ -z "$input_file" ]; then
+          local dat_name="$program_noext.dat"
+          if [ -f "$dat_name" ]; then
+            >&2 echo "Using $dat_name as an input file"
+            input_file="$dat_name"
+          fi
+        fi
+        ddlog -i "$program" -L '${builtins.toPath ./lib}'
+        (cd "''${program_noext}_ddlog" && cargo build --release)
+        if [ -z "$input_file" ]; then
+          "''${program_noext}_ddlog/target/release/$(basename "$program_noext")_cli"
+        else
+          "''${program_noext}_ddlog/target/release/$(basename "$program_noext")_cli" --no-print < "$input_file"
+        fi
+      }
+    '';
+  });
+
+in dev-shell // {
+
+  inherit 
+    stack-shell
+  ;
+
+}
--- a/default.nix
+++ b/default.nix
@@ -47,6 +47,8 @@ let
 
       unset STACK_IN_NIX_SHELL
 
+      alias stack="stack --nix-shell-file ${builtins.toPath ./.}/tools/stack-shell.nix --nix-pure --no-install-ghc --skip-ghc-check"
+
       function initial-setup {
         rustup default ${rust-toolchain}
         stack build
--- a/default.nix
+++ b/default.nix
@@ -71,11 +71,11 @@ let
           fi
         fi
         ddlog -i "$program" -L '${builtins.toPath ./lib}' || return
-        (cd "''${program_noext}_ddlog" && cargo build)
+        (cd "''${program_noext}_ddlog" && cargo build --release)
         if [ -z "$input_file" ]; then
-          "''${program_noext}_ddlog/target/debug/$(basename "$program_noext")_cli"
+          "''${program_noext}_ddlog/target/release/$(basename "$program_noext")_cli"
         else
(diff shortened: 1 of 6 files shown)


default.nix
```diff
@@ -70,8 +70,8 @@ let
             input_file="$dat_name"
           fi
         fi
-        ddlog -i "$program" -L '${builtins.toPath ./lib}'
-        (cd "''${program_noext}_ddlog" && cargo build --release)
```
ryzhyk: There are a couple of problems with debug builds in Rust: they are 10x slower (literally) and also take 10x more disk space than release builds. The release build is indeed slow, but after a recent refactoring only the first build is really slow; subsequent rebuilds usually take a few seconds, so I would stick to release builds here.

ryzhyk:
Thanks for contributing! I'm gonna need a bit of time to try it out (and hopefully learn Nix a bit along the way) and review the PR.

default.nix
danbst: indeed, rebuild with `--release` is only a few seconds longer than with `--debug` on a multicore machine. And disk space usage is greatly reduced. I'll revert back

danbst:
found and fixed one potential issue (tested it works on Ubuntu 17.10). Now this project uses GHC exclusively from Nix, which actually speeds up initial setup, since no need for Stack to download GHC.

The version of GHC specified in `default.nix` is now definitional, LTS-specified ghc version isn't used.

ryzhyk:
Does using Nix to install dependencies necessarily imply that users must work in nix-shell? We already use a pretty idiosyncratic set of tools and requiring users to learn `nix-shell` is adding one more hoop to jump through.

danbst:
> Does using Nix to install dependencies necessarily imply that users must work in nix-shell?

kinda yes. Note that `nix-shell` in nutshell is starting a bash process and setting up certain environment variables (PATH, CLASSPATH, PS1, LD_LIBRARY_PATH, and several others), which include all the needed software.

> We already use a pretty idiosyncratic set of tools...

this is where `nix-shell` approach shines - users don't have to care about that AT ALL. This is why I've removed all the prerequisites section, because users no longer have to even know, for example, that they require flatbuffers to run simple CLI.

If you have multi-tool multi-language project setup, then Nix and Docker are best friends here, with main difference being that Nix doesn't require containerization.

> ... and requiring users to learn nix-shell is adding one more hoop to jump through.

I'd argue it is one less hoop, because users no longer have to care about installing all the required tools. See, for example, `buildInputs`:
```
    buildInputs = [
      pkgs.flatbuffers
      pkgs.zlib
      pkgs.adoptopenjdk-bin
      pkgs.gitMinimal
      pkgs.rustup
      pkgs.cacert
    ];
```
This roughly corresponds to `apt install flatbuffers zlib openjdk git rustup`, but distro-agnostic (will work on any ubuntu and arch linux), so users don't even have to know about these tools, whether they are present in repositories, how to install that from sources and  so on.

It's possible to run the `ddlog` without entering `nix-shell`.
1. similar to `docker run ....` you can use `nix-shell --run "ddlog arg1 arg2"`
2. you can export and save environment with `nix-shell --run "declare -px" > .env`, which you can `source .env` later (`virtualenv` approach).
```
❯ ddlog
ddlog: command not found
❯ nix-shell --run "declare -px" > .env
❯ bash -c 'source .env; ddlog'
Usage: ddlog [OPTION...]
  -h       --help                           Display help message.
  -v       --version                        Display DDlog version.
  -i FILE                                   DDlog program to compile.
           --action=ACTION                  Action: [validate, compile]
  -L PATH                                   Extra DDlog library directory.
  -o DIR   --output-dir=DIR                 Output directory (default based on program name).
           --dynlib                         Generate dynamic library.
  -j       --java                           Generate Java bindings.
           --output-internal-relations      All non-input relations are marked as output relations.
           --output-input-relations=PREFIX  Mirror each input relation into an output relation named by prepending the prefix.
           --no-dynlib                      Do not generate dynamic library (default).
           --staticlib                      Generate static library (default).
           --no-staticlib                   Do not generate static library.

ddlog: input file not specified
```

These 2 things definitely require learning of `nix-shell` arguments and behavior. But don't require knowing Nix syntax.

danbst:
Back to flatbuffers case. I've extracted code related to installing that
```
  java = pkgs.adoptopenjdk-bin;
  ...

  flatbuffers-java = pkgs.runCommand "flatbuffers-java" {
    buildInputs = [ java ];
  } ''
    mkdir -p $out
    cp -r ${pkgs.flatbuffers.src}/java $out/
    chmod -R +w $out/java
    javac $out/java/com/google/flatbuffers/*.java
    rm -rf $out/java/com/google/flatbuffers/*.java
  '';

  stack-shell = pkgs.haskell.lib.buildStackProject {
    ...
    CLASSPATH = "${flatbuffers-java}/java";
    buildInputs = [
      ...
      pkgs.flatbuffers
      java
      ...
    ];
```
Note how this will use both source of a package (`pkgs.flatbuffers.src`), a binary result (`pkgs.flatbuffers`) and a compiled Java classet (`flatbuffers-java`). Users don't have to compile `flatbuffers` themselves, but still can set CLASSPATH with correct library.

ryzhyk:
Thanks for the explanation! At this point I think we should support Nix-based build as an option (perhaps, the primary one), but still document manual dependency installation as we do now and improve `install-dependencies.sh`. If that sounds like a good plan to you, I can make the changes and submit a PR against your PR (probably within a couple of days).

One more question: What steps should users take when pulling a new version of ddlog from git with potentially updated dependencies (e.g., we might require rust 1.42+ or a more recent Haskell LTS in the future)? I suspect there is a nix command to magically update all dependencies?

danbst:
> One more question: What steps should users take when pulling a new version of ddlog from git with potentially updated dependencies (e.g., we might require rust 1.42+ or a more recent Haskell LTS in the future)? I suspect there is a nix command to magically update all dependencies?

no magic here, just run `nix-shell`, `stack build` and `cargo build`. Each of these tools will manage the spec changes in an incremental fashion.

danbst:
@ryzhyk ok, I found the necessary libraries required to make static linking. So now `package.yaml` has no Nix-specific fixes. :tada: 

But also note, that static linking with glibc isn't ideal. There is a dedicated project which brings musl-based linking to Nix-based Haskell projects, which can be explored in future.
https://github.com/nh2/static-haskell-nix

ryzhyk:
@danbst, I think this is ready to be merged. Really appreciate your help with Nix integration, including fixing all the issues. My main concern is that none of us is familiar with Nix and I'm not sure we'll have the cycles to learn it any time soon, but given that we still support the manual installation option this is not a show stopper.

The PR says that you're not sure how to test Go integration. It should be a matter of running the `go/test.sh` script. I believe that the only prerequisite for that is installing the Go compiler itself. If you have time to add it to the Nix script, it will be very much appreciated.  Otherwise, no worries, I think we can do without Go support in Nix for now.

Just curious: how did you come across DDlog?  Are you interested in using it in one of your projects or just helping out?

danbst:
> Just curious: how did you come across DDlog?

I need a triple store, or graph database with some datalog-like processing, so I grepped github.com for "datalog" keyword. Not sure if bottom-up is really what I want, but right now I'm fine with it as my dataset is of near zero size :) I also like the type system here, multiline strings was one feature I really wanted.

danbst:
> It should be a matter of running the go/test.sh script. 

okay, added `go`, now `go/test.sh` passes

> we'll have the cycles to learn it any time soon

it's a pity, but understandable. This also means the issue https://github.com/vmware/differential-datalog/issues/550 is not yet resolved, as you still need a non-Nix way to make robust prerequisites setup. Without docker or Nix, I believe you'll be in trouble :)

danbst:
> I installed glibc and gmp packages using nix-env, but the linker does not seem to find static libraries from these packages.

btw, I'd suggest you to uninstall those. `nix-shell` now covers that.

ryzhyk:
> > It should be a matter of running the go/test.sh script.
> 
> okay, added `go`, now `go/test.sh` passes

Thanks!

PR #560, opened 2020-03-18
MacOS compatibility issue.

ryzhyk:
We use the `SOCK_CLOEXEC` flag to create sockets that are not inherited
by child processes.  MacOS apparently does not support the `SOCK_CLOEXEC` flag.
It appears that older versions of `libc` crate allowed using this flag
even on MacOS (presumably, with no effect), but this hole was closed in
`libc 0.2.68`.  The workaround is to use a separate `fcntl` call to set
this flag.  This is generally not recommended due to the potential race
condition when a parallel thread that forks the current process at the
same time, but I don't think this scenario is possible in D3log.

@d-e-s-o, any chance you could have a look at this? Do you remember why SOCK_CLOEXEC was needed in the first place?


ryzhyk:
Thanks, @d-e-s-o!

PR #565, opened 2020-03-19
Fixed bigint to signed<N> conversion.

ryzhyk:
This code was untested and had multiple issues.  Most importantly, even
after truncating `bigint` to the right width, it cannot be safely
converted using `.to_sN`, as truncated negative number is interpreted as
a positive and may not fit in the `signed<N>` range, so we convert to
`bit<N>` first and then coerce to `signed<N>`.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -71,7 +71,7 @@ test-simple:
     tags:
         - ddlog-ci-1
     script:
-    - stack --no-terminal test --ta "-p simple"
+    - stack --no-terminal test --ta '-p "$(NF) == \"generate simple\" || ($(NF-1) == \"compiler tests\" && $(NF) == \"simple\")"'
     - cd test/datalog_tests/simple_ddlog/ && LIBRARY_PATH=/usr/lib/x86_64-linux-gnu/ cargo build --features='profile'
 
 test-simple2:
--- a/test/datalog_tests/simple2.dl
+++ b/test/datalog_tests/simple2.dl
@@ -1,3 +1,6 @@
+import fp
+import log
+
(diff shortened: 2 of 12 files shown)


test/datalog_tests/simple2.dl
```diff
@@ -42,3 +45,18 @@ output relation TArrng1Arrng2(x: bigint)
 TArrng1Arrng2(x.f2.f2) :-
     TArrng1[t],
     TArrng2[(x, t.0.f2.f2)].
+
+
+function ftoi_(l: double): signed<32> = {
```
mbudiu-vmw: you could also uncomment this in the souffle_lib.dl file

.gitlab-ci.yml
```diff
@@ -71,7 +71,7 @@ test-simple:
     tags:
         - ddlog-ci-1
     script:
-    - stack --no-terminal test --ta "-p simple"
+    - stack --no-terminal test --ta '-p "$(NF) == \"generate simple\" || ($(NF-1) == \"compiler tests\" && $(NF) == \"simple\")"'
```
mbudiu-vmw: this is very obscure

test/datalog_tests/simple2.dl
ryzhyk: Do we have/need tests for that?

mbudiu-vmw: Previously it wouldn't even compile (the generated Rust). So that's one test.
Second, if that works I may be able to enable at least one Souffle test that is using it (if everything else works too).

ryzhyk: Ok, uncommented `ftoi`, `ftou`. Will merge once tests pass.

PR #566, opened 2020-03-20
JSON: (De)serialization of Option<> types.

ryzhyk:
`Option<>` types should be handled in a special way during
serialization/deserialization.  In particular, when deserializing, a
missing field of type `Option<T>` should deserialize as `None`, as
opposed to returuning an error.  The `serde` crate already implements
this behavior for the Rust `Option<>` type, but this is not the same as
`Option` in DDlog, which actually compiles to `std_Option`.

The solution is to implement `From/Into` conversion between `Option` and
`std_Option` types and to annotate the DDlog `Option` declaration with:

```
 #[rust="serde(from=\"Option<A>\", into=\"Option<A>\", bound(serialize=\"A: Clone+Serialize\"))"]
 typedef Option<'A> = None
                    | Some{x: 'A}
```

The tricky part is the `A: Clone` trait bound, as `serde` requires any type
annotated by `serde(into)` to implement `Clone`.

One side effect of this patch is that we now require Rust 1.41+, before
which the following was considered an orphan implementation.

```
impl<T> From<std_Option<T>> for Option<T>
```

so we also update docker, nix config, and README to Rust 1.41.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 [![License: MIT](https://img.shields.io/badge/License-MIT-green.svg)](https://opensource.org/licenses/MIT)
 [![Build Status](https://travis-ci.com/vmware/differential-datalog.svg?branch=master)](https://travis-ci.com/vmware/differential-datalog)
 [![pipeline status](https://gitlab.com/ddlog/differential-datalog/badges/master/pipeline.svg)](https://gitlab.com/ddlog/differential-datalog/commits/master)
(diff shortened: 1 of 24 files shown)


ryzhyk:
> How does None turn into Jason null?

That's serde's default way to serialize `None`. I don't know why it doesn't just skip such fields, but I guess `null` works too?

mbudiu-vmw:
Hope it does

PR #568, opened 2020-03-23
Several new library functions.

ryzhyk:
See commit comments.

@mbudiu-vmw , there's a new souffle test failure :(

### DIFF
--- a/lib/json.dl
+++ b/lib/json.dl
@@ -34,6 +34,128 @@ typedef JsonValue = // Represents a JSON null value.
                   | // Represents a JSON object.
                     JsonObject{o: Map<string, JsonValue>}
 
+/* Get JsonValue value as `bool`.  Returns `None` if value is not `JsonBool`.
+ */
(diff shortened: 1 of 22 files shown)


lib/json.dl
```diff
@@ -34,6 +34,128 @@ typedef JsonValue = // Represents a JSON null value.
...
+{
+    match (v) {
+        JsonBool{b} -> Some{b},
+        _ -> None
+    }
+}
+
+/* Get JsonValue value as `bool`.  Returns `def` if value is not `JsonBool`.
+ */
+function jval_as_bool_or(v: JsonValue, def: bool): bool =
```
mbudiu-vmw: The name is misleading, sounds like boolean or. Should probably be orDefault

ryzhyk: `or_default` is reserved for functions that return default value of the return type on failure.
I can see that this is slightly confusing, but it is uniform with function naming in this and other libs (and also with the Rust standard library).

PR #576, opened 2020-03-25
lib/regex.dl: Two new functions.

ryzhyk:
Also added tests for this library in `test/datalog_tests/regex_test.dl`.

```
/*
 * Returns the leftmost-first match in `text`. If `regex` is not a valid regular
 * expression or no match exists, then `None` is returned.
 */
extern function regex_first_match(regex: string, text: string): Option<string>

/*
 * Returns all successive non-overlapping matches in `text`.
 * If `regex` is not a valid regular expression or no match exists,
 * an empty vector is returned.
 */
extern function regex_all_matches(regex: string, text: string): Vec<string>
```

### DIFF
--- a/lib/regex.dl
+++ b/lib/regex.dl
@@ -11,4 +11,17 @@
  * Returns false if `regex` is not a valid regular expression or if no match was
  * found.
  */
-extern function regex_match(regex: string, str: string): bool
+extern function regex_match(regex: string, text: string): bool
+
+/*
+ * Returns the leftmost-first match in `text`. If `regex` is not a valid regular
+ * expression or no match exists, then `None` is returned.
+ */
+extern function regex_first_match(regex: string, text: string): Option<string>
+
+/*
+ * Returns all successive non-overlapping matches in `text`.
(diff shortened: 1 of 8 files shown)


lib/regex.dl
```diff
@@ -11,4 +11,17 @@
  * Returns false if `regex` is not a valid regular expression or if no match was
  * found.
  */
-extern function regex_match(regex: string, str: string): bool
+extern function regex_match(regex: string, text: string): bool
```
mbudiu-vmw: I wonder whether you may want to have an extern type for regex, which is a compiled regular expression.
If you want to use a regex in a rule it would be nice to be able to compile it.

ryzhyk: The current API is a convenience API that combines regex compilation and matching. It works nicely for our existing use cases when we are porting, e.g., Groovy code like `str =~ /^[\d\.v]+/`.

I agree that a complete regex library would have the API that separates compilation from matching.

ryzhyk: (and we should build this library at some point)

PR #577, opened 2020-03-26
Antrea: Don't aggregate outputs.

ryzhyk:
We discovered that the Antrea controller ends up spending most of the
time, (99.5% of the time for some benchmarks) aggregating node and pod
ids when forming output records.   We eliminate the overheads by
avoiding the aggregation and outputing individual components of these
records instead.

@antoninbas , as you can see the set of output relations is completely different now.
You can experiment with this branch and approve the PR once/if it works for you.


ryzhyk:
Shall I keep this PR open until you've done the integration?

antoninbas:
We can keep it open for now, I'll get back to you later today to share status

PR #578, opened 2020-03-27
std.dl: string_to_uppercase(), string_reverse().

ryzhyk:
Two new std library functions:

```
extern function string_to_uppercase(s: string): string
extern function string_reverse(s: string): string
```

### DIFF
--- a/lib/std.dl
+++ b/lib/std.dl
@@ -159,6 +159,12 @@ extern function string_trim(s: string): string
 extern function string_to_lowercase(s: string): string
 extern function str_to_lower(s: string): string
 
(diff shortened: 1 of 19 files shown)


mbudiu-vmw:
Maybe it's time for a separate string library. But not in this PR.

ryzhyk:
> Maybe it's time for a separate string library. But not in this PR.

It's unfortunately too late for that unless we want to sacrifice backwards compatibility.  Strings, vectors, maps, refs, and Option's live in `std.dl`. We don't have a re-export feature to transparently move them to their own modules.

All new types live in separate libraries.

PR #581, opened 2020-03-28
Cleanup stack test

mbudiu-vmw:


mbudiu-vmw:
@blp : if we merge this first, then I expect your PR should no longer fail

test/Spec.hs
```diff
@@ -168,18 +128,19 @@ parseValidate file java program = do
...
 -- compile a program that is supposed to fail compilation
-compileFailingProgram :: String -> String -> IO String
-compileFailingProgram file program =
-    (show <$> parseValidate file False program) `catch`
-             (\e -> return $ show (e::SomeException))
+compileFailingProgram :: String -> String -> IO()
+compileFailingProgram file program = do
+   prog <- (show <$> parseValidate file False program) `catch`
+               (\e -> let _ = (e :: SomeException) in return "")
+   assertBool "Compilation should have failed" ("" == prog)
```
ryzhyk: Are we going too far here? For programs that fail compilation ".ast" files actually contain error messages, which we want to check.

java/ddlogapi/DDlogAPI.java
```diff
@@ -468,4 +470,134 @@ static public synchronized void logSetDefaultCallback(ObjIntConsumer<String> cb,
...
+            System.err.println("Error running " + String.join(" ", commands));
+            System.err.println(ex.getMessage());
+            return 1;
+        }
+    }
+
+    public static String libName(String lib) {
+        String os = System.getProperty("os.name").toLowerCase();
+        if (os.equals("darwin"))
+            return "lib" + lib + ".dynlib";
```
lalithsuresh: This line should be updated, following https://github.com/vmware/differential-datalog/pull/587#event-3179753749

1. -> os.equals("mac os x")
2. -> "dylib", not "dynlib"

mbudiu-vmw:
@ryzhyk : I think that the CI will pass if we merge this; it failed because of the previous PR #587, but that is now fixed.

PR #587, opened 2020-03-30
Simplify Translator.compileAndLoad()

lalithsuresh:
The earlier version used a Classloader trick to edit the `java.library.path` variable.
This is not portable and does work across different types of JDKs (it failed on OpenJDK 12).
This fix instead invokes System.load() against the absolute path of the created library.

Invoking System.load() within compileAndLoad() is redundant with the System.loadLibrary()
attempt that is attempted in the static constructor within the DDlogAPI() class.
This works out fine if java.library.path includes the current directory.

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/translator/Translator.java
+++ b/sql/src/main/java/com/vmware/ddlog/translator/Translator.java
@@ -13,9 +13,12 @@ package com.vmware.ddlog.translator;
 
 import com.facebook.presto.sql.parser.ParsingOptions;
 import com.facebook.presto.sql.parser.SqlParser;
-import com.facebook.presto.sql.tree.*;
-import com.vmware.ddlog.ir.*;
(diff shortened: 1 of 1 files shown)


lalithsuresh:
@mbudiu-vmw sounds good. Left a comment on #581. Can we merge this PR?

ryzhyk:
@mbudiu-vmw 
This PR broke master.

mbudiu-vmw:
the tests seemed to have passed.

ryzhyk:
Gitlab tests don't run on @lalithsuresh 's commits unless he starts mirroring his repo to the gitlab org.

lalithsuresh:
I see what the bug is. Seems like the default for `java.library.path` does not include "." in your CI infrastructure. And `DDlogApi` assumes the library is available within one of the paths indicated by `java.library.path`. I've sent a fix in #589, which explicitly sets the path variable for the tests.

PR #590, opened 2020-03-30
Explicitly set java.library.path to working-dir for running DynamicTe…

lalithsuresh:
…st. Else, library loading fails.


mbudiu-vmw:
Let's hope this fixes the build.

PR #599, opened 2020-04-02
Speed up compilation by reusing the same directory

mbudiu-vmw:


mbudiu-vmw:
@lalithsuresh : This will also avoid rust-compiling most tests, so the date errors should disappear.

Issue #603, opened 2020-04-03
Surprising parsing ambiguity between expressions and rules


blp:
This parses OK:

> function myfunc(x: string): string = { x }
> relation &X(x: string)
> &X("foo").

It fails if we swap lines 1 and 2:

> relation &X(x: string)
> function myfunc(x: string): string = { x }
> &X("foo").

with:
> ddlog: Failed to parse input file: "playpen3.dl" (line 3, column 3):
> unexpected '('
> expecting "#", "import", "typedef", "extern", "input", "output", "relation", "index", "function", "transformer", variable name, "&", relation name, "apply", "for" or end of input

blp:
Fixing issue #606 would solve this, if terminators were mandatory.

Another way to solve it would be to make {} a mandatory part of the syntax for functions.

ryzhyk:
Closed via #610

PR #610, opened 2020-04-04
New syntax for functions.

ryzhyk:
(Addresses #603)

This code

```
relation &X(x: string)
function myfunc(x: string): string = { x }
&X("foo").
```

fails to compile with:

```
ddlog: Failed to parse input file: "playpen3.dl" (line 3, column 3):
unexpected '('
expecting "#", "import", "typedef", "extern", "input", "output", "relation", "index", "function", "transformer", variable name, "&", relation name, "apply", "for" or end of input
```

The problem is that the syntax with optional curly braces is
inherently ambiguous.  In addition, the `=` between function
declaration and body is bad aesthetically in most cases.

We introduce new, more conventional, syntax without `=` and with
mandatory braces.  It eliminates the ambiguity and looks better imo.

We still support the old syntax to avoid breaking people's code.  We may
phase it out in the future, but I don't see much harm in it.

### DIFF
--- a/doc/language_reference/language_reference.md
+++ b/doc/language_reference/language_reference.md
@@ -170,7 +170,7 @@ identical names, their types must be identical, e.g., the following is invalid:
 arguments matching its declaration:
     ```
     typedef type1<'A,'B>
-    function f(): bool = {
+    function f(): bool {
         var x: type1<bigint> // error: not enough type arguments
     }
     ```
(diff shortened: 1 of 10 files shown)


doc/language_reference/language_reference.md
```diff
@@ -217,7 +217,7 @@ body.
 ```EBNF
 function ::= "function" func_name "(" [arg(,arg)*]")"
               ":" simple_type_spec (* return type *)
-              ["=" expr]
+              ("{" expr "} | "=" expr) (* body of the function *)
```
mbudiu-vmw, line 220: Just take it out of the doc if you want to deprecate it

mbudiu-vmw:
I will fix the SQL libs in another pr

PR #613, opened 2020-04-06
api.rs: ddlog_transaction_commit_dump_changes_as_array (untested).

ryzhyk:
When DDlog generates many updates, it can be expensive to send them all
via FFI to Java or Go.  The new method returns all updates generated by
a transaction in a single array.

@antoninbas , can you let me know if the new `ddlog_transaction_commit_dump_changes_as_array`/`ddlog_free_record_updates` API works for you (in terms of functionality and in terms of feasibility of creating Go bindings)?

### DIFF
--- a/rust/template/ddlog.h
+++ b/rust/template/ddlog.h
@@ -51,6 +51,11 @@ typedef size_t table_id;
  */
 typedef void ddlog_record;
 
+typedef struct {
+    table_id table;
+    ddlog_record *rec;
+    bool polarity;
+} ddlog_record_update;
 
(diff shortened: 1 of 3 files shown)


ryzhyk:
> I don't see how else to do this.
> I don't see a test; should I create one in Java using JNI?

That would be great (plus it would be good to export this API to Java anywat) if it's not too much work.  This API returns an array of structs.

antoninbas:
@ryzhyk I think I should be able to leverage this in the Go bindings

PR #615, opened 2020-04-07
Add Go bindings for ddlog_transaction_commit_dump_changes_as_array

antoninbas:
The new API is called CommitTransactionChangesAsArray.

Unfortunately we do not see a noticable improvement in benchmarks:
BenchmarkTransaction-4                 	     339	   3867018 ns/op
BenchmarkTransactionChangesAsArray-4   	     350	   3448197 ns/op
BenchmarkRecord-4                      	  764809	      1499 ns/op

^ this is one of the better results

It makes sense, since at least for this benchmark it seems that the
bottleneck is the actual processing of the transaction by DDlog:

      flat  flat%   sum%        cum   cum%
        2s 71.17% 71.17%         2s 71.17%  [libtypesTest_ddlog.so]
     0.47s 16.73% 87.90%      0.47s 16.73%  [libc-2.27.so]
     0.15s  5.34% 93.24%      0.15s  5.34%  [libpthread-2.27.so]
     0.11s  3.91% 97.15%      0.11s  3.91%  runtime.cgocall

A real life example (Antrea) did not seem to benefit from this change
either. Even though Antrea shows more overhead caused by
runtime.cgocall (10%), most of this overhead seems to be caused by
creating / consuming records using the record API. Still, if we ever
have a more efficient way of creating / consuming records from Go, the
relative improvement offered by CommitTransactionChangesAsArray may
become more significant.

### DIFF
--- a/go/pkg/ddlog/ddlog.go
+++ b/go/pkg/ddlog/ddlog.go
@@ -32,6 +32,18 @@ static int ddlogTransactionCommitDumpChanges(ddlog_prog hprog, uintptr_t arg) {
     return ddlog_transaction_commit_dump_changes(hprog, dumpChangesCb, arg);
 }
 
+extern void handleOutRecordArray(uintptr_t progIdx, ddlog_record_update *changes, size_t num_changes);
+
+static int ddlogTransactionCommitDumpChangesAsArray(ddlog_prog hprog, uintptr_t progIdx) {
+    ddlog_record_update *changes;
+    size_t num_changes;
+    int rc = ddlog_transaction_commit_dump_changes_as_array(hprog, &changes, &num_changes);
+    if (rc != 0 || num_changes == 0) return rc;
+    handleOutRecordArray(progIdx, changes, num_changes);
+    ddlog_free_record_updates(changes, num_changes);
+    return 0;
+}
+
 typedef void (*print_err_msg_fn)(const char *msg);
 
 typedef const char *cstring;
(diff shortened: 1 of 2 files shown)


go/pkg/ddlog/ddlog.go
```diff
@@ -32,6 +32,18 @@ static int ddlogTransactionCommitDumpChanges(ddlog_prog hprog, uintptr_t arg) {
     return ddlog_transaction_commit_dump_changes(hprog, dumpChangesCb, arg);
 }
 
+extern void handleOutRecordArray(uintptr_t progIdx, ddlog_record_update *changes, size_t num_changes);
+
+static int ddlogTransactionCommitDumpChangesAsArray(ddlog_prog hprog, uintptr_t progIdx) {
+    ddlog_record_update *changes;
+    size_t num_changes;
+    int rc = ddlog_transaction_commit_dump_changes_as_array(hprog, &changes, &num_changes);
+    if (rc != 0 || num_changes == 0) return rc;
```
antoninbas, line 41: @ryzhyk I am assuming that there is no need to call `ddlog_free_record_updates` in that case

go/pkg/ddlog/ddlog.go
```diff
@@ -302,6 +314,19 @@ func (p *Program) CommitTransaction() error {
 	return nil
 }
 
+// CommitTransactionChangesAsArray commits a transaction. It uses a different implementation from
+// CommitTransaction, which may yield better performance when many output records are
+// generated. Unlike with CommitTransaction, DDlog will not generate one callback for each output
+// record, but will return an array of output records (with polarity). Note that we still generate
+// one OutRecordHandler callback for each output record.
```
antoninbas, line 321: for the record, I tried batching callbacks as this level as well and I didn't observe any improvement

go/pkg/ddlog/ddlog.go
ryzhyk: correct

PR #622, opened 2020-04-14
Support for time datatype

mbudiu-vmw:
If this is good I will do Date in a similar way.


lib/time.dl
```diff
@@ -0,0 +1,15 @@
...
+extern function try_from_hms_nano(hours: bit<8>, minutes: bit<8>, seconds: bit<8>, microseconds: bit<32>): Result<time, string>
+extern function midnight(): time
+extern function hour(t: time): bit<8>
+extern function minute(t: time): bit<8>
+extern function second(t: time): bit<8>
+extern function millisecond(t: time): bit<16>
+extern function microsecond(t: time): bit<32>
+extern function nanosecond(t: time): bit<32>
+extern function time2string(t: time): string
+extern function parse(s: string, format: string): Result<time, string>
```
ryzhyk, line 15: Can you add examples of supported input formats?

lib/time.dl
```diff
@@ -0,0 +1,15 @@
...
+extern function try_from_hms_micro(hours: bit<8>, minutes: bit<8>, seconds: bit<8>, microseconds: bit<32>): Result<time, string>
+extern function try_from_hms_nano(hours: bit<8>, minutes: bit<8>, seconds: bit<8>, microseconds: bit<32>): Result<time, string>
+extern function midnight(): time
+extern function hour(t: time): bit<8>
+extern function minute(t: time): bit<8>
+extern function second(t: time): bit<8>
+extern function millisecond(t: time): bit<16>
+extern function microsecond(t: time): bit<32>
+extern function nanosecond(t: time): bit<32>
+extern function time2string(t: time): string
```
ryzhyk, line 14: Can we also have a version of `2string` that takes format string?

lib/time.dl
```diff
@@ -0,0 +1,15 @@
+extern type time
+
+extern function try_from_hms(hours: bit<8>, minutes: bit<8>, seconds: bit<8>): Result<time, string>
```
ryzhyk, line 3: I would prefix all these functions with `time_`. Alternatively, users can import this module using qualified name `import time as time`, but then what happens when you start adding support for dates and date-times?

test/datalog_tests/time_test.dat
```diff
@@ -0,0 +1,4 @@
+start;
```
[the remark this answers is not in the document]
mbudiu-vmw: Flatbuf is not supported. It's not clear what needs to be changed to support extern types in flatbuf.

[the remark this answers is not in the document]
ryzhyk: CLI does not use flatbuf.

lib/time.dl
mbudiu-vmw: I've done everything but add this prefix.

rust/template/src/main.rs
```diff
@@ -95,7 +94,7 @@ fn handle_cmd(
         Command::Comment => Ok(()),
         Command::Rollback => hddlog.transaction_rollback(),
         Command::Timestamp => {
-            println!("Timestamp: {}", precise_time_ns());
+            println!("Timestamp: {:#?}", std::time::Instant::now());
```
ryzhyk, line 97: why did you have to change this. The new method will produce output that does not look good, e.g., `Instant { tv_sec: 5122222, tv_nsec: 893024142 }`

PR #623, opened 2020-04-14
std Library function + flatbuf bug fixes.

ryzhyk:
See commit messages.


ryzhyk:
Mergin, as we need to release this.

Issue #633, opened 2020-04-17
RFC: Static evaluation.


ryzhyk:
This feature was proposed by @mbudiu-vmw 

Rationale
--------

We would like to be able to evaluate constant expressions, including complex ones containing function calls, once, either at the start of the program or even during compilation.  This RFC addresses the former.  We can also add compile-time evaluation as a further optimization later, but it will be applicable in restricted contexts, as, e.g., calls to extern functions are difficult to evaluate in the compiler.

This feature is a prerequisite for several other useful features, including efficient interning of string literals and support for constants.

Design
---------

- The first question is: what expressions should be evaluated statically? For instance, given an expression `f1(C1) + f2(C2)`, the options are:
  1. `f1(C1)` and `f2(C2)` are evaluated statically; the sum is evaluated dynamically
  2. `f1(C1) + f2(C2)` is evaluated statically
  3. `f1(C1)`, `f2(C2)`, `f1(C1)+f2(C2)` are all evaluated statically, with the sum expression being built out of statically computed results for `f1` and `f2`.

  Given that accessing a static expression is not free (see below) I am inclined to go with option 1, i.e., statically pre-evaluate function calls on constant arguments.

  How about `f1(f2(C))`? If we go with the policy of pre-computing function calls, then `f2(C)` and `f1(f2(C))` are both statics, with the latte referencing the former. There is a potential caveat: rust's lazy static evaluation can deadlock when evaluating mutually dependent statics, but as far as I can tell this can only occur in situations that would cause infinite recursion otherwise.

- Detecting constant sub-expressions is relatively straightforward.  The only caveat is that some extern functions, most notably logging functions, can have side effects and should not be evaluated statically. We can handle this by introducing a new annotation, e.g., `#[has_side_effects]`.  We could also take the whitelisting approach and annotate `#[pure]` extern functions, but I expect that library writers will tend to forget to add this annotation.

- Constant subexpressions will be compiled to Rust lazy statics, which is the only Rust construct (a library, actually) that allows statically evaluating complex expressions with function calls. This will introduce a bit of overhead on each access (checking an atomic variable).

ryzhyk:
This RFC was implemented in #636

PR #636, opened 2020-04-19
Static evaluation of constant expressions.

ryzhyk:
Implements RFC #633

This feature was proposed by @mbudiu-vmw

Rationale
--------

We would like to evaluate constant expressions, including complex ones
containing function calls, once, either at the start of the program or
even during compilation.  This commit addresses the former.  We may add
compile-time evaluation as a further optimization later, but it will be
applicable in restricted contexts, as, e.g., calls to extern functions
are difficult to evaluate in the compiler.

This feature is a prerequisite for several other useful features,
including efficient interning of string literals and support for
constants.

Design
---------

- The first question is: what expressions should be evaluated
  statically? For instance, given an expression `f1(C1) + f2(C2)`, the
  options are: 1. `f1(C1)` and `f2(C2)` are evaluated statically; the sum
  is evaluated dynamically 2. `f1(C1) + f2(C2)` is evaluated statically 3.
  `f1(C1)`, `f2(C2)`, `f1(C1)+f2(C2)` are all evaluated statically, with
  the sum expression being built out of statically computed results for
  `f1` and `f2`.

  Given that accessing a static expression is not free (see below), we
  go with option 1, i.e., statically pre-evaluate function
  calls on constant arguments.

  How about `f1(f2(C))`?  In this example, `f2(C)` and `f1(f2(C))` are both
  statics, with the latter referencing the former. There is a potential caveat:
  rust's lazy static evaluation can deadlock when evaluating mutually dependent
  statics, but as far as I can tell this can only occur in situations that would
  cause infinite recursion otherwise.

- Detecting constant sub-expressions is relatively straightforward.  The only
  caveat is that some extern functions, most notably logging functions, can have
  side effects and should not be evaluated statically. We handle this by introducing
  a new annotation: `#[has_side_effects]`.  Functions with this annotation as well
  as any functions that call them are not evaluated statically.

- Constant subexpressions are compiled to Rust lazy statics, which is the only Rust
  construct (a library, actually) that allows statically evaluating complex expressions
  with function calls. This introduces a bit of overhead on each access (checking an
  atomic variable).

### DIFF
--- a/doc/tutorial/tutorial.md
+++ b/doc/tutorial/tutorial.md
@@ -1,6 +1,6 @@
 # A Differential Datalog (DDlog) tutorial
 
-**Note:**: All examples from this tutorial can be found in
+**Note:** All examples from this tutorial can be found in
 [`test/datalog_tests/tutorial.dl`](../../test/datalog_tests/tutorial.dl).
 The examples can be executed using test inputs from
 [`test/datalog_tests/tutorial.dat`](../../test/datalog_tests/tutorial.dat)
@@ -37,6 +37,8 @@ program -> | compiler |-> code -> | compiler |--> executable
 ## Installing DDlog
 
 Installation instructions are found in the [README](../../README.md#Installation).
+Don't forget to set the `$DDLOG_HOME` environment variable to point to the root of the repository
+or the DDlog installation directory (when using a binary release of DDlog).
 
 ## "Hello, World!" in DDlog
 
@@ -137,10 +139,12 @@ should use the order that makes programs easiest to read.
 Compiling a DDlog program consists of two steps.
 First, run the DDlog compiler to generate the Rust program:
 ```
-ddlog -i playpen.dl -L <ddlog/lib>
+ddlog -i playpen.dl
 ```
-where `<ddlog/lib>` is the path to the `lib` directory inside the
-DDlog installation directory (when using a binary release) or
+DDlog will search for its standard libraries inside the directory specified
+in the `$DDLOG_HOME` variable.  If this variable is not set, add the `-L <ddlog/lib>`
+switch, where `<ddlog/lib>` is the path to the `lib`
+directory inside the DDlog installation directory (when using a binary release) or
 inside the DDlog repository (when building DDlog from source).
 
 Second, run the Rust compiler to generate a binary executable:
@@ -286,7 +290,7 @@ Phrases{"I am your father"}
 > ```
 > #!/bin/bash
 >
-> ddlog -i playpen.dl -L <ddlog/lib>
+> ddlog -i playpen.dl
 > (cd playpen_ddlog && cargo build --release)
 > ./playpen_ddlog/target/release/playpen_cli < playpen.dat
 > ```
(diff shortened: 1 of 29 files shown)


doc/tutorial/tutorial.md
```diff
@@ -1706,6 +1706,33 @@ to optimize data structures layout:
 extern type IObj<'A>
 ```
 
+### `#[has_side_effects]`
```
mbudiu-vmw, line 1709: These are not really tutorial items, they are highly technical.
The tutorial should only describe the core language, and should be ideally short and easy to read.

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -547,7 +544,51 @@ compileLib d specname rs_code = (typesLib, valueLib, mainLib)
     typedefs = vcat $ map (mkTypedef d) $ M.elems $ progTypedefs d
     -- Functions
     (fdef, fextern) = partition (isJust . funcDef) $ M.elems $ progFunctions d
-    funcs = vcat $ (map (mkFunc d) fextern ++ map (mkFunc d) fdef)
+    funcs = let ?statics = statics in vcat $ (map (mkFunc d) fextern ++ map (mkFunc d) fdef)
+
+-- This type stores the set of statically evaluated constant sub-expressions.
+-- We need to track the type of each expression, as generic functions like
+-- 'set_empty()' can have different types in different contexts.
```
mbudiu-vmw, line 551: One solution would be to have an earlier pass which annotates all polymorphic functions with their type.

src/Language/DifferentialDatalog/Module.hs
```diff
@@ -64,28 +64,31 @@ data DatalogModule = DatalogModule {
     moduleDefs :: DatalogProgram
 }
 
--- standard library module name
-stdname :: ModuleName
-stdname = ModuleName ["std"]
+-- Standard library module name.
+stdLibs :: [ModuleName]
+stdLibs = [ModuleName ["std"], ModuleName ["internment"]]
```
mbudiu-vmw, line 69: this should probably be documented too, because users need to know this.

src/Language/DifferentialDatalog/Syntax.hs
```diff
@@ -656,6 +656,72 @@ instance Eq e => Eq (ExprNode e) where
     (==) (ERef _ p1)              (ERef _ p2)                = p1 == p2
     (==) _                        _                          = False
 
+-- Assign rank to constructors; used in the implementation of Ord.
```
mbudiu-vmw, line 659: Why didn't we need to do this previously?

test/datalog_tests/internment_test.dl
```diff
@@ -1,12 +1,22 @@
-import internment
+input relation IInternedString(ix: istring)
+relation StaticInternedString(ix: istring)
 
-input relation IInternedString(ix: IString)
+StaticInternedString(intern("static foo")).
+StaticInternedString(i"ifoo").
```
mbudiu-vmw, line 5: I don't recall seeing the `i""` syntax in the documentation.

test/datalog_tests/simple2.dl
```diff
@@ -95,3 +95,14 @@ SumsOfDoubles(x, y, z) :-
     Doubles(ys),
     var y = FlatMap(ys),
     var z = x + y.
+
```
mbudiu-vmw, line 98: no tests with other constant expressions?

doc/tutorial/tutorial.md
ryzhyk: I think there's nothing wrong about having advanced topics covered in the tutorial. Maybe we should break it up into several chapters for easier navigation and to have all the basic stuff in one smaller document.

src/Language/DifferentialDatalog/Module.hs
ryzhyk: Documentation is on its way...

src/Language/DifferentialDatalog/Syntax.hs
ryzhyk: We never used expressions as keys in a map and so never had to implement `Ord` for them.

ryzhyk:
Pushed documentation update + a test for non-string interned object. Will merge once tests pass.

PR #638, opened 2020-04-21
Improved error messages

mbudiu-vmw:
This is how an improved error message looks:

```
error: ./test/datalog_tests/strings.fail.dl:6:24-6:28:Unknown variable: str1
    var err = "${str2++str1}"; // str1 is not declared in this scope
                       ^^^^
```

I have not yet converted all error messages, but if this is an improvement I will work to improve more of them. 

The Error.hs module has moved some functions from Util.hs to avoid circular module dependencies. The main reason I have not converted all errors is that not all functions that report errors receive the program as an argument, and I am stuffing the source code in the program object.


src/Language/DifferentialDatalog/Error.hs
```diff
@@ -0,0 +1,58 @@
...
+import Language.DifferentialDatalog.Name
+import Language.DifferentialDatalog.Syntax
+
+errBrief :: (MonadError String me) => Pos -> String -> me a
+errBrief p e = throwError $ spos p ++ ": " ++ e
+
+err :: (MonadError String me) => Pos -> DatalogProgram -> String -> me a
+err p d e = throwError $ sposFragment p (progSources d) e
+
+check :: (MonadError String me) => Bool -> Pos -> String -> me ()
```
ryzhyk, line 40: I guess the plan is to have versions of check and uniq that take DatalogProgram too?

src/Language/DifferentialDatalog/Pos.hs
```diff
@@ -50,7 +52,24 @@ spos x = let (s,e) = pos x
          in sourceName s ++ ":" ++ (show $ sourceLine s) ++ ":" ++ (show $ sourceColumn s) ++ "-"
                                 ++ (show $ sourceLine e) ++ ":" ++ (show $ sourceColumn e)
 
-nopos::Pos 
+sposFragment :: (WithPos a) => a -> M.Map String String -> String -> String
```
ryzhyk, line 55: A comment here would be good, perhaps even including an example output produced by this function.

src/Language/DifferentialDatalog/Pos.hs
```diff
@@ -50,7 +52,24 @@ spos x = let (s,e) = pos x
...
+sposFragment element sources message =
+   let (s,e)      = pos element
+       sourceFile = sourceName s
+       lineStart  = sourceLine s
+       colStart   = sourceColumn s
+       lineEnd    = sourceLine e
+       colEnd     = sourceColumn e
+       sameLine   = lineStart == lineEnd
+       width      = if sameLine then colEnd - colStart else 1
+       line       = (lines $ (sources M.! sourceFile)) !! (lineStart - 1)
```
ryzhyk, line 65: This will crash if the start line is not valid, which is very much possible if there is a bug in parser or if we use artificially manufactured pos (which we do in some places). Ideally this should not happen, but the program should not crash if it does.

Also, there is a special case when position  `==nopos`, which means that we have an object whose real position is unknown or undefined (e.g., it was injected during program transformation). We should not theoretically have errors in those, but we should not crash if this happens.

mbudiu-vmw:
A few error messages can still be improved, in particular this only shows the code fragment for one element, but the duplicate errors should show all of them.

Issue #637, opened 2020-04-21
Error parsing dumped ast


mbudiu-vmw:
While running `stack test` the following error occurs:

```
    ./test/datalog_tests/lib_test.dl:        FAIL
      Exception: Failed to parse input file: "./test/datalog_tests/lib_test.ast" (line 953, column 1):
      unexpected "t"
      expecting "index"
```

The weird thing is that line 953 column 1 does not have any "t"; I suspect this is yet another instance of a parser ambiguity.

PR #639, opened 2020-04-23
Parse.hs: two parser fixes.

ryzhyk:
- `in` is declared as a reserved word, but was used as `reservedOp` in
  two places.
- The second fix addresses #637:
  We use the `symbol` lexeme parser instead of `reserved` to reduce the
  number of reserved words that cannot be used as variable names.  This
  is safe as long as it doies not introduce ambiguity.  What I did not
  realize is that `symbol` is not atomic, i.e., can fail after consuming
  some of the input, e.g., when parsing `internment...`, the `symbol
  "intern"` parser first kicks in, but fails after the first two
  characters.  The fix is to always wrap `symbol` in `try`:
  ```
  symbol       = try . T.symbol lexer
  ```

### DIFF
--- a/src/Language/DifferentialDatalog/Parse.hs
+++ b/src/Language/DifferentialDatalog/Parse.hs
@@ -130,7 +130,7 @@ semiSep      = T.semiSep lexer
 colon        = T.colon lexer
 commaSep     = T.commaSep lexer
 commaSep1    = T.commaSep1 lexer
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
Can I merge this?

ryzhyk:
> Can I merge this?

Yes, thanks, Travis tests failed yesterday because stackage was dead.

Issue #640, opened 2020-04-24
Building on Windows


d4hines:
Hello!

I'd like to start the conversation of building on Windows. WSL is fantastic, and I recommend WSL as the primary build tool for generating the Rust lib on Windows; however, my use case requires compiling to native Windows binaries, so WSL won't work for compiling the lib.

Fortunately, you're not too far off. I was able to successfully compile and run a DDLog lib by deleting a few things from my (local, autogenerated) template. I'll make a PR with those changes so you can see them, and we can discuss what's need for an actual, working PR.

ryzhyk:
Implemented in #645.

PR #641, opened 2020-04-24
compile on windows

d4hines:
This PR is NOT a real PR!

Per #640, I just want you guys to be able to see the changes I needed to make to get DDLog to compile and run natively on Windows.


mbudiu-vmw:
Ideally the pieces you deleted would only be conditionally compiled.
For the CLI it shouldn't be too hard to write a portable version; that piece is important.
But @ryzhyk should really review this.

ryzhyk:
It seems that the three pieces that don't work on Windows are:
1. OVSDB bindings
2. libtool stuff
3. CLI

The first two can be excluded using Rust's and cargo's platform-specific compilation. I agree with Mihai, we need to fix the CLI to be platform-independent.

ryzhyk:
Oh, and there's also some Unix stuff in api.rs.

ryzhyk:
I'll start making those changes.

PR #645, opened 2020-04-26
Partial Windows support.

ryzhyk:
Addresses #640.
@d4hines, would you be able to confirm that this works for you?

This patch enables compilation of DDlog-generated Rust code on Windows.
The generated code is mostly portable with a few exceptions:
- libtool support in `build.rs`.  We disable this on Windows.
- `api.rs`: two API calls work take raw file descriptors.  On Windows,
  these must be converted to file handles.
- `cmd_parser/lib.rs`: we used a constant declaed in `libc` for STDIN
  file descriptor.  We use liteal `0` instead.

There is more non-portable code in distributed-ddlog and ovsdb, but this
is not a problem for clients who don't use these features.

### DIFF
--- a/rust/template/Cargo.toml
+++ b/rust/template/Cargo.toml
@@ -15,7 +15,7 @@ members = [
   "value"
 ]
 
-[build-dependencies]
+[target.'cfg(not(windows))'.build-dependencies]
 libtool = "0.1"
 
 [dependencies.differential_datalog]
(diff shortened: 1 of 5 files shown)


d4hines:
Is there a binary for me test, or do I need to build locally?

d4hines:
Never mind - I got it building. 

Works for me! I tested it against the program you wrote and that I incorporated into my demo. It all compiles and runs fine. Great work!

ryzhyk:
Thanks for validating! I will merge.

PR #650, opened 2020-04-30
Support for FP in souffle

mbudiu-vmw:
Fixes to enable 4 more Souffle tests to pass.
One of them (math) fails in fact because there are no reference outputs in the source directory, so it is still disabled, but compilation succeeds.


mbudiu-vmw:
The pipeline failed because the way we generate variables names has been changed - the Souffle grammar seems to allow ? anywhere in a variable name, whereas we were only allowing it in front. Now we substitute ? with _ (which is not 100% safe, but could be good enough). So the names of variables in an output relation have changed. I have committed a new reference output for souffle0.

mbudiu-vmw:
At least this test seems to have passed, including the CI.

Issue #618, opened 2020-04-12
RFC: Improved logging framework


ryzhyk:
Rationale
-----------

Improve logging framework to support emitting whether the log is part of an insert or a delete and the logical timestamp when it was generated.

Current logging mechanism, implemented as a library, emits the same log regardless of the operation, since most differential operators do not "know" the context in which they get invoked. The solution is to embed logging operations in the [inspect](https://docs.rs/differential-dataflow/0.11.0/differential_dataflow/collection/struct.Collection.html#method.inspect) operator.  This operator behaves as a pass-through filter (and thus can be embedded anywhere in the dataflow) and contains an action that can produce a side effect, such as generating a log message.  It takes three arguments: (1) an input record (in case of DDlog, this will contain all variables visible at this point in the rule), (2) weight (>0 = insert, <0 = delete), (3) logical timestamp (in DDlog, this currently corresponds to the number of updates pushed to the program, but will be changed to correspond to the number of transactions committed so far; inside a recursive fragment of the program, the timestamp is a tuple consisting of the global timestamp and iteration number inside the recursive fragment). By wrapping logging statements in an `inspect`, we will be able to add this useful information to the log.

Caveats
-----------
- Differential can generate multiple insertions and deletions for the same timestamp, which may still cause confusing log messages. The number of updates can be reduced by applying the [consolidate](https://docs.rs/differential-dataflow/0.11.0/differential_dataflow/operators/consolidate/index.html) operator before inspect; however, `consolidate` can have a non-trivial cost; in addition, even though it does not change the semantics of the program, it does change its internal behavior, which may be undesirable during debugging.

Syntax
---------

```
R(x, z) :- 
   R1(x, y),
   R2(y, z),
   Inspect debug("Evaluating R: ts=${timestamp}, weight=${weight}, x=${x}, y=${y}, z=${z}").
```
Here, `Inspect` is a new DDlog keyword that maps into DD's `inspect` operator.  The keyword is followed by an arbitrary expression whose type must be `()`.  The expression can access all program variables visible at the given point in the rule, plus two special variables: `ddlog_weight: std.DDWeight` and `ddlog_timestamp: std.DDTimestamp`, where:

```
typedef DDWeight = s64
typedef DDTimestamp = DDTSGlobal{epoch: u64}
                    | DDTSNested{epoch: u64, iteration: u64}
```

[...]

PR #653, opened 2020-05-04
Add support for Inspect operator

haroldlim:
A new Inspect operator is introduced. Refer to
https://github.com/vmware/differential-datalog/issues/618.

Parsing, syntax, and validation has been updated to support this
new operator.

Update compileRule logic to temporarily skip Inspect rule.
A placeholder mkInspect function is added that currently only
generates an empty string..


src/Language/DifferentialDatalog/NS.hs
```diff
@@ -147,6 +147,7 @@ ctxMVars d ctx =
          CtxRuleRAtom rl i        -> ([], map f2mf $ ruleRHSVars d rl i)
          CtxRuleRCond rl i        -> ([], map f2mf $ ruleRHSVars d rl i)
          CtxRuleRFlatMap rl i     -> ([], map f2mf $ ruleRHSVars d rl i)
+         CtxRuleRInspect rl i     -> ([], (map f2mf $ ruleRHSVars d rl i) ++ [("ddlog_weight", Just $ tBit 64), ("ddlog_timestamp", Just $ tBit 64)])
```
ryzhyk, line 150: The timestamp type is actually more complex.  For a non-recursive rule, the timestamp is indeed just u64. For a rule that is part of a recursive fragment, it is a pair of two timestamps: the global timestamp, which is the same as in the non-recursive case and the inner timestamp, which represents iteration of the fixed point computation.  There are two ways we can handle this. The RFC suggests:

```
typedef DDTimestamp = DDTSGlobal{epoch: u64}
                    | DDTSNested{epoch: u64, iteration: u64}
```

Now I am thinking that a better way would be to use different types depending on the context.  We can add a method to check if the rule is recursive and use the appropriate type based on that.

test/datalog_tests/simple2.dl
```diff
@@ -106,3 +106,15 @@ function weird_zero(x: 'A): usize {
...
+
+/* See #618. New Inspect operator syntax. */
+output relation InspectSimpleSum(x: bit<32>, total: bit<32>)
+
+input relation InputTuples(x: bit<32>, y: bit<32>)
+
+InspectSimpleSum(x, total) :-
+    InputTuples(x, y),
+    var total = Aggregate((x), group_sum(y)),
+    Inspect {
```
ryzhyk, line 118: You may want to also test the use of timestamp and weight variables.

lib/std.dl
```diff
@@ -19,6 +19,12 @@ typedef s128 = signed<128>
 
 typedef usize = u64
 
+typedef DDWeight = s64
+typedef DDEpoch = u64
+typedef DDIteration = u64
+typedef DDTimestamp = DDTSGlobal{epoch: DDEpoch}
```
ryzhyk, line 25: We need to provide a string conversion function for this type so that timestamps can be easily printed. The function must have the following signature: `dDTimestamp2string(ts: DDTimestamp): string`

src/Language/DifferentialDatalog/Syntax.hs
```diff
@@ -521,13 +521,14 @@ instance Show Atom where
     show = render . pp
 
 -- The RHS of a rule consists of relational atoms with
--- positive/negative polarity, Boolean conditions, aggregation and
--- disaggregation (flatmap) operations.  The last two must occur after
+-- positive/negative polarity, Boolean conditions, aggregation,
+-- disaggregation (flatmap), inspect operations.  The last three must occur after
 -- all atoms.
```
ryzhyk, line 526: `The last three must occur after all atoms.` - I don't remember when I wrote this, but this comment is wrong and must be removed.

test/Spec.hs
```diff
@@ -74,41 +74,56 @@ allTests progress = do
         let datFile = replaceExtension dlFile "dat"
         -- If there's a gzipped reference file, use that
         let gzFile = replaceExtension dlFile "dump.expected.gz"
+        -- If there's a .log.expected file, also verify log output
```
ryzhyk, line 77: Nice!

PR #658, opened 2020-05-07
Support for transcoding data from JSON to binary formats.

ryzhyk:

### DIFF
--- a/lib/json.dl
+++ b/lib/json.dl
@@ -164,3 +164,27 @@ typedef JsonNum = // Integer number.  The value must be in the range between the
                 | // NaN and infinity are not valid JSON values.
                   // When serializing to JSON, such values will be replaced with 0.
                   JsonFloat{d: double}
+
+/* Wrapper type that serializes its inner object into binary formats via JSON string.
+ *
+ * This is a workaround for a `serde` limitation that prevents certain data types from
+ * being serialized or deserialized from non-self-describing data formats like bincode.
+ * In particular, this is the case for data types annotated with JSON-related serde
+ * attributes like `tag` and `flatten`.  In practice this means that types annotated for
+ * serialization to JSON cannot be serialized to other formats.
+ *
+ * The workaround is to use different serialization logic when serializing to binary and
+ * non-binary formats.  In the binary case, we want to serialize such types into a string
+ * and then serialize the string as a field in binary data.  In the non-binary case, we
+ * stick to normal serde behavior.  `JsonWrapper` implements this choice by checking the
+ * `is_human_readable` property of the serializer.
+ *
+ * The current implementation has an important flaw: serde currently does not provide a
+ * reliable way to determine if a serializer supports self-describing data.  We use
+ * `is_human_readable` attribute as an approximation, which is not always correct: binary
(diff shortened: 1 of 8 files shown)


test/datalog_tests/json_test.dl
```diff
@@ -183,3 +183,16 @@ JsonTest(s32FromString1(),
          to_json_string_or_default(from_json_string(s32FromString1()): Result<S32FromString, string>)).
 JsonTest(s32FromString2(),
          to_json_string_or_default(from_json_string(s32FromString2()): Result<S32FromString, string>)).
+
+typedef WrappedEnum = JsonWrapper<TaggedEnum>
```
mbudiu-vmw, line 187: no reference output?

lib/json.dl
```diff
@@ -20,7 +20,7 @@ extern function to_json_string(x: 'T): Result<string, string>
 
 /* Represents any valid JSON value.
  */
-#[rust="serde(from = \"serde_json::value::Value\", into = \"serde_json::value::Value\")"]
+#[rust="serde(from = \"__json::ValueWrapper\", into = \"__json::ValueWrapper\")"]
```
mbudiu-vmw, line 23: if the previous annotation was obscure, this one is doubly obscure.
Maybe once this API stabilizes we can wrap it into something that looks nicer for the user.

src/Language/DifferentialDatalog/Attribute.hs
```diff
@@ -127,6 +134,23 @@ tdefCheckSizeAttr d TypeDef{..} =
          [Attribute{..}] -> err d attrPos $ "Invalid 'size' attribute: size must be an integer between 0 and " ++ show (maxBound::Int)
          _                    -> err d tdefPos $ "Multiple 'size' attributes are not allowed"
 
+{- 'custom_serde' attribute: Tells DDlog not to generate `Serialize` and
```
mbudiu-vmw, line 137: it is not useful to have this documentation just in the code, you have to surface it to the users as well.
there should be a section on annotations in the reference document?

lib/json.dl
```diff
@@ -164,3 +164,27 @@ typedef JsonNum = // Integer number.  The value must be in the range between the
                 | // NaN and infinity are not valid JSON values.
                   // When serializing to JSON, such values will be replaced with 0.
                   JsonFloat{d: double}
+
+/* Wrapper type that serializes its inner object into binary formats via JSON string.
```
mbudiu-vmw, line 168: I don't really understand the connection between JSON, strings, and binary serialization.

test/datalog_tests/json_test.dl
ryzhyk: Just added that.

lib/json.dl
ryzhyk: Fortunately, this is internal to the library, users don't need to understand the annotations.

src/Language/DifferentialDatalog/Attribute.hs
ryzhyk: There is a section in the tutorial. I'll add the new attribute there.

PR #660, opened 2020-05-08
OVSDB: Skip DeltaMinus for non-root tables.

### LINKED ISSUE #655: OVSDB bindings can omit deletes and DeltaMinus tables when "isRoot": false
OVSDB has a "garbage collection" mechanism: in tables that have `"isRoot": false`, ovsdb-server automatically deletes rows that don't have any references from other rows. Currently, the OVSDB bindings for DDlog still send "delete" operations for these tables. These can be omitted, as (I think) the whole DeltaMinus table.

In addition to reducing memory and time, doing this will increase reliability because it eliminates the possibility of DDlog trying to delete a row that still has references. (This came up because I was seeing transaction errors for this reason while debugging.)

ryzhyk:
Resolves #655.

OVSDB has a "garbage collection" mechanism: in tables that have
"isRoot": false, ovsdb-server automatically deletes rows that don't have
any references from other rows. Currently, the OVSDB bindings for DDlog
still send "delete" operations for these tables. These can be omitted,
as the whole DeltaMinus table.

In addition to reducing memory and time, doing this will increase
reliability because it eliminates the possibility of DDlog trying to
delete a row that still has references.

### DIFF
--- a/rust/template/ddlog.h
+++ b/rust/template/ddlog.h
@@ -351,7 +351,7 @@ extern int ddlog_apply_ovsdb_updates(ddlog_prog hprog, const char *prefix,
 /*
  * Dump Delta-Plus, Delta-Minus, and Delta-Update tables for OVSDB table
  * `table` declared in DDlog module `module`, as a sequence of OVSDB insert,
- * delete, and update commands in * JSON format.
+ * delete, and update commands in JSON format.
  *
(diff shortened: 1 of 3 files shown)


ryzhyk:
It's a trivial change, so I will just go ahead and merge it. We can revise this feature in the future.

PR #661, opened 2020-05-08
ovsdb2ddlog: --rw option.

### LINKED ISSUE #659: feature request: --rw option to ovsdb2ddlog
ovsdb2ddlog has a `--ro` option for tables that are partially read-write. OVN has a couple of tables that only have one or two writable columns, so it'd be nice to just specify those, e.g. with `--rw` in place of `--ro`. It leads to easier maintenance as columns are added.

(Presumably only `--ro` or `--rw` could be used for a particular table.)

ryzhyk:
Resolves #659.

ovsdb2ddlog has a --ro option for tables that are partially read-write.
OVN has a couple of tables that only have one or two writable columns,
so it'd be nice to just specify those, e.g. with --rw in place of --ro.
It leads to easier maintenance as columns are added.

This commit introduces --rw option, which is mutually exclusive with
--ro, i.e., at most one of them can be used for a particular table.

@blp

### DIFF
--- a/src/Language/DifferentialDatalog/OVSDB/Compile.hs
+++ b/src/Language/DifferentialDatalog/OVSDB/Compile.hs
@@ -28,7 +28,10 @@ Module     : OVSDB.Compile
 Description: Compile 'OVSDB schema' to DDlog.
 -}
 
-module Language.DifferentialDatalog.OVSDB.Compile (compileSchema, compileSchemaFile) where
+module Language.DifferentialDatalog.OVSDB.Compile (
+    OutputRelConfig,
+    compileSchema,
+    compileSchemaFile) where
 
 import Prelude hiding((<>), readFile, writeFile)
 import Text.PrettyPrint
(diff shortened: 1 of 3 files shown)


src/Language/DifferentialDatalog/OVSDB/Compile.hs
```diff
@@ -85,11 +88,15 @@ data TableKind = TableInput
                | TableDeltaUpdate
                deriving(Eq)
 
+-- | Output relation configuration:
+-- (name, Left <set_of_readonly_columns> or Right <Set of rw columns>)
+type OutputRelConfig = (String, Either [String] [String])
```
mbudiu-vmw, line 93: Since you went to the trouble of defining a type you could probably have defined one with named fields; this would make it easier to read the code. But this is fine, since it's not used that much.

src/Language/DifferentialDatalog/OVSDB/Compile.hs
```diff
@@ -306,10 +316,11 @@ uuidCol = TableColumn { columnPos       = nopos
...
 
-tableGetNonROCols :: (?outputs::[(String, [String])]) => Table -> [TableColumn]
+tableGetNonROCols :: (?outputs::[OutputRelConfig]) => Table -> [TableColumn]
 tableGetNonROCols t =
     case lookup (name t) ?outputs of
-         Nothing -> ovscols
-         Just ro -> filter (\col -> notElem (name col) ro) ovscols
+         Nothing         -> ovscols
+         Just (Left ro)  -> filter (\col -> notElem (name col) ro) ovscols
+         Just (Right rw) -> filter (\col -> elem (name col) rw || (name col == "_uuid")) ovscols
```
mbudiu-vmw, line 324: this thing about uuid does not seem to be documented.
Also, the name of this function has a negation, it would be nicer if it was GetRWCols instead.

ryzhyk: > this thing about uuid does not seem to be documented.

done

> Also, the name of this function has a negation, it would be nicer if it was GetRWCols instead.

there's actually a subtle difference in this case: NonRO means that the column is not labeled as read-only. It can still be read-only if it comes from a read-only table.

PR #666, opened 2020-05-15
Implemented SQL over operator

mbudiu-vmw:


mbudiu-vmw:
@lalithsuresh : this is a big PR, but hopefully you can check just the queries that you care about.

PR #672, opened 2020-05-17
D3log Accumulator

sugohugo:
Adds the `DistributingAccumulator` component to D3log, as well as the `ip_discovery` example from the [RADIO'20 paper](https://github.com/ryzhyk/radio2020-d3log).

## Rationale
To increase fault-tolerance and allow dynamic re-configuration in a distributed setting, a possible approach is to keep track of relation updates received and sent by nodes in an efficient accumulated manner. Upon node failure or addition of a new sink to the network, a single transaction including all accumulated updates then suffices to have a consistent state in the network.

## Design
An Accumulator implements the `Accumulator` trait, which is an `Observer` and `Observable` simultaneously. The `DistributingAccumulator` is a subtype that observes a single observable and can be by multiple observers.

Conceptually, the `DistributingAccumulator` consists of two different components:
the `AccumulatingObserver` and a `TxnDistributor`. It encapsulates the functionality
of those components in a single structure.

The `AccumulatingObserver` is a simple proxy between an observable and an observer that
inspects the data that passes through. It accumulates the data to keep track of the
current state. The data structure used for aggregating the updates is a simple `HashMap<RelId, HashSet<V>>`, that maps each relation to a set of values currently included in it. This assumes that the Accumulator always starts with an empty set and should never accumulate negative multiplicities.

The `TxnDistributor` is the inverse of the `TxnMux` class, it listens to a single observable and
is able to send data to multiple observers. 

The following schema visualizes the relationship between the different components:
```
                   +----------------------------------------------------------------------+                     
                   |                                                                      |                     
                   |                      DistributingAccumulator                         |                     
                   |                                                                      |   /--Observer       
                   |  +----------------------+                 +---------------------+  /-|---                  
                   |  |                      |                 |                     ---  |                     
 Observable -------|---AccumulatingObserver  |-----------------|  TxnDistributor     -----|------Observer       
                   |  |                      |                 |                     | \--|                     
                   |  |                      |                 |                     |    |---                  
                   |  +----------------------+                 +---------------------+    |   \- Observer       
                   |                                                                      |                     
                   |                                                                      |                     
                   +----------------------------------------------------------------------+     
 ```

When a component subscribes to `DistributingAccumulator`, the accumulator acquires a lock for the `TxnDistributor`, collects the accumulated updates from `AccumulatingObserver` and sends it to the component and then subscribes the component to `TxnDistributor`, releasing the lock afterwards.


sugohugo:
With a file as a source (`sources/file.rs`), a race condition occurred in the following scenario: Until now a file source has only been handed over via the `add_observable` method to a `TxnMux`. If one simply uses the Observable interface and the subscribe method, the following occurs:
 1. the file source is subscribed to in some method, e.g. in `instantiate.rs`
 2. the method ends and the file source goes out of scope, thus triggering the `Drop` trait
 3. the implementation of `Drop` consists of simply unsubscribing everything from the file source
 4. when unsubscribing, the opened file descriptor will be closed. However, therefore also the spawned thread that reads from the file and sends transactions to the observer cannot access the file anymore and aborts. Depending on how the calls were ordered in instantiate.rs and the timing, transactions have been sent or not.

After trying different solutions with the goal of changing as few as possible, I arrived at the conclusion that the easiest thing to do would be to not implement Drop for a file source. If one calls unsubscribe on the file source still all resources should be freed and cleaned up. I think it is also a more intuitive solution, because the file source Observable can be an independent entity without the need of having an owner. 

However, I am not fully aware of the consequences of this decision and just tried if all tests still work (which they should). @d-e-s-o , @ryzhyk suggested if I could ask you to take a look at it because you're the one that implemented it in the first place. Would this be possible? I would very much appreciated it.

d-e-s-o:
Glad to see progress being made on `d3log` :-)

> However, I am not fully aware of the consequences of this decision and just tried if all tests still work (which they should). @d-e-s-o , @ryzhyk suggested if I could ask you to take a look at it because you're the one that implemented it in the first place. Would this be possible?

I am happy to take a look at the solution to the race condition, but I probably won't have the time to provide a detailed review on the full pull request. Are https://github.com/vmware/differential-datalog/pull/672/commits/242b1da7dc4f2ed70c04bfa3bbcb142afd1b79a0 and https://github.com/vmware/differential-datalog/pull/672/commits/ffb80fc80d987c1233a4930fed90eb977dee4dcb the two commits of interest? By any chance, do you have a test reproducing the issue?

d-e-s-o:
> With a file as a source (`sources/file.rs`), a race condition occurred in the following scenario: Until now a file source has only been handed over via the `add_observable` method to a `TxnMux`. If one simply uses the Observable interface and the subscribe method, the following occurs:
> 
> 1. the file source is subscribed to in some method, e.g. in `instantiate.rs`
> 2. the method ends and the file source goes out of scope, thus triggering the `Drop` trait
> 3. the implementation of `Drop` consists of simply unsubscribing everything from the file source
> 4. when unsubscribing, the opened file descriptor will be closed. However, therefore also the spawned thread that reads from the file and sends transactions to the observer cannot access the file anymore and aborts. Depending on how the calls were ordered in instantiate.rs and the timing, transactions have been sent or not.
> 
> After trying different solutions with the goal of changing as few as possible, I arrived at the conclusion that the easiest thing to do would be to not implement Drop for a file source. If one calls unsubscribe on the file source still all resources should be freed and cleaned up. I think it is also a more intuitive solution, because the file source Observable can be an independent entity without the need of having an owner.

Okay, so if I understand the issue correctly, then I am not sure I would go with simply leaking the `FileSource` internals. I looked at your version of `instantiate.rs` and yes, the way `FileSource` is used will result in issues because, as you say, it won't survive past the end of `add_file_sources`.

What I think would need to happen is that you pass out the `FileSource` objects created, ultimately embedding them in the `Realization` being returned. That's exactly the purpose of `Realization`: to keep all the state that got instantiated. Once it is dropped, the entire "instantiation" is torn down. You'd almost certainly want to keep file sources until that point, as the idea is that those can provide updates later on, if the file is updated (that may make more sense for other types of sources, but file sources already support that use case if I remember correctly). So while leaking the internals would yield the same outcome, it will cause trouble more trouble down the road.

If you look at `create_txn_mux` you will see that `TxnMux` essentially is used the same way as the `FileSource` in your case (i.e., created and then subscribed) and so it has a similar problem and, hence, the same solution should apply. Let me know if I misunderstood.

sugohugo:
Thanks a lot for the quick and very valuable feedback! 

> Are 242b1da and ffb80fc the two commits of interest? By any chance, do you have a test reproducing the issue?

Yes, you're right about the commits, those are basically the two different approaches I had. However, the first one turned out to be more difficult than I thought, because the created thread will of course never end due to the infinite `read_line()` calls as long as there is an open file descriptor.

> What I think would need to happen is that you pass out the FileSource objects created, ultimately embedding them in the Realization being returned. That's exactly the purpose of Realization: to keep all the state that got instantiated. Once it is dropped, the entire "instantiation" is torn down. 

That sounds completely like the right thing to do, I agree. I was just looking for a quick and non-disruptive fix for the issue, but as you mentioned, it is not very elegant. 

Currently, `Realization` just consists of the `TxnMux` because everything is registered to it as you mentioned and adding `FileSource` to `Realization` is a plausible solution. However, I will run into the same kind of problem with `TcpReceiver` further down the road, so maybe it would be more feasible to have an `add_observable` method also in the accumulator, thus `FileSource` or `TcpReceiver` is owned by `DistributingAccumulator`, which itself is owned by `TxnMux`. 

When using the accumulator for sinks, one would need a different approach, as the accumulator should not take ownership over the server, as multiple accumulators are needed for different output relations. So what I have thought about now is whether the `DistributingAccumulator` should be part of `Realization` instead for sinks. One anyhow needs some way of accessing the accumulator when a new node wants to subscribe for a certain output relation, which lead me to the thought of adding some kind of `Map<RelId, DistributingAccumulator>` to `Realization`.

I do not know how aware you are about the concept of the accumulator (and also how much time you want to spend on this discussion), so maybe @ryzhyk can also add his thoughts about this, especially how the concept of adding a new node as subscriber to an existing node should work.

d-e-s-o:
> I do not know how aware you are about the concept of the accumulator (and also how much time you want to spend on this discussion), so maybe @ryzhyk can also add his thoughts about this, especially how the concept of adding a new node as subscriber to an existing node should work.

I remember the concept of the accumulator coming up in design discussions in the past. It will only be Friday earliest that I have some time to think through that, I am afraid.

sugohugo:
> I remember the concept of the accumulator coming up in design discussions in the past. It will only be Friday the earliest that I have some time to think through that, I am afraid.

No problem, I think it won't be necessary anymore. I had a discussion with @ryzhyk 12 hours ago which lead to the decision of expanding the `Realization` to contain more objects than just the `TxnMux`, as you mentioned. Sorry for not telling earlier, thanks a lot again!

ryzhyk:
@d-e-s-o, thank you so much for helping to clarify things!

ryzhyk:
@sugohugo , I converted the PR into a draft for the time being.

sugohugo:
I added the deletion functionality, changed the structure of the `Realization` struct and added a new example for D3log computation `dbpedia_ownership`, based on company ownership data of dbpedia. For this, file sources can now parse the `"exit;"` command, which signals the end of the computation and triggers the output of a timing (using `std::time::Instance`).

There are still possible optimizations, however, in the short-term I'd like to come to a final state of the PR for the scope of my bachelor thesis, if possible in your opinion. If you're interested, below are the results of my (for now very simple) benchmarks using the `dbpedia_ownership` example (time in ms).
image

ryzhyk:
Great, thank you! Yes, sounds like you're very close to wrapping up the PR. Can you please add a README to dbpedia_ownership explaining what the example does and how to run it, including the purpose of the patch file? Are there any other documentation improvements that would help future d3log developers to understand your accumulator implementation?  You mention possible optimizations.  Can you please document them as well?

Now that I look at the implementation, I realize the clearing the state in `on_completed` is probably not the right design, as it conflates the notions that a data source will not produce any more data and that the data source has been removed and should be considered empty. I suppose we'll have to leave this for future work.

ryzhyk:
PS. The description of this PR (Rationale + Design) should probably be copied to the README in the `accumulate` directory.

sugohugo:
That's great to hear! I added some documentation and additional README files, I hope this helps in clarification and future development. I tried providing good code documentation for the accumulator, so I hope this also helps. Additionally, the unit tests in the Rust files also exemplify what the expected usage should be. 

I agree with "abusing" `on_completed` now for signalling a source failure. Probably the best thing to do would be to extend the `Observer` interface with an additional method (e.g. `on_failure`). This would lead to some refactoring and also some adaption of unit tests for a clean implementation, so I would prefer leaving this as future work.

sugohugo:
Added some fixes for the Gitlab CI, now all tests are running through.

PR #673, opened 2020-05-17
API+CLI support to dump initial state.

ryzhyk:
DDlog's normal mode of operation is to dump changes to output relations
at the end of every transaction.  The client can aggregate these changes
outside of DDlog using some application-specific logic.  However, until
now, the API lacked support for dumping initial contents of output
relations.

This commit closes the gap by extending the DDlog API to return initial
output snapshot from `ddlog_run()`.  It also modifies the CLI to dump
initial state by default and introduces a CLI flag to disable this
behavior.  It also makes the CLI "usage" message more informative.

@mbudiu-vmw , I have so far fixed up souffle tests by simply dropping
initial state on the floor. I do the same in Go and Java API, and in
distributed-ddlog.


mbudiu-vmw:
We can make this even more general by allowing the user to supply the first commit explicitly. This would make some more soufflé programs work.

PR #677, opened 2020-05-24
ovsdb2ddlog: Encode optional columns as Option<>.

ryzhyk:
OVSDB columns representing sets are normally encoded as a DDlog sets.
However in the special case where column's multiplicity bounds are min: 0 an
max: 1, `Option<>` is a more natural representation.  This commit
implements this optimization.

But now we have to deal with the complication that the Rust code that
converts OVSDB JSON to DDlog records must correctly deserialize
arrays into `Option<>`.  This code is agnostic of the OVSDB schema and
cannot easily tell what the target type is.  We therefore change the
`FromRecord` implementation for `Option<>` to handle this case.  This in
turn required adding a new attributes `#[custom_from_record]` to
override the autogenerated implementation of `FromRecord` created by
DDlog.

### DIFF
--- a/lib/std.dl
+++ b/lib/std.dl
@@ -19,7 +19,7 @@ typedef s128 = signed<128>
 
 typedef usize = u64
 
-/* 
+/*
  * Weight of a record in a DDlog collection.  This type is internal to the 
  * DDlog engine and is only visible to the programmer inside the `Inspect`
  * operator.  Positive weight indicates that the record is being added
@@ -80,6 +80,8 @@ extern function pow32(base:'A, exp: bit<32>): 'A
  */
 
 #[rust="serde(from=\"Option<A>\", into=\"Option<A>\", bound(serialize=\"A: Clone+Serialize\"))"]
+// This type has a custrom FromRecord implementation in std.rs.
+#[custom_from_record]
 typedef Option<'A> = None
                    | Some{x: 'A}
 
--- a/lib/std.rs
+++ b/lib/std.rs
@@ -145,6 +145,62 @@ impl<T> From<std_Option<T>> for Option<T> {
     }
 }
 
+impl<A: record::FromRecord + serde::de::DeserializeOwned + Default> record::FromRecord
+    for std_Option<A>
+{
+    fn from_record(val: &record::Record) -> result::Result<Self, String> {
+        match val {
+            record::Record::PosStruct(constr, args) => match constr.as_ref() {
+                "std.None" if args.len() == 0 => Ok(std_Option::std_None {}),
+                "std.Some" if args.len() == 1 => Ok(std_Option::std_Some {
+                    x: <A>::from_record(&args[0])?,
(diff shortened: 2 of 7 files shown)


src/Language/DifferentialDatalog/Attribute.hs
```diff
@@ -157,6 +164,23 @@ tdefGetCustomSerdeAttr d tdef =
          Left e  -> error e
          Right b -> b
 
+{- 'custom_from_record' attribute: Tells DDlog not to generate `FromRecord`
+ - implementation for a type.  The user must write their own implementations in
```
mbudiu-vmw, line 168: it would be nice if the signature of this implementation was documented someplace.

lib/std.dl
```diff
@@ -80,6 +80,7 @@ extern function pow32(base:'A, exp: bit<32>): 'A
  */
 
 #[rust="serde(from=\"Option<A>\", into=\"Option<A>\", bound(serialize=\"A: Clone+Serialize\"))"]
+#[custom_from_record]
```
mbudiu-vmw, line 83: I would add here a comment indicating what this means: the implementation must be given in the corresponding rust file, and it must have two functions with some signatures.

lib/std.rs
```diff
@@ -145,6 +145,62 @@ impl<T> From<std_Option<T>> for Option<T> {
     }
 }
 
+impl<A: record::FromRecord + serde::de::DeserializeOwned + Default> record::FromRecord
+    for std_Option<A>
+{
+    fn from_record(val: &record::Record) -> result::Result<Self, String> {
```
mbudiu-vmw, line 151: it's in fact trickier, because the implementation has to know about this Record type. So then that has to be documented someplace.

lib/std.rs
```diff
@@ -145,6 +145,62 @@ impl<T> From<std_Option<T>> for Option<T> {
...
+                    x: record::arg_extract::<A>(args, "x")?,
+                }),
+                c => result::Result::Err(format!(
+                    "unknown constructor {} of type std_Option in {:?}",
+                    c, *val
+                )),
+            },
+            /* `Option` encoded as an array of size 0 or 1.  This is, for instance, useful when
+             * interfacing with OVSDB. */
+            record::Record::Array(kind, records) => match (records.len()) {
```
mbudiu-vmw, line 175: what about other collections?

src/Language/DifferentialDatalog/Attribute.hs
ryzhyk: I don't think users should be writing these annotations. This is an ugly workaround for the OVSDB problem that I could not solve properly without a major type system overhaul.

lib/std.rs
ryzhyk: sets, maps, etc., are extern types and already have their own manual implementations of `FromRecord` that support arrays.

PR #681, opened 2020-06-05
Type inference, part 1.

ryzhyk:
This includes some refactoring of existing code for type inference and the first part of the actual inference algorithm (constraint generation).

### DIFF
--- a/src/Language/DifferentialDatalog/Expr.hs
+++ b/src/Language/DifferentialDatalog/Expr.hs
@@ -70,7 +70,9 @@ import Language.DifferentialDatalog.Syntax
 import Language.DifferentialDatalog.NS
 import Language.DifferentialDatalog.Util
 import Language.DifferentialDatalog.Name
-import Language.DifferentialDatalog.Type
+import {-# SOURCE #-} Language.DifferentialDatalog.Type
+import Language.DifferentialDatalog.ECtx
+import Language.DifferentialDatalog.Var
 import Language.DifferentialDatalog.Function
 
 -- depth-first fold of an expression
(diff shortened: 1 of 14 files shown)


src/Language/DifferentialDatalog/TypeInference.hs
```diff
@@ -0,0 +1,637 @@
...
+                         | TypeInferenceAmbiguity
+
+-- Uniquely identifies an expression in a DDlog program.
+data DDExpr = DDExpr {ddexprCtx::ECtx, ddexprExpr::Expr}
+
+-- A type constraint is an atomic predicate or a disjuncion of conjuntions of
+-- predicates.  We use a unification-based solver to resolv type constraints,
+-- which does not handle arbitrary Boolean combination constraints, hence this
+-- restricted structure.
+data Constraint = CPredicate Predicate
```
mbudiu-vmw, line 59: To give nice error messages you may also want to keep with each constraint lineage information - this can be either something that implements pos, or another constraint; this way, when a constraint fails you can indicate where it originated and finally point to the source-level position that caused the constraint to be generated. Look at the llvm error messages when type unification fails for an inspiration.

src/Language/DifferentialDatalog/TypeInference.hs
```diff
@@ -0,0 +1,637 @@
...
+deIsNum :: DDExpr -> Constraint
+deIsNum de = CDisj $ deIsInt_ de ++ deIsFP_ de
+
+-- Type expression is a shared reference type with specified inner type.
+deIsSharedRef :: (?d::DatalogProgram) => DDExpr -> TExpr -> [Constraint]
+deIsSharedRef de t =
+    cdisj $ map (\tdef -> [PEq (TETypeOfExpr de) (TEExtern (name tdef) [t])]) sref_tdefs
+    where
+    -- All shared reference types.
+    sref_tdefs = filter (tdefGetSharedRefAttr ?d)
```
mbudiu-vmw, line 187: I find it a bit strange that you can lookup stuff in the program while creating constraints, but I don't see why this would not work.

src/Language/DifferentialDatalog/TypeInference.hs
```diff
@@ -0,0 +1,637 @@
...
+    [ PEq (TETypeOfExpr de) (TEExtern mAP_TYPE [TETVar ddexprCtx (name k), TETVar ddexprCtx (name v)])
+    , PEq t (TETuple (Just 2) $ M.fromList [(0, TETVar ddexprCtx (name k)), (1, TETVar ddexprCtx (name v))])]
+    where
+    TypeDef{tdefArgs=[k,v],..} = getType ?d mAP_TYPE
+
+-- Convert type to a type expression, replacing type arguments ('A, 'B, ...)
+-- with type constants 'TETArg "A", TETArg "B", ...'.  For example, when
+-- generating type inference in the body of a function, we treat its type
+-- arguments as constants.  Inferred types for variables and expressions inside
+-- the body of the function may depend on this constants.
```
mbudiu-vmw, line 226: "these" constants.
Could the constants be just other type variables?
This can happen when analyzing a generic function calling another generic function.

src/Language/DifferentialDatalog/TypeInference.hs
```diff
@@ -0,0 +1,637 @@
...
+--
+-- '|e| = |e1| = |e2| and is_bits|e1|'
+                                                | elem exprBOp [BOr, BAnd, BXor] =
+    [ TETypeOfExpr l === TETypeOfExpr r
+    , TETypeOfExpr de === TETypeOfExpr l
+    , deIsBits l ]
+
+-- Binary operator 'e1 ++ e2'.
+--
+-- is_String(|e1|) and is_String(|e2|) or is_Bit |e1| and is_Bit |e2|
```
mbudiu-vmw, line 568: I was expecting a constraint on the lengths for the bits, but if you go bottom up it's not necessary

src/Language/DifferentialDatalog/Type.hs
```diff
@@ -869,3 +870,40 @@ checkIterable :: (MonadError String me, WithType a) => String -> Pos -> DatalogP
...
+varCheckType :: (MonadError String me) => DatalogProgram -> Var -> me Type
+varCheckType d v =
+    case varTypeMaybe d v of
+         Nothing -> err d (pos v) $ "Type of variable '" ++ name v ++ "' is unknown"
+         Just t  -> return t
+
+varType :: DatalogProgram -> Var -> Type
+varType d v =
+    case varTypeMaybe d v of
+         Nothing -> error $ "Type of variable '" ++ name v ++ "' is unknown"
```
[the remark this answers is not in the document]
ryzhyk: No, this is an internal error, it's only there for debugging.

src/Language/DifferentialDatalog/Expr.hs
```diff
@@ -238,39 +240,54 @@ exprVarOccurrences ctx e = exprCollectCtx (\ctx' e' ->
...
-                       EBinding _ v e'' -> (v, ctx') : e''
-                       ETyped _ e'' _   -> e''
-                       _                -> []) ctx e
+                  let e_ = exprMap fst e' in
+                  (E e_,
+                   case e' of
+                        EStruct _ _ fs   -> concatMap (snd . snd) fs
+                        ETuple _ fs      -> concatMap snd fs
+                        EVarDecl _ _     -> [ExprVar ctx' e_]
+                        -- Inside positive literals, variables are declared
```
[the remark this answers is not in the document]
ryzhyk: Because we had separate functions that knew about variable declarations in rules. It was a mess.

src/Language/DifferentialDatalog/NS.hs
```diff
@@ -87,16 +81,17 @@ checkTransformer p d n = case lookupTransformer d n of
...
-checkVar :: (MonadError String me) => Pos -> DatalogProgram -> ECtx -> String -> me Field
+checkVar :: (MonadError String me) => Pos -> DatalogProgram -> ECtx -> String -> me Var
 checkVar p d c n = case lookupVar d c n of
                         Nothing -> err d p $ "Unknown variable: " ++ n -- ++ ". All known variables: " ++ (show $ (\(ls,vs) -> (map name ls, map name vs)) $ ctxVars d c)
                         Just v  -> return v
 
-getVar :: DatalogProgram -> ECtx -> String -> Field
-getVar d c n = fromJust $ lookupVar d c n
+getVar :: DatalogProgram -> ECtx -> String -> Var
+getVar d c n =
```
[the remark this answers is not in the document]
ryzhyk: Same thing, this is only there to debug compiler crashes.

src/Language/DifferentialDatalog/NS.hs
```diff
@@ -87,16 +81,17 @@ checkTransformer p d n = case lookupTransformer d n of
 getTransformer :: DatalogProgram -> String -> Transformer
 getTransformer d n = fromJust $ lookupTransformer d n
 
-lookupVar :: DatalogProgram -> ECtx -> String -> Maybe Field
+lookupVar :: DatalogProgram -> ECtx -> String -> Maybe Var
```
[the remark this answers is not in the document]
ryzhyk: `Var` is a new abstraction. Given a `Var` you can always find out its type. `Field` combines variable name and type.  Like I said, a mess...

src/Language/DifferentialDatalog/TypeInference.hs
ryzhyk: I plan to do this in the next PR. It will be clearer how to generate explanations and what state needs to be tracked as I make progress on the algorithm.

src/Language/DifferentialDatalog/TypeInference.hs
ryzhyk: This is a bit confusing, but when doing type inference inside a function, this functions type arguments are constants in the sense that in the resulting typing variable types can be expressed in terms of these constants and we don't expect to assign concrete types to these constants.  When we call another generic function, it's a completely different situation. That function's type arguments are variables that we actually want to assign to concrete types, and so we model them as variables (TETVar), not constants.

src/Language/DifferentialDatalog/TypeInference.hs
ryzhyk: you're right, and this is what the implementation actually does. I will update the comment.

ryzhyk:
Thanks! I will include updated comments in the next PR.

PR #683, opened 2020-06-10
Support deserialize_from_array attribute on Optional Maps

mbudiu-vmw:
This may need more testing.


mbudiu-vmw:
Unfortunately this does not work, because the rust annotations do not seem to compose in the right way. An Option field that has an annotation is not deserialized as an Option field.

PR #688, opened 2020-06-19
Dump output of intermediate compiler passes.

ryzhyk:


app/Main.hs
```diff
@@ -169,22 +145,24 @@ parseValidate Config{..} = do
     d''' <- case confOutputInput of
          "" -> return d''
          x  ->  return $ progMirrorInputRelations d'' x
+    when confDumpFlat $
+        writeFile (replaceExtension confDatalogFile ".flat.ast") (show d''')
```
mbudiu-vmw, line 149: The ast suffix does not really make sense. I think you should use the dl suffix, then syntax highlighting tools will work properly.

test/datalog_tests/lib_test.debug.ast.expected
```diff
@@ -0,0 +1,1053 @@
...
+typedef json.JsonNum = json.JsonInt{i: std.s128} | json.JsonFloat{d: double}
+#[rust = "serde(from = \"__json::ValueWrapper\", into = \"__json::ValueWrapper\")"]
+typedef json.JsonValue = json.JsonNull{} | json.JsonBool{b: bool} | json.JsonNumber{n: json.JsonNum} | json.JsonString{s: internment.istring} | json.JsonArray{a: std.Vec<json.JsonValue>} | json.JsonObject{o: std.Map<internment.istring,json.JsonValue>}
+#[custom_serde = true]
+typedef json.JsonWrapper<'T> = json.JsonWrapper{x: 'T}
+typedef json_test.Array = std.Vec<json_test.BoolStruct>
+typedef json_test.BoolMap = std.Map<string,json_test.BoolStruct>
+typedef json_test.BoolStruct = json_test.BoolStruct{b: bool}
+typedef json_test.Enum = #[rust = "serde(rename = \"Variant1\")"]
+                         json_test.Variant1{b: bool} | #[rust = "serde(rename = \"Variant2\")"]
```
mbudiu-vmw, line 26: could the indentation be changed here?

app/Main.hs
ryzhyk: I thought about this, but unfortunately this is not valid DDlog code, as it includes flattened names containing dots.

test/datalog_tests/lib_test.debug.ast.expected
ryzhyk: Yeah, looks terrible.

app/Main.hs
```diff
@@ -169,22 +145,24 @@ parseValidate Config{..} = do
...
     d'''' <- case validate d''' of
                Left e   -> errorWithoutStackTrace $ "error: " ++ e
                Right d'''' -> return d''''
+    when confDumpValid $
+        writeFile (replaceExtension confDatalogFile ".valid.ast") (show d'''')
     d' <- case confDebugHooks of
          False -> return d''''
          True  -> return $ progInjectDebuggingHooks d''''
+    when confDumpDebug $
+        writeFile (replaceExtension confDatalogFile ".debug.ast") (show d')
```
haroldlim, line 159: Should we add check to only dump when confDebugHooks is enabled?

Or maybe update help message that if pp-debug is passed without "-g" then this will dump the source but no debug hooks?

ryzhyk: Yeah, I was being lazy there. Let's go with the second approach, as it is more permissive, i.e., works better with automation.

PR #689, opened 2020-06-25
Update debugging hooks function and format of inputs/outputs

haroldlim:
1. Include operator type in the debug event
2. Convert the inputs and outputs into record before writing to file.


src/Language/DifferentialDatalog/Debug.hs
```diff
@@ -123,11 +123,20 @@ generateInspectDebug d ruleIdx rule index =
...
                 then eVar $ exprVar $ enode $ atomVal $ rhsAtom $ head $ ruleRHS rule
                 else head $ Compile.recordAfterPrefix d rule (index - 1)
+    opType = if index == 0
+                then "Map"
+                else case (ruleRHS rule) !! index of
+                     RHSLiteral{rhsPolarity=False} -> "Antijoin"
+                     RHSInspect{} -> "Inspect"
+                     RHSFlatMap{} -> "Flatmap"
+                     RHSCondition{} -> "Condition"
+                     _ -> "Undefined" -- Should not reach this case
```
ryzhyk, line 133: Let's do
```
rhs -> error $ "generateInspectDebug " ++ show rhs
```
to catch potential compiler bugs that violate this.

lib/debug.dl
```diff
@@ -2,7 +2,9 @@
  * an external debugger tool.
  */
 
-extern function debug_event(operator_id: (u32, u32, u32), w: DDWeight, ts: 'T1, input1: 'A1, out: 'A2): ()
-extern function debug_event_join(operator_id: (u32, u32, u32), w: DDWeight, ts: 'T1, input1: 'A1, input2: 'A2, out: 'A3): ()
+extern function debug_event(operator_id: (u32, u32, u32), w: DDWeight, ts: 'T1,
```
ryzhyk, line 5: In one of future commits, can we add a typedef to `debug.dl`, e.g.,
```
typedef DDlogOpId = DDlogOpId {
  rel: u32,
  rule: u32,
  op: u32
}
```

ryzhyk: On a second thought, since this will be serialized into every debugger message, maybe we should keep it as a tuple for more concise output:
```
typedef DDlogOpId = (u32, u32, u32)
```

PR #691, opened 2020-07-03
New C APIs for ovn-northd-ddlog

blp:
This adds the C APIs I needed to make reconnection work with output-only tables in OVN in commit https://github.com/blp/ovs-reviews/commit/a0cda0ec5b93276f657c9aa88658c14a1300236d

### DIFF
--- a/rust/template/ddlog.h
+++ b/rust/template/ddlog.h
@@ -700,6 +700,33 @@ extern int ddlog_dump_ovsdb_output_table(
     char **json);
 
 /*
+ * Serializes table record 'rec' into a JSON object for an OVSDB insert,
+ * update, or delete operation, respectively, to apply to OVSDB table 'table'.
+ *
+ * On success, returns `0` and stores a pointer to JSON string in
+ * `json`.  This pointer must be later deallocated by calling
+ * `ddlog_free_json()`
+ *
+ * On error, returns a negative number and writes error message
+ * (see `print_err_msg` parameter to `ddlog_run()`).
+ */
+extern int ddlog_into_ovsdb_insert_str(
+    ddlog_prog prog,
+    const char *table,
+    const ddlog_record *rec,
+    char **json);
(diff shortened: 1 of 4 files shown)


ryzhyk:
This looks great!

How do you use the new OVSDB APIs?

blp:
> How do you use the new OVSDB APIs?

This comment is a good start at explaining why I needed the APIs to generate update and insert operations: https://github.com/vmware/differential-datalog/pull/691#issuecomment-653605472

I needed the ddlog_get_named_struct_field() function to get the _uuid value from a row struct.

blp:
Argh I don't understand why that link didn't work but here's what I was referencing:

```
        /*
         * We just reconnected to the database (or connected for the first time
         * in this execution).  We assume that the contents of the output-only
         * tables might have changed (this is especially true the first time we
         * connect to the database a given execution, of course; we can't
         * assume that the tables have any particular contents in this case).
         *
         * ctx->data.output_only_data is a database reply that tells us the
         * UUIDs of the rows that exist in the database.  Our strategy is to
         * compare these UUIDs to the UUIDs of the rows that exist in the DDlog
         * analogues of these tables, and then add, delete, or update rows as
         * necessary.
         *
         * (ctx->data.output_only_data only gives row UUIDs, not full row
         * contents.  That means that for rows that exist in OVSDB and in
         * DDLog, we always send an update to set all the columns.  It wouldn't
         * save bandwidth to do anything else, since we'd always have to send
         * the full row contents in one direction and if there were differences
         * we'd have to send the contents in both directions.  With this
         * strategy we only send them in one direction even in the worst case.)
         *
         * (We can't just send an operation to delete all the rows and then
         * re-add them all in the same transaction, because ovsdb-server
         * rejecting deleting a row with a given UUID and the adding the same
         * UUID back in a single transaction.)
         */
 ```

PR #698, opened 2020-07-08
Add debug parser for the dump event file

yjiayu:


src/Language/DifferentialDatalog/Debugger/DebugEventParser.hs
```diff
@@ -0,0 +1,160 @@
...
+import Data.Functor.Identity
+
+data CollectionType  = Vector deriving (Show)
+data Operator = OpMap | OpAggregate | OpCondition | OpJoin
+                | OpAntijoin | OpInspect | OpUndefined deriving (Show)
+
+data Events = Events { events :: [Event]} deriving (Show)
+
+data Event = DebugEvent { operatorId :: (Integer, Integer, Integer)
+                         , weight :: Integer
```
ryzhyk, line 39: Unfortunately, all field names in Haskell are functions in the global namespace, so fields like `weight` and `timestamp` will likely introduce name clashes.  I therefore use qualified field names like `evtWeight`, `evtTimestamp`, etc.

src/Language/DifferentialDatalog/Debugger/DebugEventParser.hs
```diff
@@ -0,0 +1,160 @@
...
+                              <*> (comma *> decimal) <*> (comma *> decimal) <*> (comma *> operatorParser)
+                              <*> (comma *> recordParser) <*> (comma *> recordParser)
+
+debugJoinEventParser :: ParsecT String u Identity Event
+debugJoinEventParser = DebugJoinEvent <$> parens ((,,) <$> (decimal <* comma) <*> (decimal <* comma) <*> decimal)
+                              <*> (comma *> decimal) <*> (comma *> decimal) <*> (comma *> operatorParser)
+                              <*> (comma *> recordParser) <*> (comma *> recordParser) <*> (comma *> recordParser)
+
+eventParser :: ParsecT String u Identity Event
+eventParser = debugEventParser <|> debugJoinEventParser
```
ryzhyk, line 157: Parsec does not do lookahead automatically, so this rule will always run `debugEvenParser`, which will successfully consume the first few fields, and then fail if this is a join event. Try this approach instead:

```
debugEventParser = do
    opid <- parens ((,,) <$> (decimal <* comma) <*> (decimal <* comma) <*> decimal)
    w <- comma *> decimal
    ts <- comma *> decimal <* comma
    op <- operatorParser
    case op of
             OpJoin -> ... -- parse three records
                              return DebugJoinEvent{ ....}
             _ ->  ... -- parse two records
                     return DebugEvent{ ...}
```

ryzhyk:
@yjiayu , one more comment: in future PRs you might want to add gitlab jobs to test your debugger work (see `.gitlab-ci.yml`)

PR #699, opened 2020-07-09
Streams and multisets.

ryzhyk:
Support for stream and multiset relations.  See commit messages for details.


rust/template/differential_datalog/program.rs
```diff
@@ -1662,8 +1716,12 @@ impl Program {
                     while cursor.val_valid(&storage) && *cursor.key(&storage) == k {
                         let mut weight = 0;
                         cursor.map_times(&storage, |_, diff| weight += diff);
-                        assert!(weight >= 0);
-                        if weight > 0 {
+                        //assert!(weight >= 0);
```
mbudiu-vmw, line 1719: I think that what you are saying here is that we don't have a semantics for operations that combine sets and multisets. This is the most important thing to figure out. Ideally we would be very explicit about this, giving users control by using either distinct or a conversion in the other way, that casts a set to a multiset preserving weights. I was a little uncomfortable about implementing sets with multisets even previously, because we never had a proof that we achieve the expected datalog semantics by inserting distinct in some places. Is this something that the Vienna guys would be interested in working on?

rust/template/differential_datalog/program.rs
```diff
@@ -2359,16 +2436,104 @@ impl RunningProgram {
...
+                ve.insert(-1);
             }
         }
     }
 
+    /* Update delta set of an input stream relation before performing an update.
+     * `ds` is delta since start of transaction.
+     * `x` is the value being inserted or deleted.
+     * `insert` indicates type of update (`true` for insert, `false` for delete) */
+    fn stream_update(
```
mbudiu-vmw, line 2458: is there any opportunity for code reuse by inheriting the behavior?

rust/template/differential_datalog/program.rs
```diff
@@ -2594,29 +2769,37 @@ impl RunningProgram {
...
-            // first delete, then insert to avoid duplicate key
-            // errors in `apply_updates()`
-            for (k, w) in rel.delta() {
-                if *w {
+    fn delta_undo_updates(relid: RelId, ds: &DeltaSet, updates: &mut Vec<Update<DDValue>>) {
+        // first delete, then insert to avoid duplicate key
+        // errors in `apply_updates()`
+        for (k, w) in ds {
+            if *w >= 0 {
+                for _ in 0..*w {
```
mbudiu-vmw, line 2777: this looks inefficient; weights could potentially be very large

rust/template/differential_datalog/valmap.rs
```diff
@@ -90,8 +90,12 @@ impl<V: Display + Ord + Clone> DeltaMap<V> {
         for (relid, map) in &self.map {
             w.write_fmt(format_args!("{}:\n", R::relid2name(*relid).unwrap()))?;
             for (val, weight) in map {
-                w.write_fmt(format_args!("{}\n", *val))?;
-                assert_eq!(*weight, 1, "val={}, weight={}", *val, *weight);
+                if *weight == 1 {
+                    w.write_fmt(format_args!("{}\n", *val))?;
```
mbudiu-vmw, line 94: Doesn't this case make life harder for everyone involved?

go/pkg/ddlog/ddlog.go
```diff
@@ -22,10 +22,10 @@ static void freeCmdArray(ddlog_cmd **ca) {
     free(ca);
 }
 
-extern void handleOutRecord(uintptr_t progIdx, table_id table, ddlog_record *rec, bool polarity);
+extern void handleOutRecord(uintptr_t progIdx, table_id table, ddlog_record *rec, ssize_t weight);
```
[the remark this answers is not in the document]
ryzhyk: It will; but at least the breakage will manifest at compile time.

java/ddlogapi/DDlogAPI.java
```diff
@@ -359,11 +358,11 @@ public int clearRelation(int relid) throws DDlogException {
     }
 
     // Callback invoked from commit_dump_changes.
-    void onDelta(int tableid, long handle, boolean polarity) {
+    void onDelta(int tableid, long handle, long weight) {
         if (this.deltaCallback != null) {
-            DDlogCommand.Kind kind = polarity ? DDlogCommand.Kind.Insert : DDlogCommand.Kind.DeleteVal;
+            DDlogCommand.Kind kind = weight > 0 ? DDlogCommand.Kind.Insert : DDlogCommand.Kind.DeleteVal;
```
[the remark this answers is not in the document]
ryzhyk: I tried that, but then realized that I still need a way to distinguish DeleteVal and DeleteKey, and potentially other commands in the future, so we need both weight and kind.

rust/template/ddlog.h
```diff
@@ -110,7 +110,7 @@ typedef void ddlog_record;
 typedef struct {
     table_id table;
     ddlog_record *rec;
-    bool polarity;
+    ssize_t weight;
```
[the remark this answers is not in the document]
ryzhyk: `ssize_t` is a signed type.

rust/template/differential_datalog/program.rs
ryzhyk: Actually, I don't see how this combination is problematic.  Inputs are always multisets, it's just that sometimes we force them to have unit multiplicities (before feeding data to differential).

This particular code is just me being unwilling to change the semantics of indexes from sets to multisets.

rust/template/differential_datalog/program.rs
ryzhyk: I know what you mean but in this particular place it's not too bad, as there will be at most one compensation command for each input command from the user.

rust/template/differential_datalog/valmap.rs
ryzhyk: This makes sure that output _sets_ are always be printed without multiplicities, which are always "+1".

Issue #649, opened 2020-04-30
Infinite loop in compiler


mbudiu-vmw:
The following program with recursive types seems to cause the compiler to go into an infinite loop;

```
typedef TSeq = Option<(string, Ref<TSeq>)>

relation Rletter(_l:string)
relation Rseq(_s:TSeq)
Rseq(Some{(_l, _s)}) :- Rletter(_l), Rseq(_s).
```

ryzhyk:
Resolved in #709

PR #709, opened 2020-07-20
Infinite recursion in type inference.

ryzhyk:
This addresses issue #649 that existed in the earlier version of the
type inference engine and was successfully replicated in the new one.

When performing type inference on recursive types, the algorithm tried
to expand type aliases proactively, leading to infinite recursion.  The
solution is to do it on-demand instead.

Once the problem was fixed, I ran into another issue--Rust does not Rust
recursive type aliases, e.g., this is ok:

```
typedef TSeq = TSeq1{x: (string, Ref<TSeq>)}
             | TSeqNone
```

but not this:

```
typedef TSeq = Option<(string, Ref<TSeq>)>
```

(the difference is that the latter is a type alias).

So I added a new check to disallow such definitions.


test/datalog_tests/lib_test.debug.ast.expected
```diff
@@ -1392,21 +1392,21 @@ internment_test::AllInternedString[(internment_test::AllInternedString{.ix=ix}:
...
+internment_test::Projections[(internment_test::Projections{.inp=internment_test::istruct2struct((internment::ival(i): internment_test::IStruct)), .p=("f1=" ++ (std::__builtin_2string(f1): string))}: internment_test::Projections)] :- internment_test::IStruct[(i@ ((&(internment_test::IStruct{.u=((&(internment_test::Tag1{.f1=(f1: bool)}: internment_test::IUnion)): internment::Intern<internment_test::IUnion>), .t=(_: internment::Intern<(std::s32, double)>), .x=(_: bigint)}: internment_test::IStruct)): internment::Intern<internment_test::IStruct>))], Inspect debug::debug_event((32'd284, 32'd0, 32'd0), ddlog_weight, ddlog_timestamp, "Map", i, (internment_test::Projections{.inp=internment_test::istruct2struct((internment::ival(i): internment_test::IStruct)), .p=("f1=" ++ (std::__builtin_2string(f1): string))}: internment_test::Projections)).
```
mbudiu-vmw, line 1397: We still have builtin2string?

Issue #707, opened 2020-07-17
RFC: `?` operator


ryzhyk:
I am thinking of adding syntactic sugar to work with `Result<>` and `Option<>` types.

DDlog does not have NULLs (NULLs are evil!).  Instead missing values are represented in a type-safe fashion using `Option` and `Result` types. The programmer must open up these types using pattern matching or unwrap them using convenience functions like `unwrap_or` and `unwrap_or_default` in order to access their inner values.

The downside of this is that the code gets bloated, especially when working with many, potentially nested, optional values.  Rust has the `?` operator that makes error handling extremely ergonomic: it unwraps a `Result`, returning from the current function on error.  Groovy has something similar, but the semantics of Groovy's `?` operator is much more confusing.

So what should the DDlog `?` operator do? I suggest an approach similar to Rust:
- `?` applies to expressions of type `Option` and `Result` only.
- if the value is `Some` or `Ok`, it just returns the inner value
- otherwise (`None` or `Err`), terminates the current function, returning a value according to the following table. 

| expression type | expression value | function return type |     return value     |
|:---------------:|:----------------:|:--------------------:|:--------------------:|
| Option< X >       | None             | T                    | default value of T   |
| Result<X,E>     | Err{e}           | Result<_,E>          | Err{e}               |
| Result<X,E>     | Err{e}           | Result<_,F>          | rejected by compiler |
| Result<X,E>     | Err{e}           | any other type T     | default value of T   |

Example
--------

```
function foo(): bool
{
   ...
    var x = match (y) {
        None -> return false,
        Some{z} -> {
             z+5
        }
    };
    ...
}
```

becomes

```
function foo(): bool
{
   ...
   var x = y? + 5;
   ...
}
```

One open question is whether `?` should be allowed outside of a function, in particular inside a rule:
- in a relation literal
- in an assignment clause
- in a filter clause

In all these cases, the behavior of `?` would still be well defined (just replace the "function return type" with expression type expected in the given context in the above table), but it may be less intuitive to the user.

@mbudiu-vmw , @blp

mbudiu-vmw:
I think it's dangerous to allow it everywhere. I would suggest by starting to only allow it in functions that already return Option or Result, there the meaning is very clear. If this is not enough to make the code readable we can talk again.
But people have to think how to handle these errors, just returning default(T) will prevent them from thinking.

ryzhyk:
[...]

PR #710, opened 2020-07-22
?-operator

ryzhyk:
See RFC #707.

We introduce Rust-style error handling operator `?`.

The `?` operator can placed after an expression that returns `Option<>` or
`Result<>` inside a function scope.  Similar to `unwrap_` functions, it
extracts the inner value on success.  On error, it returns the error value
(or `None`) from the function, e.g.:

```
/* Lookup item in the inventory and return its price in cents. */
function get_price_in_cents_(inventory: Map<string, string>, item: string): Option<u64> {
    Some{ inventory.get(item)?.parse_dec_u64()? * 100 }
}
```

is equivalent to

```
function get_price_in_cents(inventory: Map<string, string>, item: string): Option<u64> {
    match (inventory.get(item)) {
        None -> None,
        Some{price} -> match (parse_dec_u64(price)) {
                           None    -> None,
                           Some{p} -> Some{100 * p}
                       }
    }
}

```

The `?` operator can only be used inside a function whose return type is
`Option` or `Result`.  The following table summarizes the behvaior of `?`
for various combinations of expression type and function return type.

| Function return type |     Expression type    | Expression value |  ? behaves as  |
|:--------------------:|:----------------------:|------------------|:--------------:|
| Option<T1>           | Option<T2>             | None             |   return None  |
| Option<T1>           | Option<T2>             | Some{x}          |        x       |
| Option<T1>           | Result<V,E>            | Err{e}           |   return None  |
| Option<T1>           | Result<V,E>            | Ok{v}            |        v       |
| Result<X,E>          | Result<Y,E>            | Err{e}           | return Err{e}  |
| Result<X,E>          | Result<Y,E>            | Ok{v}            |        v       |
| Result<V,E>          | Option<T>              |                  |     invalid    |
| Result<X,E1>         | Result<Y,E2>, E2 != E1 |                  |     invalid    |

### DIFF
--- a/src/Language/DifferentialDatalog/TypeInference.hs
+++ b/src/Language/DifferentialDatalog/TypeInference.hs
@@ -444,6 +444,24 @@ inferTypes d es = do
                  EBreak{}    -> annotated
                  EReturn{}   -> annotated
                  ERef{}      -> annotated
+                 ETry{..} | isOption ?d inner_type && isOption ?d funcType
+                             -> E $ EMatch (pos e) inner_expr
+                                    [(eStruct nONE_CONSTRUCTOR [] inner_type, eReturn (eStruct nONE_CONSTRUCTOR [] funcType) t),
+                                     (eStruct sOME_CONSTRUCTOR [("x", eVarDecl "__x" t)] inner_type, eVar "__x")]
+                          | isResult ?d inner_type && isOption ?d funcType
+                             -> E $ EMatch (pos e) inner_expr
+                                    [(eStruct eRR_CONSTRUCTOR [("err", eTyped ePHolder inner_etype)] inner_type, eReturn (eStruct nONE_CONSTRUCTOR [] funcType) t),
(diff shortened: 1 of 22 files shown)


src/Language/DifferentialDatalog/TypeInference.hs
```diff
@@ -444,6 +444,24 @@ inferTypes d es = do
                  EBreak{}    -> annotated
                  EReturn{}   -> annotated
                  ERef{}      -> annotated
+                 ETry{..} | isOption ?d inner_type && isOption ?d funcType
+                             -> E $ EMatch (pos e) inner_expr
+                                    [(eStruct nONE_CONSTRUCTOR [] inner_type, eReturn (eStruct nONE_CONSTRUCTOR [] funcType) t),
+                                     (eStruct sOME_CONSTRUCTOR [("x", eVarDecl "__x" t)] inner_type, eVar "__x")]
```
mbudiu-vmw, line 450: Is there any danger of a name clash?

ryzhyk: There is, hence the undescores.

PR #715, opened 2020-07-24
Vector and map literals.

ryzhyk:
Introduced syntactic sugar for vector and map literals.

Vector literals:
===============

```
var v = [0, x, f(y)];
```

is equivalent to:

```
var v = vec_with_capacity(3);
v.push(0);
v.push(x);
v.push(f(y));
...
```

Map literals:
=============

Likewise, **map literals** can be used to create maps by listing their
key-value pairs:

```
var m = ["foo" -> 0,
         "bar" -> 1,
         "foobar" -> 2];
...
```

To keep the implementation simple, map and vector literals are expanded
inside the parser, so we don't need to introduce new kinds of
nodes for them in the AST.  This may lead to less clear error messages.
In addition, this means that empty vector and map literals are not
allowed, as the parser cannot tell a vector from a map.  One needs to
write `vec_empty()` or `map_empty()` for empty vectors and maps
respectively.

-----

This PR also allows variable shadowing, i.e., declaring a variable that shadows
existing variable name.  This simplifies the implementation of the new feature,
but also there is no good reason to disallow this.

### DIFF
--- a/doc/language_reference/language_reference.md
+++ b/doc/language_reference/language_reference.md
@@ -337,6 +337,8 @@ term ::= "_"                 (* wildcard *)
        | bool_literal        (* Boolean literal *)
        | fp_literal          (* floating point literal *)
        | string_literal      (* string literal *)
+       | vec_literal         (* vector literal *)
+       | map_literal         (* map literal *)
        | cons_term           (* type constructor invocation *)
        | apply_term          (* function application *)
        | var_term            (* variable reference *)
@@ -409,9 +411,11 @@ they produce results of type `string`, e.g.:
 Other terms:
 
 ```EBNF
+vec_literal  :: "[" expr ("," expr)* "]"
+map_literal  :: "[" expr "->" expr ("," expr "->" expr)* "]"
 bool_literal ::= "true" | "false"
 cons_term    ::= (* positional arguments *)
-                 cons_name ["{" [expr (,expr)*] "}"]
+                 cons_name ["{" [expr ("," expr)*] "}"]
                  (* named arguments *)
                | cons_name ["{" "." field_name "=" expr
                             ("," "." field_name "=" expr)* "}"]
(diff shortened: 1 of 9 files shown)


ryzhyk:
@mbudiu-vmw , I need to release this today, but can you please review when you have time?

doc/language_reference/language_reference.md
```diff
@@ -409,9 +411,11 @@ they produce results of type `string`, e.g.:
 Other terms:
 
 ```EBNF
+vec_literal  :: "[" expr ("," expr)* "]"
```
mbudiu-vmw, line 414: I think you should allow empty vectors in this grammar. 
The fact that the implementation may not is a bug.

src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -583,13 +583,13 @@ exprValidate1 d _ _   (ESlice p _ h l)    =
     check d (h >= l) p
           $ "Upper bound of the slice must be greater than lower bound"
 exprValidate1 _ _ _   EMatch{}            = return ()
-exprValidate1 d _ ctx (EVarDecl p v)      = do
+exprValidate1 d _ ctx (EVarDecl p _)      = do
     check d (ctxInSetL ctx || ctxInMatchPat ctx) p "Variable declaration is not allowed in this context"
-    checkNoVar p d ctx v
+    --checkNoVar p d ctx v
```
mbudiu-vmw, line 588: do you want to delete these?

test/datalog_tests/function.fail.ast.expected
```diff
@@ -18,10 +14,6 @@ expression '(var w: bool)'
         C0{.x = w: bool} -> w,
                 ^^^^^^^
 
-error: ./test/datalog_tests/function.fail.dl:7:5-7:11: Variable v already defined in this scope
```
mbudiu-vmw, line 21: should this be at least a warning?

test/datalog_tests/function.fail.dl
```diff
@@ -20,21 +20,6 @@ function vars(): () = {
...
-typedef Alt = C0{x: bit<32>}
-            | C1{x: bit<32>}
-
-// shadow argument name
-function shadow(v: string): () = {
-    var a: Alt = C0{1};
-
-    var i: bit<32> = match (a) {
-        C0{.x = v} -> v,
-        C1{v} -> v
```
mbudiu-vmw, line 32: This looks quite ambiguous. It would be nicer if you required a `var` to always be used when introducing a new var. If you just use an identifier this could be an error.

PR #723, opened 2020-07-29
lib/url.dl: Library for working with URLs.

ryzhyk:
DDlog bindings for the `url` crate.


lib/url.dl
```diff
@@ -0,0 +1,95 @@
+/* Bindings for the Rust URL library.
+ */
+
+extern type Url
+
+/* Parse an absolute URL from a string. */
+extern function url_parse(url: string): Result<Url, string>
+
+/* Parse a string as an URL, with this URL as the base URL.
```
mbudiu-vmw, line 9: `url` instead of `this`?

lib/url.dl
```diff
@@ -0,0 +1,95 @@
...
+ * Note: a trailing slash is significant. Without it, the last
```
mbudiu-vmw, line 10: trailing slash in url?

lib/url.dl
```diff
@@ -0,0 +1,95 @@
...
+ * path component is considered to be a “file” name to be removed
```
mbudiu-vmw, line 11: maybe an example would help, I find this comment quite hard to parse, although it's probably borrowed from the Rust library.

lib/url.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+impl record::Mutator<url_Url> for record::Record {
+    fn mutate(&self, t: &mut url_Url) -> result::Result<(), String> {
+        *t = url_Url::from_record(self)?;
+        Ok(())
+    }
+}
+
+pub fn url_url_parse(s: &String) -> std_Result<url_Url, String> {
```
mbudiu-vmw, line 41: I wonder whether we could generate some of this code automatically from the dl declarations.
Perhaps we could have an annotations for the ones where we can.

PR #725, opened 2020-08-05
Support query for output record derivation trace.

yjiayu:
1. Add maps to store all debug event entries and extract output record and its input record
   and store these information.
2. Add helper function to get predecessor operatorId for input record from syntax tree
3. Add tree-like 'DebuggerRecordNode' structure to store all possible derivation results
   for a given output record with operator id.


debugger/Main.hs
```diff
@@ -21,19 +21,126 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
...
+parseProgram Config{..} = do
+    fdata <- readFile confDatalogFile
+    (d, _, _) <- parseDatalogProgram (takeDirectory confDatalogFile:confLibDirs) True fdata confDatalogFile
+    d'' <- case confOutputInternal of
+        False -> return d
+        True ->  return $ progOutputInternalRelations d
+    d''' <- case confOutputInput of
+         "" -> return d''
+         x  ->  return $ progMirrorInputRelations d'' x
+    when confDumpFlat $
```
haroldlim, line 123: I don't think we need to add this option to dump the transformed program in the debugger program. If we want to dump it, we can always use the ddlog binary to do it.

debugger/Main.hs
```diff
@@ -21,19 +21,126 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
...
+        operatorInput = InputOp operatorId
+        dbgRecord = DebuggerRecord {dbgRecord=outputRecord, dbgOperatorId=operatorInput}
+        dbgRecordNodes = queryDerivations dbgRecord dgbRecordMap
+    in trace (show dbgRecordNodes) $ queryAll events dgbRecordMap
+
+parseProgram :: Config -> IO (DatalogProgram)
+parseProgram Config{..} = do
+    fdata <- readFile confDatalogFile
+    (d, _, _) <- parseDatalogProgram (takeDirectory confDatalogFile:confLibDirs) True fdata confDatalogFile
+    d'' <- case confOutputInternal of
```
haroldlim, line 117: This means that we have to ensure that we pass the same options when we compiled the ddlog program and  here.
Instead of loading the original program here (and ensuring the same options are passed), I wonder if we can just load the dumped transformed source here? I haven't thought about it deeply yet, so not sure if it's possible to generate DatalogProgram from an ast file?

src/Language/DifferentialDatalog/Debugger/DebugState.hs
```diff
@@ -0,0 +1,179 @@
+{-
+Copyright (c) 2018-2020 VMware, Inc.
```
haroldlim, line 2: Not sure what the convention is for new files, but usually the year here should be the year it was added (i.e., just 2020)

debugger/Main.hs
```diff
@@ -21,19 +21,126 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
...
+           Just(p) -> addOption config (LibDir $ p ++ "/lib")
+           _       -> return config
+       do
+            datalogProg <- parseProgram config'
+            events <- parseEventFromDumpFile config'
+            let recordMap = handleDebugEvents events emptyDebuggerMaps datalogProg
+                s = queryAll events (dbgRecordMap recordMap)
+            putStr $ show s
+
+queryAll :: [Event] -> DebuggerRecordMap -> String
```
haroldlim, line 103: So, currently the debugger tool will just print/query all derivations of all output records?

Not required for this change, but I wonder if we can make it more interactive (i.e., list out the output records, and then we can specify which output record we want to trace the derivation).

yjiayu: Yeah, for now the debugger tool's function is 
queryDerivations :: DebuggerRecord -> DebuggerRecordMap -> DebuggerRecordNode
It takes one output record and debuggerRecordMap (got built when parsing log events) and return the root node of the derivations since the derivation result is a tree like structure.
According to Leonid, we are planning to design a UI that will take the debugger output and then display it in a better way, probably is a next step work.

PR #729, opened 2020-08-08
Use pre-transformed program's rhs index when computing operator id

haroldlim:
When injecting debugging hooks, the aggregate function is
transformed into 2 terms. Using the post-transformed program
to get the rhs index for operator id computation causes issues
with the debugger tool to correlate events to source.
Instead, when transforming the program, the original indices are
maintained and then used when computing operator id.


haroldlim:
I think we can just merge this to master directly. @yjiayu  can then rebase his change on top of the latest in master.

PR #734, opened 2020-08-12
Add support for exporting timely events as CSV file.

gatoWololo:
Here is the working initial version of the timely logging infrastructure code. I'll make a few comments in the code below for questions I have about how things should be done...


rust/template/differential_datalog/Cargo.toml
```diff
@@ -16,6 +16,7 @@ ordered-float = {version = "2.0.0", features = ["serde"]}
 fnv = "1.0.2"
 timely = "0.11"
 libc = "0.2"
+csv = "1.1"
```
gatoWololo, line 19: I don't know how you feel about adding more dependencies to the project? We can also consider other file formats if CSV seems to verbose or slow.

rust/template/differential_datalog/program.rs
```diff
@@ -1179,14 +1181,21 @@ impl Program {
...
+                            TimelyEvent::Progress(_) |
+                            TimelyEvent::PushProgress(_) => profcpu.load(Ordering::Acquire),
                             _ => false
-                        }).map(|x|(x, get_prof_context())).collect();
+                        }).collect();
+
                         if !filtered.is_empty() {
                             //eprintln!("timely event {:?}", filtered);
-                            prof_send1.send(ProfMsg::TimelyMessage(filtered.drain(..).collect())).unwrap();
+                            // Context doesn't change within a batch, so attach once.
```
gatoWololo, line 1196: Before, the context String was attached to every message. From what I can tell, this string doesn't change at the "batch" granularity. Is this true? So I only include it once per batch (this allows the profile_statistics code to use the same vector as the existing profiling code `Vec<(Duration, usize, TimelyEvent)>`. Mainly:
```diff
- TimelyMessage(Vec<((Duration, usize, TimelyEvent), String)>)
+ TimelyMessage((Vec<(Duration, usize, TimelyEvent)>, String))

File: rust/template/differential_datalog/program.rs [hunk 3]
```diff
@@ -1618,9 +1627,10 @@ impl Program {
 
     /* Profiler thread function */
     fn prof_thread_func(chan: mpsc::Receiver<ProfMsg>, profile: Arc<Mutex<Profile>>) {
+        let mut stats = Statistics::new("stats.csv").expect("TODO");
```
gatoWololo, line 1630: How should I handle failure to create this file? Currently I panic on failure. Should I instead continue executing without writing statistics and print a warning?

Should the file be named this or have a different name?

gatoWololo, line 1630: This thread seems to be spawned whether or not profile_cpu is on? And it merely blocks on `recv()` forever if profiling isn't on. This means that the `stat.csv` file is always created but without cpu profiling it is empty. Is this okay? I can change the logic so the empty file isn't created when logging is off.

rust/template/differential_datalog/profile_statistics.rs
```diff
@@ -0,0 +1,301 @@
+use std::sync::mpsc::{Receiver, Sender, SyncSender};
```
gatoWololo, line 1: This file needs to have the comments and documentation revised before merging by the way.

gatoWololo:
(I know why the build is failing btw: I forgot to include a change to one of the Haskell files...)

rust/template/differential_datalog/Cargo.toml
ryzhyk: I've no problem with adding this dependency.

rust/template/differential_datalog/profile_statistics.rs
```diff
@@ -0,0 +1,301 @@
...
+                }
+            }
+        }
+    }
+}
+
+type LogEvents = Vec<(Duration, usize, TimelyEvent)>;
+type MessageBatch = (Duration, LogEvents);
+
+pub fn record_events_csv(path: &str) -> csv::Result<(Arc<Mutex<Sender<MessageBatch>>>, JoinHandle<()>)> {
```
ryzhyk, line 283: where is this function used?

rust/template/src/api.rs
```diff
@@ -494,6 +500,17 @@ impl HDDlog {
         }
     }
 
+    fn record_enable_timely_profiling(&self, enable: bool) {
+        if let Some(ref f) = self.replay_file {
+            let _ = f.lock().unwrap().record_timely_profiling(enable).map_err(|_| {
+                self.eprintln(
+                    "ddlog_cpu_profiling_enable(): failed to record invocation in replay file",
```
ryzhyk, line 507: cpu -> timely

rust/template/src/api.rs
```diff
@@ -168,6 +168,12 @@ impl HDDlog {
         self.prog.lock().unwrap().enable_cpu_profiling(enable);
     }
 
+    pub fn enable_timely_profiling(&self, enable: bool) {
```
ryzhyk, line 171: you also need to declare `ddlog_enable_timely_profiling()`, by analogy with `ddlog_enable_cpu_profiling()` in this function.

PR #752, opened 2020-08-25
Worked on std

Kixiron:
Fixed some clippy lints on the std's rust code

### DIFF
--- a/lib/internment.rs
+++ b/lib/internment.rs
@@ -1,7 +1,5 @@
-use arc_interner;
 use differential_datalog::record;
 use differential_datalog::record::*;
-use serde;
 use std::cmp;
 use std::fmt;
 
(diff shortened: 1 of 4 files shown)


lib/internment.rs
```diff
@@ -54,7 +54,13 @@ impl<A: Eq + Hash + Send + Sync + 'static> internment_Intern<A> {
             intern: arc_interner::ArcIntern::new(x),
         }
     }
-    pub fn as_ref(&self) -> &A {
+}
+
+impl<A> AsRef<A> for internment_Intern<A>
```
ryzhyk, line 59: I was a bit paranoid about doing this.  This type is designed for use in auto-generated code where we usually don't want any magic conversions.  I cannot think of a specific scenario where this would cause problems though, so I guess it's ok.

Kixiron: It really shouldn't matter since `AsRef` is in the prelude, worst comes to worst we can use qualified syntax with `AsRef::as_ref(thing)` instead of `thing.as_ref()`

PR #754, opened 2020-08-27
Improved static variable generation

Kixiron:
* ~~Collected program statics are now nested within a single `lazy_static!` invocation (theoretically reduces compile times)~~
* Relation statics were refactored into one core function
* Relation statics now have their hashmaps pre-allocated, should reduce memory usage/churn and make startups quicker (affects large programs more)
* C-Name maps now store `&'static CStr`s instead of `CStrings`. This brings a *huge* memory benefit since now names are only baked into the binary instead of being uselessly heap-allocated
* Changed the C-Name maps to panicking instead of silently returning a corrupted/dummy value, a panic only in the instance of something being *terribly* wrong is better than silent state corruption and un-huntable bugs
* Docs! Both the generated relation statics have (kinda bad) docs and (most) of the functions that create them have 'em as well

### DIFF
--- a/src/Language/DifferentialDatalog/Compile.hs
+++ b/src/Language/DifferentialDatalog/Compile.hs
@@ -624,13 +624,19 @@ collectStatics d = execState (progExprMapCtxM d (checkStaticExpr)) M.empty
 -- computed by 'collectStatics'.
 mkStatics :: DatalogProgram -> Statics -> Doc
 mkStatics d statics =
-    vcat $
-    map (\((e, t), i) ->
-            let static_name = "__STATIC_" <> pp i in
-            let ?statics = deleteStatic e t statics in
-            let e' = mkExpr d CtxTop (eTyped e t) EVal in
-            "lazy_static!{ pub static ref" <+> static_name <> ":" <+> mkType t <+> "=" <+> e' <+> "; }")
-        $ M.toList statics
+    "lazy_static! {"
+    $$ (nest' $ vcat $ static_decls)
+    $$ "}"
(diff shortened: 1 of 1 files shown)


src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -624,13 +624,19 @@ collectStatics d = execState (progExprMapCtxM d (checkStaticExpr)) M.empty
...
 mkStatics :: DatalogProgram -> Statics -> Doc
 mkStatics d statics =
-    vcat $
-    map (\((e, t), i) ->
-            let static_name = "__STATIC_" <> pp i in
-            let ?statics = deleteStatic e t statics in
-            let e' = mkExpr d CtxTop (eTyped e t) EVal in
-            "lazy_static!{ pub static ref" <+> static_name <> ":" <+> mkType t <+> "=" <+> e' <+> "; }")
-        $ M.toList statics
+    "lazy_static! {"
```
ryzhyk, line 627: Unfortunately, this does not work.  I don't remember the exact error, but basically rustc failed when there were too many static variables within a single `lazy_static`.

Kixiron: That's really bizarre, and rather unfortunate. I revered the change, should be good now

ryzhyk:
@Kixiron , have you seen my DM in gitter?

Kixiron:
Corruption wasn't the best term, I was referring to the `.unwrap_or_else()`s that returned new `CString`s in the event of an error

PR #755, opened 2020-08-28
Switch from lazy_static to once_cell

Kixiron:
* Switches all generated static variables to use [`once_cell::sync::Lazy`](https://docs.rs/once_cell/1.4.1/once_cell/sync/struct.Lazy.html) instead of `lazy_static!`

#### Rationale:
`once_cell` is soon to be [part of the standard library](https://github.com/rust-lang/rfcs/pull/2788) (and is available now [on nightly](https://doc.rust-lang.org/nightly/std/lazy/struct.SyncLazy.html)). That allows a lighter dependency load, in addition to a lighter compile-time load now (due to less macros being expanded) and in the future (when it's in the std, that's one less dep to compile). `once_cell` allows greater flexibility due to having sync and unsync variants of the `Lazy` type and the `OnceCell` type. Usage of `OnceCell` can allow what would otherwise be circular dependencies within statics, which could be useful in the future. `once_cell` is also marginally faster at runtime, but I doubt we'd be able to see the effects of that.

### DIFF
--- a/lib/intern.rs
+++ b/lib/intern.rs
@@ -1,6 +1,6 @@
 use differential_datalog::record;
 use differential_datalog::record::*;
-use lazy_static::lazy_static;
+use once_cell::sync::Lazy;
 use serde;
 use std::collections;
 use std::marker;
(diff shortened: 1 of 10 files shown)


ryzhyk:
@Kixiron, check out `../../tools/prepush.sh.  If you symlink your `../../.git/hooks/pre-push` to it, it will run clippy + rustfmt on every push.  The `install-dependencies.sh` script does this automatically.

ryzhyk:
Looks good, modulo lint failures.
@Kixiron , let me know if you are still working on this; otherwise I'll make the fixes and merge.

Kixiron:
I believe I've fixed them all, but gitlab won't update with my changes

PR #758, opened 2020-09-01
Allow a semicolon after a sequence of expressions.

ryzhyk:
We used to disallow a semicolon after a sequence of expressions, e.g.,
`{ e1; e2; }` was illegal and had to be written as either
`{ e1; e2 }` (to return the value of `e2`) or `{ e1; e2; ()}`
to discard the result of `e2` and return an empty tuple.

This annoyed most users, including yours truly.  So we now allow
`{ e1; e2; }`, which gets desugared into `{ e1; e2; () }`.

Along the way, I fixed a bug in type inference when the type of
`continue`, `break`, or `return` is not constrained by the environment.
Instead of crashing, we now assign unit type to them.

### DIFF
--- a/doc/tutorial/tutorial.md
+++ b/doc/tutorial/tutorial.md
@@ -734,7 +734,15 @@ function addr_port(ip: ip_addr_t, proto: string, preferred_port: bit<16>): strin
 
 The result computed by a function is the value of the last expression evaluated
 or the value produced by the first `return` statement encountered when
-evaluating the function.  If the `else` is missing the value `()` (empty tuple)
+evaluating the function.  A semicolon at the end of a sequence of expressions
+discards the value of the last expressions and produces an empty tuple.  For
+instance, expressions `{ var x: u32 = y + 1; x }` and
+`{ var x: u32 = y + 1; x; }` have different meaning.  The former returns the
+32-bit value of variable `x`, whereas the latter is equivalent to
+`{ var x: u32 = y + 1; x; () }`, which produces an empty tuple `()`,
+described [below](#tuples).
+
+If the `else` clause of an `if-else` expression is missing the value `()`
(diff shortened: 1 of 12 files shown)


doc/tutorial/tutorial.md
```diff
@@ -734,7 +734,15 @@ function addr_port(ip: ip_addr_t, proto: string, preferred_port: bit<16>): strin
 
 The result computed by a function is the value of the last expression evaluated
 or the value produced by the first `return` statement encountered when
-evaluating the function.  If the `else` is missing the value `()` (empty tuple)
+evaluating the function.  A semicolon at the end of a sequence of expressions
```
mbudiu-vmw, line 737: I am not sure this is a good idea. I think it will trip a lot of people who come from a C-like language.
I think it should be a warning.

src/Language/DifferentialDatalog/Syntax.hs
```diff
@@ -799,7 +799,7 @@ instance PP e => PP (ExprNode e) where
                                $$
                                "}"
     pp (EVarDecl _ v)        = "var" <+> pp v
-    pp (ESeq _ l r)          = parens $ (pp l <> semi) $$ pp r
+    pp (ESeq _ l r)          = braces $ (pp l <> semi) $$ pp r
```
mbudiu-vmw, line 802: the sql compiler may need to change because of this too.

doc/tutorial/tutorial.md
ryzhyk: There is no room for confusion here.  If the expression is expected to return a type that is not an empty tuple, you will get a type error.  If it returns `()`, then discarding it is a no-op.

mbudiu-vmw: the type error may be confusing, especially if it involves some form of unification.

PR #763, opened 2020-09-12
New Rust module structure.

ryzhyk:
This commit changes the structure of the generated Rust project in order
to improve ergonomics and speed up compilation.

We used to place all function and type declarations in a single module
in `types/lib.rs`.  This was suboptimal for several reasons.  First,
we had to rely on name mangling for namespace isolation, e.g., `None`
would get compiled to `std_Some::std_None` (and even that was not
guaranteed to avoid conflicts altogether).  This made writing extern
functions and types, as well as using the Rust API cumbersome.  Second,
since all declarations were placed in the same module, it was very easy
to run into name clashes with manually written Rust code.  Third,
since the entire module is re-generated and re-compiled every time, no
incremental compilation was possible.

We change this structure by generating a Rust module hierarchy that
precisely matches the DDlog hierarchy, with one Rust module for each
DDlog module containing all type and functiond declarations, as well as
all extern Rust code for the module.

This addresses the first two problems above and makes a step towards
solving the third problem.  Compilation is not yet incremental; however
it is not in principle possible to make the DDlog compiler incremental
by only re-compiling affected modules.  In order to also compile Rust
code incrementally, we would have to further break up the generated code
into multiple crates, which will be much easier with the current
structure.

One side effect of this refactoring is that I had to rename the `std`
library to `ddlog_std` to avoid name clashes with the Rust `std` library
that gets automatically imported by all programs.  I tried different
techniques to avoid the clashes, but there does not seem to exist a
robust way.

----------

This commit makes another change that helps speed up compilation.  This
should have been done in a separate commit, but the two ended up being
too tangled to separate them.

We used to implement `Serialize` and `Deserialize` traits for the
`DDValue` type using `serde_typetag` crate that uses some black magic to
auto-generate serialization/deserialization code for trait objects.
Turns out this magic is extremely expensive in terms of compilation time
and generated code size.  It literally slows down compilation by a
factor or two for some of our programs.  I don't have an explanation and
filed an issue against typetag: dtolnay/typetag#25.

As a workaround, we now generate serialization/deserialization code
in the DDlog compiler, and it's much more lightweight.  Note that
`DDValue` serialization/deserialization is only used by the
distributed-ddlog crate.  More specifically, it needs to
serialize/deserialize `Update<DDValue>` objects.  Since these objects
contain relation id for each value, we can implement a custom
deserializer that first reads relations id and picks the value type to
deserialize based on that.  See details in
`src/lib.rs/UpdateSerializer`.


lib/ddlog_std.dl
```diff
@@ -6,10 +6,15 @@
 import debug
 import log
 
+#[alias]
```
mbudiu-vmw, line 9: is this documented? should users care?

rust/template/types/flatbuf.rs
```diff
@@ -1,9 +1,11 @@
 //! Serialize DDlog types to/from FlatBuffers.
 
 use super::*;
+use ::ordered_float::OrderedFloat;
```
mbudiu-vmw, line 4: how come these are needed now but not before?

src/Language/DifferentialDatalog/Attribute.hs
```diff
@@ -197,6 +203,23 @@ tdefGetSharedRefAttr d tdef =
          Left e  -> error e
          Right b -> b
 
+{- 'alias' attribute: Tells DDlog not to generate Rust declaration for the type
```
mbudiu-vmw, line 206: this may be useful for ddlog programmers too, so perhaps it should be documented.
but if it only applies to scalar types the value could be limited?

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -343,26 +349,37 @@ emptyCompilerState = CompilerState {
     cArrangements = M.empty
 }
 
--- Convert name to a valid Rust identifier by replacing "::" with "_"
-rname :: String -> Doc
-rname = pp . replace "::" "_"
+-- Relations, indexes, relation transformers and Value's are stored in the flat namespace.
+-- Flatter a name by replacing "::" with "_"
```
mbudiu-vmw, line 353: flatten?

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -440,15 +457,20 @@ tuple :: [Doc] -> Doc
 tuple [x] = x
 tuple xs = parens $ hsep $ punctuate comma xs
 
-tupleTypeName :: [a] -> Doc
-tupleTypeName xs = "tuple" <> pp (length xs)
+tupleTypeName :: Bool -> [a] -> Doc
+tupleTypeName local xs =
+    tupleTypeName' (if local then "crate" else "::types") xs
```
mbudiu-vmw, line 462: create a function `localPrefix :: bool -> string` for the expression in parenthesis?

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -457,27 +479,32 @@ isStructType TStruct{..} | length typeCons == 1 = True
...
+       then rnameScoped local tname
+       else rnameScoped local tname <> "::" <> nameLocal c
 
 -- | Create a compilable Cargo crate.  If the crate already exists, only writes files
 -- modified by the recompilation.
 --
 -- 'specname' - will be used as Cargo package and library names
 --
--- 'rs_code' - additional Rust code to be added to the generated program 'lib.rs'.
+-- 'modules' - list of modules in the original DDlog program. This function
```
mbudiu-vmw, line 495: I don't understand this English, can you rephrase?

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -540,28 +568,46 @@ compile d_unoptimized specname rs_code toml_code dir crate_types = do
 
 -- | Compile Datalog program into Rust code.
 --
--- Returns Rust code for three crates that comp:
+-- Returns Rust code for three crates:
```
mbudiu-vmw, line 571: three? The first seems to be a map with one per module.

mbudiu-vmw (approved):
this is actually quite nice.

src/Language/DifferentialDatalog/Compile.hs
ryzhyk: A bit of flattery cannot harm.

src/Language/DifferentialDatalog/Attribute.hs
ryzhyk: I haven't thought the implications through, so it's probably better to reserve this for internal use only.  This is only needed for types that conflict with Rust builtins. There is a small and finite number of those.

src/Language/DifferentialDatalog/Compile.hs
ryzhyk: There's one module per DDlog module, but all of them live in the same crate (for now).

PR #765, opened 2020-09-14
Nested ts

ryzhyk:
This option is useful in recursive programs where the recursive fragment of the
program may require more than `2^16` iterations.  Since in practice most
recursive computations perform fewer iterations, DDlog uses a 16-bit iteration
counter (or timestamp in the Differential Dataflow terminology) by default.
When this is not sufficient, the `nested-ts-32` can be used to tell DDlog to use
32-bit iteration counter instead.

Consider the following example that enumerates all numbers from `0` to `65599`
and stores the last 70 numbers in an output relation:

```
relation Rb(x: u32)
output relation Rque(x: u32)

Rb(0).
Rb(x+1) :- Rb(x), x < 65599.
Rque(x) :- Rb(x), x > 65530.
```

The recursive fragment of this program performs `65600` iterations; hence a
16-bit counter is not sufficient.

If the DDlog project was compiled without the `nested-ts-32` flag, 32-bit
timestamps can be enabled when compiling the generated Rust project via
the `nested_ts_32` feature:

```
cargo build --release --features "nested_ts_32"
```


mbudiu-vmw (approved):
you should crash on timestamp overflow. I don't see where you do that.

ryzhyk:
> you should crash on timestamp overflow. I don't see where you do that.

This logic is in timely. Both `TS` and `u32` implement the `PathSummary` trait that tells timely when the maximal timestamp value has been reached.  It appears, timely chooses to just drop updates that are going to overflow rather than crash. I'll try to figure out how to change its behavior, but this is orthogobal to this PR that simply lets you choose a different TS size.

Issue #669, opened 2020-05-16
RFC: Compiler support for closures.


ryzhyk:
Closure types:

```
// Variable whose type is closure.
var closure1: |u32, Option<string>|:  u32

// parens can be used for better readability.
var closure2: (|u32, Option<string>|:  u32)

// Alias to a closure type.
typedef Closure<'X> = |'X|: bool
typedef StringClosure = Closure<string>

/*  Some higher-order functions. */
fn map(v: Vec<'X>, f: |'X|: 'Y): Vec<'Y>
fn any(v: Vec<'X>, f: |'X|: bool): bool
fn find(v: Vec<'X>, f: |'X|: bool): Option<'X>

/* A closure can modify its mutable arguments */

// Apply `f` to all elements of a JSON document recursively.
fn map(doc: JsonValue, f: |mut JsonValue|: ())
```

Creating closures:

```
// Closure type inferred from context
|x, y| {
    x + if (y.is_none()) {0} else {5}
}

// The user may type-annotate all or some of the arguments.
|x, y: Option<string>| {
    x + if (y.is_none()) {0} else {5}
}

// Or the returned type.
|x, y: Option<string>|: u32 {
    x + if (y.is_none()) {0} else {5}
}

{
    var json_doc = .....;
    // New attribute to add to each JSON map in the document.
    var new_attr = "bar";
    // When type annotations are provided, the `mut` attribute must be specified too.
    json_doc.map(|v: mut JsonValue| match (v) {
        // The closure can access variables visible in the local scope in read-only mode.
        // Variables are captured using copy semantics, i.e., all local variables accessed by
        // the closure are cloned and stored with the closure.
        JsonMap{m} -> m.insert("foo", JsonString{new_attr}),
        _ -> ()
    })
}
```

@mbudiu-vmw , @blp

mbudiu-vmw:
I think you should call them lambdas, not closures.
Also, I really don't like the Rust syntax. I like the C# and TypeScript syntax much better
`(a: type, b: type) => { body }`
or
`(a: type, b: type) => expression`
The type of a lambda is written as `(a: type) => resultType`. Here we have to be careful; in typescript I think they made the wrong choice, since in `a => resultType` the identifier `a` is not a type, but an argument name! I think it should be the opposite.
In typescript you can also write function types as `function(a: type): resultType`.
See https://www.typescriptlang.org/docs/handbook/functions.html

ryzhyk:
[...]

Issue #663, opened 2020-05-14
Custom `Ord` trait for DDlog values.


ryzhyk:
The default `Ord` implementation for `Intern` relies on pointer comparison and is therefore non-deterministic across DDlog runs.  This is necessary to speedup sorting inside DD. We still want DDlog outputs to be deterministic, though, which requires a different, less efficient, but predictable comparison trait.  DDlog expressions should be evaluated using this trait. DDlog sets should also use this trait to order things internally.

PR #772, opened 2020-10-02
Support for lambda expressions and closures.

### LINKED ISSUE #768: Parser reports unknown function when it's an unknown overload
This is surprising at first. Example:

```
$ cat foo.dl
function a(x: bit<16>, y: bit<16>): bit<16> { x + y }
function b() { a(0) }
$ ddlog -i foo.dl -L ~/differential-datalog/lib/
ddlog: error: foo.dl:2:16-2:21: Unknown function 'a'
function b() { a(0) }
               ^^^^^
```

ryzhyk:
Implements RFC #669.
Also fixes #768 along the way.

See new tutorial section on closures for more details.

Language extensions
-------------------

* Extended the type system with function types:
  ```
  typedef usize_to_string: function(x: usize): tring
  ```

* Lambda expressions
  ```
  var captured = 10;
  var f = |x: usize| x * captured;
  ```
  A lambda expression returns a closute at runtime (note the use of a captured variable).

* Function application syntax.  We used to parse function invocations as
  `<function_name>(<expr>*)`.  This has been changed to `<expr>(<expr*>)`,
  where the first expression must return a value of function type and
  can be (1) function name, (2) lambda expression, (3) any other
  expression that returns closure.

  ```
  // function call.
  f(x);

  // invoke a closure.
  var closure = |a| a * a;
  closure(x);
  ```

* Generalized polymorphic functions.  Although this is not strictly
  required for the above features, this PR also improves support for
  ad hoc polymorphism.  Instead of using the first argument to
  disambiguate functions with the same name, we now use the combination
  of all arguments and the return type.

Libraries
---------

Now that we have closures, we can implement various standard
higher-order functions over containers: map, filter, flatmap, find, all,
etc.  These now live in three new libraries that will keep growing.

* `vec.dl`

* `set.dl`

* `map.dl`


doc/language_reference/language_reference.md
```diff
@@ -426,6 +432,9 @@ vardecl_term ::= "var" var_name
 
 match_term   ::= "match" "(" expr ")" "{" match_clause (,match_clause)*"}"
 match_clause ::= pattern "->" expr
+
+lambda_term  ::= "function" "(" [expr ("," expr)*] ")" [":" simple_type_spec] expr
```
mbudiu-vmw, line 436: no braces required here?

lib/map.dl
```diff
@@ -0,0 +1,150 @@
+/*
+ * Functions for working with maps (see basic map operations defined in
+ * `ddlog_std.dl`).
+ */
+
+/* Applies closure `f` to each value in the map. */
+function map(m: Map<'K,'V1>, f: function('V1): 'V2): Map<'K,'V2> {
+    var res = map_empty();
```
mbudiu-vmw, line 8: how about a constructor with size hint?

lib/map.rs
```diff
@@ -0,0 +1,11 @@
+use crate::closure::Closure;
+use crate::ddlog_std;
+
+pub fn map_map_in_place<K: Ord, V>(
+    m: &mut ddlog_std::Map<K, V>,
+    f: &Box<dyn Closure<(*const K, *mut V), ()>>,
```
mbudiu-vmw, line 6: Do we document somewhere the relationship between ddlog types and rust types for extern function writers?

lib/vec.dl
```diff
@@ -0,0 +1,152 @@
...
+/* Sorts the vector with a key extraction function, but may not preserve the
+ * order of equal elements. */
+function sort_by(v: mut Vec<'A>, f: function('A): 'B) {
+    vec_sort_by(v, f)
+}
+
+/* Returns the element that gives the minimum value from the specified function.
+ * If several elements are equally minimum, the first element is returned.
+ * If the vector is empty, `None` is returned. */
+function min_by(v: Vec<'A>, f: function('A): 'B): Option<'A> {
```
mbudiu-vmw, line 34: arg_min?

lib/vec.dl
```diff
@@ -0,0 +1,152 @@
...
+        }
+    };
+    if (del > 0) {
+        v.truncate(len - (del as usize));
+    }
+}
+
+/* Returns `true` iff all elements of the vector satisfy predicate `f`. */
+function all(v: Vec<'A>, f: function('A): bool): bool {
+    for (x in v) {
```
mbudiu-vmw, line 107: in the future we'll just have an Iterable trait...

lib/vec.rs
```diff
@@ -0,0 +1,20 @@
+use crate::closure::Closure;
+use crate::ddlog_std;
+
+pub fn vec_sort_by<A, B: Ord>(v: &mut ddlog_std::Vec<A>, f: &Box<dyn Closure<*const A, B>>) {
+    v.x.sort_unstable_by_key(|x| f.call(x))
```
mbudiu-vmw, line 5: if there is a stable sort, why not expose that one too?

rust/template/types/closure.rs
```diff
@@ -0,0 +1,412 @@
...
+            self.f as *const (fn(Args, &Captured) -> Output) as usize,
+            &self.captured as *const Captured as usize,
+        )
+    }
+
+    fn eq_dyn(&self, other: &dyn Closure<Args, Output>) -> bool {
+        /* Compare function pointers.  If equal, it is safe to compare captured variables. */
+        let (other_f, other_captured) = other.internals();
+        if (other_f == (self.f as *const (fn(Args, &Captured) -> Output) as usize)) {
+            unsafe { *(other_captured as *const Captured) == self.captured }
```
mbudiu-vmw, line 104: So you compare pointer values?
Isn't there a safe way to do this?

mbudiu-vmw: I wonder whether this will cause problems in a distributed settings. I think that we have an invariant so far that the ordering/hash/equality of values on different machines is the same.

rust/template/types/closure.rs
```diff
@@ -0,0 +1,412 @@
...
+
+impl<'de, Args: 'static + Clone, Output: 'static + Clone> Deserialize<'de>
+    for Box<dyn Closure<Args, Output>>
+{
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        Err(D::Error::custom(
+            "Deserialization of closures is not implemented.",
```
mbudiu-vmw, line 241: I guess we won't worry about moving closures between machines then.

mbudiu-vmw (commented):
I did about 1/3, I will continue later.

doc/language_reference/language_reference.md
ryzhyk: Good point. I don't think they are strictly necessary, but we may want to requite them for consistency.

lib/map.dl
ryzhyk: Maps and sets don't have one. Only vectors do.

lib/map.rs
ryzhyk: Not at the moment, but we should.

lib/vec.dl
ryzhyk: Speaking of which, we don't have a way to implement lazy iterators at the moment (as they require either lifetimes or reference counting), so for example, if we have  `e.filter(..).all(..)`, then `filter` will iterate to the end of the collection even if `all` can terminate early.

lib/vec.rs
ryzhyk: We can also have `vec_sort_by_stable`.  Stable sort is less efficient though.

rust/template/types/closure.rs
ryzhyk: > Isn't there a safe way to do this?
Not if pointers can point to functions of different types, which is the case here, because even closures of the same type can differ in the type of their captured values.

> I wonder whether this will cause problems in a distributed settings. I think that we have an invariant so far that the ordering/hash/equality of values on different machines is the same.

It's not a problem for closures, as closures cannot be sent across the network atm (we'd have to parse and compile DDlog code on the fly).  But this may be a problem, e.g., for interned values whose order will definitely change as you send them across the network. #663 would address this problem.

lib/map.dl
mbudiu-vmw: that seems a misdesign.

lib/vec.dl
mbudiu-vmw: I think it should be possible to do that, but I don't know enough about Rust to see the potential difficulties.
Each element should have a reference in exactly one iterator, so I imagine it should be possible to do that.

lib/map.dl
ryzhyk: You cannot easily pre-allocate memory for a map or set as they are backed by trees.  I don't think I've ever seen a library in any language that does that.

lib/vec.dl
ryzhyk: Example: `v.map(...).find(..)`  

The return value of `map` is `Vec`.  DDlog returns by-value, so we need to create a fully populated vector before passing it to the next transformer (`find`).

Let's say, we change `map` to return `Iterator<>` instead. The iterator object would internally hold a reference to the original vector `v`.  This is safe as long as the iterator is destroyed before anyone tries to modify or destroy `v`, which is the case here. But not in this code:

```
var iter = v.map();
v.push(x);
iter.find(...);
```

In Rust this cannot happen thanks to the borrow checker.  In fact, the above code will be rejected by the Rust compiler.  But DDlog does not currently have a way to track ownership.

mbudiu-vmw: perhaps you should be able to obtain an iterator only from an immutable vector. this is a reasonable constraint.

ryzhyk: We'll need to add a concept of mutable variables  for that, but this still does not address the lifetime issue.  The iterator can outlive the vector it was create from. Without reference counting or a borrow checker we cannot support this safely.

mbudiu-vmw: how about an iterator that actually consumes the vector?

ryzhyk: That might work, but again this is a pretty big change to the language (we'd have to annotate functions to consume their arguments and enforce that they are not used afterwards). And then you don't always want to consume the argument.  Point is, you cannot easily retrofit a new ownership model in the language.  There is a lot of complexity in Rust that takes care of that.

src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -557,11 +557,9 @@ exprValidate1 _ _ ctx EVar{..} | ctxInRuleRHSPositivePattern ctx
                                           = return ()
 exprValidate1 d _ ctx (EVar p v)          = do _ <- checkVar p d ctx v
                                                return ()
-exprValidate1 d _ _   (EApply p fnames as) = do
-    fs <- concat <$> mapM (\fname -> checkFuncs p d fname (length as)) fnames
-    check d (length fs > 0) p $ "Unknown function '" ++ (nameLocalStr $ head fnames) ++ "'"
-    check d (length fs == 1 || (not $ null as)) p $ "Ambiguous function name '" ++ (nameLocalStr $ head fnames) ++ "' may refer to:\n  " ++
-                                                    (intercalate "\n  " $ map (\f -> name f ++ " at " ++ spos f) fs)
+exprValidate1 _ _ _   EApply{}            = return ()
```
mbudiu-vmw, line 560: why isn't this checked anymore?

mbudiu-vmw: I am guessing this moved to typechecking

src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -684,9 +680,22 @@ exprValidate3 d ctx e@(EMatch _ x cs) = do
...
-    let fun = applyExprGetFunc d ctx e
-    mapM_ (\(a, mut) -> when mut $ checkLExpr d ctx a)
-          $ zip as (map argMut $ funcArgs fun)
+exprValidate3 d ctx e@(EApply _ f as)    = do
+    let ft = exprType' d (CtxApplyFunc e ctx) f
+    mapIdxM_ (\(a, mut) i -> when mut $ checkLExpr d (CtxApplyArg e ctx i) a)
+             $ zip as (map atypeMut $ typeFuncArgs ft)
 
+{-
+-- We currently allow converting functions that return values by-reference into
```
mbudiu-vmw, line 689: how about adding a warning function and then calling it?

test/datalog_tests/vec_test.dl
```diff
@@ -0,0 +1,125 @@
...
+IntVecTransforms("not in [-5, 0, 7]", {
+                     var res = int_vec();
+                     var exclude = [-5, 0];
+                     exclude.push(7);
+                     res.retain(|x| is_none(exclude.find(|y| y==x )));
+                     res
+                 }).
+
+// all.
+IntVecTransforms("not in [-5, 0, 100] uaing 'all'", {
```
mbudiu-vmw, line 44: typo

test/run-souffle-tests.py
```diff
@@ -98,7 +98,9 @@
     "sort",             # 202
     "tak",              # 197
     "tic-tac-toe",      # 197
-    "weighted_distances" # 197
+    "weighted_distances", # 197
+    "multiple_inequalities",
```
mbudiu-vmw, line 102: you will have to delete these if you merge the other pr

mbudiu-vmw (approved):
I have to confess that I was a bit overwhelmed.

doc/language_reference/language_reference.md
ryzhyk: fixed this

lib/vec.dl
ryzhyk: `arg_min` is more standard math terminology, but `min_by` is going to be easier to locate for someone who is not familiar with the terminology.  Ok, I guess I'll change to `arg_min`

src/Language/DifferentialDatalog/Validate.hs
ryzhyk: yes, and also to `exprValidate1 EFunc{}`

test/run-souffle-tests.py
ryzhyk: Ok, will make sure to remember this when rebasing the other PR.

PR #775, opened 2020-10-05
Support for souffle's 'range' operator

### LINKED ISSUE #773: Two new souffle test failures.
I did not look closely into this, but it appears that Souffle has a new operator we don't support, used in the following tests: multiple_inequalities, rangeop.

mbudiu-vmw:
Fixes #773 
I have also deprecated the std library 'range' function, since the definition seems to be non-standard.


ryzhyk:
I will merge the other PR first and will then rebase this one on top and fix the OVN issue before merging.

PR #779, opened 2020-10-19
Updated Rust keywords

Kixiron:
Updated Rust keywords & sorted both the datalog and rust keyword lists. Added all keywords and reserved names for Rust 2018 edition, excluding weak keywords like `union` and `'static` (List of keywords came from [the rust reference](https://doc.rust-lang.org/reference/keywords.html))


ryzhyk:
Looks good (syntax error aside), thanks! Let us merge the `group_by` work first and I will then rebase and merge this.

PR #782, opened 2020-10-21
differential_datalog refactors

Kixiron:
The start of an effort to refactor the rust code used by ddlog, this PR fixes some lints, makes some documentation show up to `rustdoc`, splits some components of the former `program.rs` into separate files. Lots of work still to do (especially with `Program::run` and transitioning from strings-as-errors to an error enum), but I'm going to try and keep each PR small-ish and contained so they're easier to review

This PR is best reviewed commit-by-commit


ryzhyk:
Nice, thanks!  The only part I think needs refinement is the split between `mod.rs` and `arrangement.rs`.  Some of the stuff that ended up in `arrangement.rs` is really part of  `struct Program` and should be in `mod.rs`. I can try to figure our a cleaner split.

ryzhyk:
Also, I kind of like errors as strings.  That's what they invariably end up being converted to.  Is there a particular reason you want an enum instead?

Kixiron:
The big reason is that error enums are more useful for users. They're strongly typed, so users can meaningfully handle them, ex.
```rust
impl Program {
    pub fn transaction_start(&mut self) -> Response<()> {
        if self.transaction_in_progress {
            return Err(DatalogError::TransactionInProgress);
        }

        self.transaction_in_progress = true;
        Ok(())
    }
}

// User code
match prog.transaction_start() {
    Ok(()) => {},
    Err(DatalogError::TransactionInProgress) => eprintln!("The transaction failed!"),
    err => eprintln!("Ddlog error: {:?}", err),
}
```
Not the best example since it's rather trivial and immediately makes things into strings, but I think it illustrates the point that you can more intelligently handle the failure of code. If you wanted to do this right now, you'd need something like this
```rust
match prog.transaction_start() {
    Ok(()) => {},
    Err(err) if err == "transaction already in progress" => eprintln!("The transaction failed!"),
    err => eprintln!("Ddlog error: {:?}", err),
}
```
But this is fragile since the error string can change at any time, silently breaking your program. Taken to logical conclusions, this makes any change to error strings a semver violation that requires both documenting somewhere and making a new release (this isn't done/needed now, but it's a good consideration for the future). Additionally, stringly typed errors makes documenting errors significantly harder, take `Program::clear_relation` for instance
```rust
impl Program {
    pub fn clear_relation(&mut self, relid: RelId) -> Response<()> {
        if !self.transaction_in_progress {
            return Err("clear_relation: no transaction in progress".to_string());
        }

        let updates = {
            let rel = self
                .relations
                .get_mut(&relid)
                .ok_or_else(|| format!("clear_relation: unknown input relation {}", relid))?;

            match rel {
                RelationInstance::Stream { .. } => {
                    return Err("clear_relation: operation not supported for streams".to_string())
                }
                RelationInstance::Multiset { elements, .. } => {
                    let mut updates: Vec<Update<DDValue>> = Vec::with_capacity(elements.len());
                    Self::delta_undo_updates(relid, elements, &mut updates);

                    updates
                }
                RelationInstance::Flat { elements, .. } => {
                    let mut updates: Vec<Update<DDValue>> = Vec::with_capacity(elements.len());
                    for v in elements.iter() {
                        updates.push(Update::DeleteValue {
                            relid,
                            v: v.clone(),
                        });
                    }

                    updates
                }
                RelationInstance::Indexed { elements, .. } => {
                    let mut updates: Vec<Update<DDValue>> = Vec::with_capacity(elements.len());
                    for k in elements.keys() {
                        updates.push(Update::DeleteKey {
                            relid,
                            k: k.clone(),
                        });
                    }

                    updates
                }
            }
        };

        self.apply_updates(updates.into_iter())
    }
}
```
This can return a few different errors:
- `"clear_relation: no transaction in progress"`
- `"clear_relation: unknown input relation {random_relid}"`
- `"clear_relation: operation not supported for streams"`
- `"apply_updates: no transaction in progress"` (This one technically shouldn't be able to be triggered, but it still exists in the execution path)

So to intelligently handle errors from this function, we'd need to do something like this:
```rust
match prog.clear_relation(relid) {
    Ok(()) => success_stuff(),
    Err(err) if err == "clear_relation: no transaction in progress" || "apply_updates: no transaction in progress" => {
        retry_starting_transaction()
    }
    Err(err) if err.starts_with("clear_relation: unknown input relation ") => {
        handle_unknown_relation()
    }
    Err(err) if err == "clear_relation: operation not supported for streams" => {
        retry_without_stream()
    }
}
```
Again, this method of error handling is fragile, requires the user to read the source code, etc. Contrasting with the enum'd version, I personally think the enums are much clearer, not to mention more stable and rusty
```rust
match prog.clear_relation(relid) {
    Ok(()) => success_stuff(),
    Err(DatalogError::NoTransaction) => retry_starting_transaction(),
    Err(DatalogError::UnknownRelation { relation_id }) => {
        warn!("attempted to clear unknown relation: {}", relation_id);
        handle_unknown_relation()
    },
    Err(DatalogError::ClearStream) => retry_without_stream(),
}
```
(Note that both of the `no transaction in progress` strings were covered by the `DatalogError::NoTransaction` variant, and that adding error logging to the unknown relation branch was made trivial, where with the stringly-typed version it would have taken string trimming)

With all that in mind, it also allows easily [documenting possible errors](https://rust-lang.github.io/api-guidelines/documentation.html#c-failure), giving us
```rust
impl Program {
    /// Deletes all values in an input table
    ///
    /// # Errors
    ///
    /// Returns [`DatalogError::NoTransaction`] if not currently inside a transaction,
    /// [`DatalogError::UnknownRelation`] if the given `relid` doesn't exist and
    /// [`DatalogError::ClearStream`] if the given `relid` is a stream
    pub fn clear_relation(&mut self, relid: RelId) -> Response<()> { ... }
}
```
(Note that the linked enum variants would be linked to by `rustdoc` because of [intra-doc links](https://blog.rust-lang.org/inside-rust/2020/09/17/stabilizing-intra-doc-links.html) once they stabilize in Rust v1.48, making documentation very easy for users to navigate)

mbudiu-vmw:
Let me add two more reasons:
- localization
- the ability to control behavior (e.g., -Werror for gcc)

ryzhyk:
Yes, yes, I get theory.  I guess I've just never found myself do any of those intelligent things with errors.  They just end up being formatted as strings and output to the user.  But I am ok with using enums.

Kixiron:
@ryzhyk which bits that I put in `arrangement.rs` should stay with the `Program` struct?

ryzhyk:
Let me do it the other way and list bits that, I think, _belong_ to `arrangement.rs`.  The original monolithic `program.rs` contained two kinds of declarations: declarations related to the static representation of a DDlog program (`struct Program`) and declarations related to its runtime instantiation (`struct RunningProgram`).  For instance, `enum Arrangement` is part of the former, whereas `enum ArrangedCollection` belongs to the latter.

So, the `arrangement.rs` module can encapsulate everything related to runtime representation of arrangements, namely:
- `ArrangedCollection`
- `enum A`
- `struct Arrangements`
- `semijoin_arranged`
- `antijoin_arranged`

Everything else should go back to `program.rs`, imo.

Does this make sense?

Kixiron:
I'll admit I'm not entirely happy with (just) that as it leaves `mod.rs` at a monolithic 3000 lines. I'd still like to pull out `Arrangement`, `XFormArrangement`, `XFormCollection` and all of the type aliases out into separate files somehow since they're separate components depended on by `Program`

ryzhyk:
I agree that the remaining `mod.rs` is still too large.  We can always decompose it further. For instance, all the types you mention (`Arrangement`, `XFormArrangement`, `XFormCollection`, etc.) could live in a separate module called `program.rs`, as they are all related to the static representation of the program.

ryzhyk:
How about we merge this PR and if you feel like decomposing things further, it'll be a separate PR?

gitlab is not picking up force-pushed changes for some reason so let me trigger tests in my fork of the repo, and merge once tests pass.

Kixiron:
Sure, sounds good

ryzhyk:
There's one failed test so far:
https://gitlab.com/ddlog/differential-datalog-ryzhyk/-/pipelines/207410178

ryzhyk:
@Kixiron , are you still working on this PR or do you want me to fix up and merge it (there's a compilation failure right now)?

Kixiron:
That'd be amazing, my local setup is misbehaving and I'm baffled as to why the error is happening

ryzhyk:
No worries, I'll do it. Like I said on gitter, I'll add some automation to simplify local testing (it's great to have external contributors to motivate this type of work ;-) ).  What error are you getting?

Kixiron:
I believe it was because `flatc` wasn't on my `PATH`, think I fixed it now

ryzhyk:
One remaining problem is that the `is_update_kind` macro does not work with Rust 1.41. I personally don't like macros and try to avoid them when possible, this being a case in point.  But I also don't see the methods declared with this macro being used anywhere, so looks like it can be safely removed.  @Kixiron, any objections?

Kixiron:
They're functions for the externally-facing API, the innards can just be copy+pasted out if they don't work with the version the project uses (I use 1.47.0 and ddlog compiles perfectly, so if you bumped it shouldn't be a problem)

ryzhyk:
I also use 1.47 locally, but we deliberately use the oldest supported compiler in CI to catch this type of issue.  I am not in principal against requiring a more recent compiler. We've bumped the oldest supported version before and will no doubt do it again in the future, but in this particular case I prefer to just write those functions explicitly.

PR #787, opened 2020-10-27
Eliminate the `Value` wrapper around DDlog values.

ryzhyk:
This commit eliminates the `Value` wrapper around DDlog types stored in
relations.  The wrapper was necessary in order to implement the
`DDValConvert` trait for such types in order to convert them between
normal and type-erased representation.  This design was ugly and
expensive.  Most importantly, the wrapper was visible via the Rust API
and required the user to type auto-generated struct names like
`__Tuple2__internment_Intern____Stringval_internment_Intern____Stringval`.
In addition, auto-derived trait implementations for the wrapper types
(especially serde traits) increased compilation time.  Finally, all of
this junk lived in a separate `value` crate that had to be imported by
clients.

An alternative to the wrapper is to implement `trait DDlogConvert` for
the actual DDlog types rather than wrapping them, which is tricky due to
Rust's orphan rules: a trait must be implemented either in the crate
that declares the trait (in this case, `differential_datalog`) or in the
crate that declares the implementing type.  This does not work for
tuples: since `DDlogConvert` can not be implemented for a generic type,
we cannot provide generic implementations for all tuples in
`differential_datalog`.  On the other hand, we also cannot implement it
in the `types` crate, since tuples are a builtin Rust type.

The workaround is to replace tuples with custom tuple structs
`tuple2<T1,T2>`, `tuple3<T1,T2,T3>`.  We already declare those in the
`types` crate, but previously only used them for tuples with >12 fields,
for which Rust does not implement the various crates we care about.  So
the only change required was to map all DDlog tuple types to `tupleN`
instead of Rust tuples.  We also has to change a bunch of Rust libraries
to work with this convention.

With this change, we can eliminate the `Value` wrapper and place all
`DDlogConvert` implementations either in the `differential_datalog`
crate (for Rust types like `uNN`, `iNN`, `String`, `bool`, ...) or
in the `types` crate.

We more remaining contents of the `value` crate to the top-level
generated crate and get rid of `value` altogether.


src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -585,17 +573,15 @@ compile d_unoptimized specname modules rs_code toml_code dir crate_types = do
...
-compileLib :: (?cfg::Config) => DatalogProgram -> String -> [DatalogModule] -> M.Map ModuleName Doc -> (M.Map ModuleName Doc, Doc, Doc)
-compileLib d specname modules rs_code = (typeLibAllFuncs, valueLib, mainLib)
+compileLib :: (?cfg::Config) => DatalogProgram -> String -> [DatalogModule] -> M.Map ModuleName Doc -> (M.Map ModuleName Doc, Doc)
+compileLib d specname modules rs_code = (typeLib, mainLib)
     where
     modules' = addMissingModules modules
     statics = collectStatics d
-    -- Start with empty modules, except the main module that contains statics
-    -- declarations.
+    -- Start with empty modules, except the main module that contains statics- declarations.
```
mbudiu-vmw, line 581: forgot to delete a dash

mbudiu-vmw (approved):
Most of this looks like an almost mechanical translation. It is surprising how simple this change is, I think it's a sign of a good design.

PR #789, opened 2020-10-28
Allowed trailing commas in relation definitions

Kixiron:
Allows trailing commas in relations, so now this code is valid

```prolog
input relation A(
    field: Value,
    field1: Value1,
    field2: Value2,
    field3: Value3,
)
```

### DIFF
--- a/src/Language/DifferentialDatalog/Parse.hs
+++ b/src/Language/DifferentialDatalog/Parse.hs
@@ -285,11 +285,11 @@ imprt = Import nopos <$ reserved "import" <*> modname <*> (option (ModuleName []
 modname = ModuleName <$> modIdent `sepBy1` reservedOp "::"
 
 typeDef = (TypeDef nopos []) <$ reserved "typedef" <*> typeIdent <*>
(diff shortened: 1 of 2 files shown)


ryzhyk:
This looks good, although it will feel a bit weird that we allow

```
input relation A(
    field: Value,
    field1: Value1,
    field2: Value2,
    field3: Value3,
)
```

but not

```
typedef A = A {
    field: Value,
    field1: Value1,
    field2: Value2,
    field3: Value3,
}
```

Anyway, I'm fine with merging this, and perhaps we can fix the rest in future PRs.

Kixiron:
I got those as well as function arguments (I think)

ryzhyk:
Nice!  Do you feel like modifying some of the tests, e.g., some of the declarations in `test/datalog_tests/simple.dl` to use your new syntax?

Kixiron:
Sure, I'll try it

Kixiron:
Sorry that took so long, I got really busy. It should be good now, fingers crossed!

ryzhyk:
Awesome, thanks!

PR #792, opened 2020-10-29
Use type hints in Python; support for Souffle sum types

mbudiu-vmw:


mbudiu-vmw:
I am using typing rules for python 3.6, which is what I have on my machine.
I am not 100% sure whether they are compatible with later versions of Python. I really hope they are.

ryzhyk:
Looks good, but the `adt_enum` test that this PR enables is failing: https://gitlab.com/ddlog/differential-datalog-mbudiu/-/pipelines/209024573

mbudiu-vmw:
Used to work earlier, I will fix it tomorrow

mbudiu-vmw:
Pushed a fix.

PR #793, opened 2020-10-30
test.sh: A script to selectively run DDlog tests.

ryzhyk:
The DDlog test suite consists of many individual scripts all over the
repo.  Until now the only place that "knew" how to run all these scripts
correctly was the `.gitlab-ci.yml` file.  This was kind of acceptable
because the test suite is too slow to run in its entirety on a single
machine.  Still, running individual test scripts and having to check
`.gitlab-ci.yml` for correct arguments is cumbersome and creates an
extra barrier for external contributions.

So here is finally a script to run all DDlog tests from one place.  In
order to make it easier to pick and choose tests to run, we break them
into groups.  The script takes a list of test names, group names, or a
mix of both as arguments and runs all requested tests.  It can also be
invoked as `./test.sh all` to run all available tests.

Other features:
- Measure the runtime of each test
- Redirect stdout and stderr to file

The `.gitlab-ci.yml` file now uses `test.sh` to run all tests.

I also refactored some test scripts to either clone or pull external git
repos (ovs, ovn, test data repos), which was previously done in
`.gitlab-ci.yml`.

@Kixiron

### DIFF
--- a/rust/template/distributed_datalog/src/instantiate.rs
+++ b/rust/template/distributed_datalog/src/instantiate.rs
@@ -39,8 +39,8 @@ use crate::tcp_channel::TcpSender;
 use crate::txnmux::TxnMux;
 use crate::DDlogServer;
 
-use serde::Serialize;
 use serde::de::DeserializeOwned;
+use serde::Serialize;
 
 /// A mapping from member address to relation IDs used for describing
 /// output relationships.
@@ -175,7 +175,7 @@ where
 enum SinkRealization<C, S>
 where
     C: DDlogConvert,
-    S: Serialize + Debug + Send + 'static
+    S: Serialize + Debug + Send + 'static,
 {
     File(SharedObserver<FileSink<C>>),
     Node(SharedObserver<TcpSender<S>>),
--- a/test.sh
+++ b/test.sh
@@ -1,0 +1,453 @@
+#!/bin/bash
+
+set -e
+
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+NC='\033[0m' # No Color
+THIS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+LOG_DIR="${THIS_DIR}/testsuite.log"
+
+fail() {
+    printf "${RED}FAIL${NC}"
+}
+
+ok() {
+    printf "${GREEN}OK${NC}"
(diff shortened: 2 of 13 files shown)


Kixiron:
This looks great, thank you so much!

ryzhyk:
> This looks great, thank you so much!

For your other PR, you can try running `./test.sh basic` to run a subset of basic tests or just `./test.sh simple` to run the `simple.dl` program to quickly (well, relatively) check you haven't broken stuff before pushing.  It takes a while to get these things right, so feedback is appreciated.

test.sh
```diff
@@ -0,0 +1,455 @@
...
+        do
+            IFS=":" read -ra tokens <<< "$tst_with_descr"
+            tst=${tokens[0]}
+            description=${tokens[1]}
+            printf "        %-25s %s\n" "${tst}" "${description}"
+        done
+    done
+else
+    if [ "x$1" == "xall" ]; then
+        test_list=()
```
mbudiu-vmw, line 368: maybe you should call this group_list

test.sh
```diff
@@ -0,0 +1,455 @@
...
+            fail
+            if [ -z "${NO_STD_REDIRECT}" ]; then
+                printf " [output saved in '${log_file_name}']"
+            fi
+            failed=$((failed+1))
+        fi
+        end=`date +%s`
+        runtime=$((end-start))
+        printf " (${runtime}s)\n"
+    done
```
mbudiu-vmw, line 441: For scripts of this complexity I usually use Python, but hopefully this won't change much from now on.

mbudiu-vmw (approved):

rust/template/distributed_datalog/src/instantiate.rs
```diff
@@ -39,8 +39,8 @@ use crate::tcp_channel::TcpSender;
 use crate::txnmux::TxnMux;
 use crate::DDlogServer;
 
-use serde::Serialize;
 use serde::de::DeserializeOwned;
+use serde::Serialize;
```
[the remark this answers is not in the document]
ryzhyk: Yes, at one point we changed the Rust project setup to not use workspaces, which caused linter to only run on the top-level project. This should be fixed by this PR.

test.sh
ryzhyk: It's really a mix of individual test names and group names.

PR #796, opened 2020-10-30
Expose dump_index API in Java

mbudiu-vmw:


mbudiu-vmw:
@lalithsuresh : this also documents the dump_table Java API a bit better.

java/test7/A.java
```diff
@@ -30,6 +30,7 @@ public void run() throws DDlogException {
         DDlogRecord rec = c.value();
         System.out.println(rec.toString());
         result.dispose();
+        this.api.dumpIndex("AII", i -> System.out.println(i));
```
ryzhyk, line 33: This test doesn't seem to have reference output.

Also, this test is not currently part of the test suite.  Can you add it to `test.sh`, by analogy with `java1`,...,`java4` and do `.gitlab-ci.yml`?

PR #797, opened 2020-11-03
Add DDlogJooqProvider and a corresponding test

lalithsuresh:
The high-level idea here is to intercept a JOOQ connection and redirect calls to DDlog. You can see the test I added for usage.

Given that DDlog only supports standing queries and JDBC/JOOQ has no means of returning incremental changes, the only queries we can support via this API are of the form "select * from X". We should be able to support inserts/updates as well, which I'm exploring (parsing the SQL string like I'm doing now is one way, but in the worst case I can fall back to triggers to learn of changes made to base tables).

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
+++ b/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
@@ -1,0 +1,110 @@
+/*
+ * Copyright 2018-2020 VMware, Inc. All Rights Reserved.
+ * SPDX-License-Identifier: BSD-2
+ */
+package com.vmware.ddlog;
+
+import ddlogapi.DDlogAPI;
+import ddlogapi.DDlogException;
+import ddlogapi.DDlogRecord;
+import org.jooq.DSLContext;
+import org.jooq.Field;
+import org.jooq.Record;
+import org.jooq.Result;
+import org.jooq.SQLDialect;
+import org.jooq.Table;
+import org.jooq.impl.DSL;
+import org.jooq.tools.jdbc.MockDataProvider;
+import org.jooq.tools.jdbc.MockExecuteContext;
+import org.jooq.tools.jdbc.MockResult;
+
+import javax.annotation.Nullable;
+import java.sql.SQLException;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class DDlogJooqProvider implements MockDataProvider {
+    private static final String INTEGER_TYPE = "java.lang.Integer";
+    private static final String STRING_TYPE = "java.lang.String";
+    private static final String BOOLEAN_TYPE = "java.lang.Boolean";
+    private static final String LONG_TYPE = "java.lang.Long";
+
+    private final DDlogAPI dDlogAPI;
+    private final Map<String, List<Field<?>>> tables = new HashMap<>();
+
+    public DDlogJooqProvider(final DDlogAPI dDlogAPI, final List<String> sqlStatements) {
+        this.dDlogAPI = dDlogAPI;
+        final DSLContext using = DSL.using("jdbc:h2:mem:");
+        for (final String sql : sqlStatements) {
+            using.execute(sql);
+        }
+        for (final Table<?> table: using.meta().getTables()) {
+            if (table.getSchema().getName().equals("PUBLIC")) {
+                tables.put(table.getName(), Arrays.asList(table.fields()));
+            }
+        }
+    }
+
+    @Override
+    public MockResult[] execute(final MockExecuteContext ctx) throws SQLException {
+        DSLContext create = DSL.using(SQLDialect.H2);
+        MockResult[] mock = new MockResult[1];
+        // The execute context contains SQL string(s), bind values, and other meta-data
+        String sql = ctx.sql();
+
+        // Exceptions are propagated through the JDBC and jOOQ APIs
+        if (sql.toUpperCase().startsWith("DROP")) {
+            throw new SQLException("Statement not supported: " + sql);
+        }
+        else if (sql.toUpperCase().startsWith("SELECT")) {
+            final String[] s = ctx.sql().toUpperCase().split(" ");
+            if (!s[s.length - 2].equals("FROM")) {
+                throw new SQLException("Statement not supported: " + sql);
+            }
+            final String tableName = s[s.length - 1];
+
+            final List<Field<?>> fields = tables.get(tableName);
+            Result<Record> result = create.newResult(fields);
+            try {
+                dDlogAPI.dumpTable("R" + tableName.toLowerCase(), (record, l) -> {
+                    final Record jooqRecord = create.newRecord(fields);
+                    final Object[] returnValue = new Object[fields.size()];
+                    for (int i = 0; i < fields.size(); i++) {
+                        returnValue[i] = structToValue(fields.get(i), record.getStructField(i));
+                    }
+                    jooqRecord.fromArray(returnValue);
+                    result.add(jooqRecord);
+                });
+            } catch (final DDlogException e) {
+                e.printStackTrace();
+            }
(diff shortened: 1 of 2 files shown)


ryzhyk:
@mbudiu-vmw, will this get tested by `sql_test` in `test.sh`?  If not, how hard is it to add tests for this functionality?

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    @Override
+    public MockResult[] execute(final MockExecuteContext ctx) throws SQLException {
+        final MockResult[] mock = new MockResult[1];
+        // The execute context contains SQL string(s), bind values, and other meta-data
+        final String sql = ctx.sql();
+
+        if (sql.toUpperCase().startsWith("SELECT")) {
```
mbudiu-vmw, line 56: Your SQL parser is a bit rudimentary.
Maybe you want to call this function executeSelectStar

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -0,0 +1,107 @@
...
+            final String[] s = ctx.sql().toUpperCase().split(" ");
+            if (!s[s.length - 2].equals("FROM")) {
```
mbudiu-vmw, line 58: You are not checking that the select expression is *.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -0,0 +1,107 @@
...
+                throw new SQLException("Statement not supported: " + sql);
+            }
+            final String tableName = s[s.length - 1];
+            final List<Field<?>> fields = tables.get(tableName);
```
mbudiu-vmw, line 62: This could be null.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -0,0 +1,107 @@
...
+            final Result<Record> result = dslContext.newResult(fields);
+            try {
+                dDlogAPI.dumpTable("R" + tableName.toLowerCase(), (record, l) -> {
+                    final Record jooqRecord = dslContext.newRecord(fields);
+                    final Object[] returnValue = new Object[fields.size()];
+                    for (int i = 0; i < fields.size(); i++) {
```
mbudiu-vmw, line 68: Assert that the size is the same?

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -0,0 +1,107 @@
...
+            // Exceptions are propagated through the JDBC and jOOQ APIs
+            throw new SQLException("Statement not supported: " + sql);
+        }
+        return mock;
+    }
+
+    @Nullable
+    private Object structToValue(final Field<?> field, final DDlogRecord record) {
+        final Class<?> cls = field.getType();
+        if (record.isStruct() && record.getStructName().equals("std.None")) {
```
mbudiu-vmw, line 88: this does not look like the following one at all. Is it ddlog_std?

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final String INTEGER_TYPE = "java.lang.Integer";
+    private static final String STRING_TYPE = "java.lang.String";
+    private static final String BOOLEAN_TYPE = "java.lang.Boolean";
+    private static final String LONG_TYPE = "java.lang.Long";
+
+    private final DDlogAPI dDlogAPI;
+    private final DSLContext dslContext;
+    private final Map<String, List<Field<?>>> tables = new HashMap<>();
+
+    public DDlogJooqProvider(final DDlogAPI dDlogAPI, final List<String> sqlStatements) {
```
mbudiu-vmw, line 37: I don't know this Jooq stuff, but I would expect there's a way to mock just input tables, not statements, and then statements can run on these tables.

sql/src/test/java/ddlog/JooqProviderTest.java
```diff
@@ -0,0 +1,86 @@
...
+        DDlogAPI.loadDDlog();
+
+        final DDlogAPI dDlogAPI = new DDlogAPI(1, null, true);
+        final int numInserts = 5;
+        for (int i = 0; i < numInserts; i++) {
+            final DDlogRecord rec = new DDlogRecord(i);
+            final DDlogRecord cap = new DDlogRecord(20);
+            final DDlogRecord struct = DDlogRecord.makeStruct("Thosts", rec, cap);
+            final int id = dDlogAPI.getTableId("Rhosts");
+            final DDlogRecCommand command = new DDlogRecCommand(DDlogCommand.Kind.Insert, id, struct);
```
mbudiu-vmw, line 56: Don't you want to insert everything in one transaction? Should be significantly more efficient.

sql/src/test/java/ddlog/JooqProviderTest.java
```diff
@@ -0,0 +1,86 @@
...
+        // Pass the mock connection to a jOOQ DSLContext:
+        DSLContext create = DSL.using(connection);
+        final Result<Record> fetch = create.fetch("select * from hostsv");
+        assertEquals(numInserts, fetch.size());
+        assertArrayEquals(new int[]{0, 1, 2, 3, 4},
+                          fetch.stream().mapToInt(r -> r.get(0, Integer.class)).toArray());
+    }
+
+    public File writeProgramToFile(String programBody) throws IOException {
+        File tmp = new File("/tmp/program.dl");
```
mbudiu-vmw, line 80: why don't you make this one a static final?

mbudiu-vmw (changes requested):

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
ryzhyk: It's now called `ddlog_std::None`. `std.None` is from an old version of DDlog.

mbudiu-vmw:
> @mbudiu-vmw, will this get tested by `sql_test` in `test.sh`? If not, how hard is it to add tests for this functionality?

I believe it will, since it's a maven `@Test`

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
lalithsuresh: What's the correct way to query the size of the ddlog record? I see `record.getStructField(int i)`, but it does not say how many fields the struct has.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
lalithsuresh: This affects the API a bit. Here, I the user supply the SQL statements to make a schema out of (and we extract the information we need it from it, like the mapping of table names to JOOQ Fields). The alternative is that the user supplies a JOOQ connection with the schema already initialized.

sql/src/test/java/ddlog/JooqProviderTest.java
lalithsuresh: This was just for tests. I'll fix it.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
lalithsuresh: In a sense, it'd be an internal error (ddlog or our sql-to-ddlog compiler) for record.<size()> to not be equal to fields.size(). And if it is not the case, we'll correctly get an out of bounds exception. Seems excessive to check every record in dumpTable() for their size. Am I missing something?

mbudiu-vmw (approved):

PR #798, opened 2020-11-07
Added the Vec.first() and Vec.last() methods

Kixiron:
Adds utility methods to get the first and last elements of vectors

### DIFF
--- a/lib/vec.dl
+++ b/lib/vec.dl
@@ -148,12 +148,12 @@ function fold(v: Vec<'A>, f: function('B, 'A): 'B, initializer: 'B): 'B {
 }
 
 // Gets the first element of a vector if it exists
-function first(vec: Vec<'T>) -> Option<'T> {
+function first(vec: Vec<'T>): Option<'T> {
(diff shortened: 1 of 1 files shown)


ryzhyk:
Thanks! I normally add tests for every new function, but this is innocuous enough to merge without tests. Just FYI, we are having some issues with the gitlab CI setup, causing this pipeline to fail. I restarted it and will merge even if some tests fail due to docker issues.

lib/vec.dl
```diff
@@ -147,6 +147,16 @@ function fold(v: Vec<'A>, f: function('B, 'A): 'B, initializer: 'B): 'B {
     res
 }
 
+// Gets the first element of a vector if it exists
+function first(vec: Vec<'T>) -> Option<'T> {
```
ryzhyk, line 151: There's actually a syntax error here and in the other new function: `->` should be `:`.

Kixiron:
That's my bad, force of habit from Rust 😄

PR #802, opened 2020-11-07
Put C API behind feature gate

Kixiron:
Puts the C API behind the `c_api` feature (on by default), allowing applications who don't need a C api to remove it, removing the binary bloat and exposed symbols from their final executable


ryzhyk:
Looks good.  This may require changes to the OVN repo, which, I think, is the only client of the C API at the moment. I'll have another look later today and will fix OVN if needed before merging.

PR #803, opened 2020-11-07
Update num to 0.3

Kixiron:
Updates the `num` crate to version `0.3`


ryzhyk:
This PR changes the output of the `simple.dl` test.  I'm still figuring out if it's harmless.

ryzhyk:
`num 0.3` appears to contain two changes that affect the output of the test suite.  First, it reports overflow in `double` to `float` conversion as `inf` instead of `NaN`. Second, it seems to have changed the `Hash` implementation for the bigint type.  Both seem harmless.  I'll push a change to the reference output and merge.

PR #806, opened 2020-11-10
ddlog-jooq: support inserts and batch queries

lalithsuresh:
This patch adds support for simple insert queries and batch queries via a jooq connection.

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
+++ b/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
@@ -12,6 +12,7 @@ import ddlogapi.DDlogRecord;
 import org.jooq.DSLContext;
 import org.jooq.Field;
 import org.jooq.Record;
+import org.jooq.Record1;
 import org.jooq.Result;
 import org.jooq.Table;
 import org.jooq.impl.DSL;
@@ -27,6 +28,8 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 
+import static org.jooq.impl.DSL.field;
+
 public class DDlogJooqProvider implements MockDataProvider {
     private static final String INTEGER_TYPE = "java.lang.Integer";
     private static final String STRING_TYPE = "java.lang.String";
@@ -36,11 +39,13 @@ public class DDlogJooqProvider implements MockDataProvider {
     private static final String DDLOG_NONE_TYPE = "ddlog_std::None";
     private final DDlogAPI dDlogAPI;
     private final DSLContext dslContext;
+    private final Field<Integer> updateCountField;
     private final Map<String, List<Field<?>>> tables = new HashMap<>();
 
     public DDlogJooqProvider(final DDlogAPI dDlogAPI, final List<String> sqlStatements) {
         this.dDlogAPI = dDlogAPI;
-        dslContext = DSL.using("jdbc:h2:mem:");
+        this.dslContext = DSL.using("jdbc:h2:mem:");
+        this.updateCountField = field("UPDATE_COUNT", Integer.class);
         for (final String sql : sqlStatements) {
             dslContext.execute(sql);
         }
@@ -53,22 +58,33 @@ public class DDlogJooqProvider implements MockDataProvider {
 
     @Override
     public MockResult[] execute(final MockExecuteContext ctx) throws SQLException {
-        final MockResult[] mock = new MockResult[1];
-        // The execute context contains SQL string(s), bind values, and other meta-data
-        final String sql = ctx.sql();
-        if (sql.toUpperCase().startsWith("SELECT")) {
-            mock[0] = executeSelectStar(sql);
-        } else if (sql.toUpperCase().startsWith("INSERT INTO")) {
-            mock[0] = executeInsert(sql);
-        } else {
-            // Exceptions are propagated through the JDBC and jOOQ APIs
-            throw new SQLException("Statement not supported: " + sql);
+        final String[] batchSql = ctx.batchSQL();
+        final MockResult[] mock = new MockResult[batchSql.length];
+        try {
+            dDlogAPI.transactionStart();
+            for (int i = 0; i < batchSql.length; i++) {
+                mock[i] = execute(batchSql[i]);
+            }
+            dDlogAPI.transactionCommit();
+        } catch (final DDlogException e) {
+            throw new RuntimeException(e);
         }
         return mock;
     }
 
+    private MockResult execute(final String sql) throws SQLException {
+        if (sql.toUpperCase().startsWith("SELECT")) {
+            return executeSelectStar(sql);
+        } else if (sql.toUpperCase().startsWith("INSERT INTO")) {
+            return executeInsert(sql);
+        } else {
+            // Exceptions are propagated through the JDBC and jOOQ APIs
+            throw new SQLException("Statement not supported: " + sql);
+        }
+    }
+
     private MockResult executeSelectStar(final String sql) throws SQLException {
-        final String[] s = sql.split("[ ]+");
+        final String[] s = sql.split("[ \t\n]+");
         if (!(s.length == 4 && s[1].equalsIgnoreCase("*") && s[2].equalsIgnoreCase("FROM"))) {
             throw new SQLException("Statement not supported: " + sql);
         }
@@ -96,7 +112,7 @@ public class DDlogJooqProvider implements MockDataProvider {
 
     private MockResult executeInsert(final String sql) throws SQLException {
         final String[] s = sql.replaceAll("[(),]", "")
-                              .split("[ ]+");
+                              .split("[ \t\n]+");
         if (!(s.length >= 4 && s[1].equalsIgnoreCase("INTO")
                 && s[3].equalsIgnoreCase("VALUES"))) {
             throw new SQLException("Statement not supported: " + sql);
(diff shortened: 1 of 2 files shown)


sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -49,26 +58,38 @@ public DDlogJooqProvider(final DDlogAPI dDlogAPI, final List<String> sqlStatemen
...
             // Exceptions are propagated through the JDBC and jOOQ APIs
             throw new SQLException("Statement not supported: " + sql);
         }
-        return mock;
     }
 
     private MockResult executeSelectStar(final String sql) throws SQLException {
-        final String[] s = sql.toUpperCase().split(" ");
-        if (!(s.length == 4 && s[1].equals("*") && s[2].equals("FROM"))) {
+        final String[] s = sql.split("[ \t\n]+");
```
mbudiu-vmw, line 87: ideally you should use the presto parser for this too.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -89,13 +110,42 @@ private MockResult executeSelectStar(final String sql) throws SQLException {
...
+        }
+        final String tableName = s[2];
+        final int valuesIndex = 4; // the index representing the 1st field of the value being inserted
+        final List<Field<?>> fields = tables.get(tableName.toUpperCase());
+        if (fields == null) {
+            throw new SQLException("Table cannot be queried: " + tableName);
+        }
+        String[] valuesTuple = Arrays.copyOfRange(s, valuesIndex, s.length);
+        final DDlogRecord dDlogRecord = toDDlogRecord(tableName, valuesTuple);
+        final int tableId = dDlogAPI.getTableId("R" + tableName.toLowerCase());
```
mbudiu-vmw, line 128: maybe it's time to create a helper function to return a relation name from a table name.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -22,21 +25,27 @@
...
 
+import static org.jooq.impl.DSL.field;
+
 public class DDlogJooqProvider implements MockDataProvider {
     private static final String INTEGER_TYPE = "java.lang.Integer";
     private static final String STRING_TYPE = "java.lang.String";
     private static final String BOOLEAN_TYPE = "java.lang.Boolean";
     private static final String LONG_TYPE = "java.lang.Long";
-
+    private static final String DDLOG_SOME_TYPE = "ddlog_std::Some";
```
mbudiu-vmw, line 38: technically these are two different constructors of the same type, the option type.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -111,4 +161,72 @@ private Object structToValue(final Field<?> field, final DDlogRecord record) {
...
+                            recordsArray[fieldIndex] = maybeOption(field,
+                                    new DDlogRecord(Long.parseLong(args[fieldIndex])));
+                            break;
+                        case STRING_TYPE:
+                            // Strings have to be escaped by single quotes
+                            if (args[fieldIndex].startsWith("'") && args[fieldIndex].endsWith("'")) {
+                                recordsArray[fieldIndex] = maybeOption(field,
+                                        new DDlogRecord(args[fieldIndex].substring(1, args[fieldIndex].length() - 1)));
+                            } else {
+                                throw new RuntimeException("Unexpected string");
```
mbudiu-vmw, line 197: it's handy to output the unexpected string for debugging.
trim() is sometimes useful too.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -111,4 +161,72 @@ private Object structToValue(final Field<?> field, final DDlogRecord record) {
...
+            final Class<?> cls = field.getType();
+            try {
+                // Handle nullable columns here
+                if (args[fieldIndex] == null) {
+                    recordsArray[fieldIndex] = DDlogRecord.makeStruct(DDLOG_NONE_TYPE, new DDlogRecord[0]);
+                }
+                else {
+                    switch (cls.getName()) {
+                        case BOOLEAN_TYPE:
+                            recordsArray[fieldIndex] = maybeOption(field,
```
mbudiu-vmw, line 180: there's a bunch of common functionality between all these cases that could be factored out, not sure it's worth.

sql/src/test/java/ddlog/JooqProviderTest.java
```diff
@@ -27,60 +25,85 @@
...
         MockConnection connection = new MockConnection(provider);
 
         // Pass the mock connection to a jOOQ DSLContext:
         DSLContext create = DSL.using(connection);
-        final Result<Record> fetch = create.fetch("select * from hostsv");
-        assertEquals(numInserts, fetch.size());
-        assertArrayEquals(new int[]{0, 1, 2, 3, 4},
-                          fetch.stream().mapToInt(r -> r.get(0, Integer.class)).toArray());
+
+        // Test 1: insert statements.
```
mbudiu-vmw, line 57: you can still make each test a separate function and call it here.

mbudiu-vmw (changes requested):

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
lalithsuresh: Yeah, I was thinking the same seeing the other issue with primary keys. I'll refactor that in. I hope it isn't too slow.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
lalithsuresh: Ack. Will do.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
lalithsuresh: So can I use just use `ddlog_std::Option` instead?

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
lalithsuresh: Not sure I can though. There's no way to pattern match on the class names, because I don't know the schema in advance. And the way to parse the supplied SQL is different per type.

lalithsuresh: Maybe this code will also get simplified if I use the presto parser here.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
mbudiu-vmw: Your code is correct, it's just that these two are constructors, not types.

sql/src/test/java/ddlog/JooqProviderTest.java
lalithsuresh: For now, these are not independent tests because I can't delete all data from tables yet. I'll refactor that out later.

lalithsuresh:
@mbudiu-vmw I've addressed your comments, except for the last one about refactoring the tests. I'll do that when I can support deletes.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -24,78 +42,226 @@
...
+     */
+    private class QueryVisitor extends AstVisitor<MockResult, String> {
+        @Override
+        protected MockResult visitQuerySpecification(final QuerySpecification node, String sql) {
+            // The checks below encode assumption A1 (see javadoc for the DDlogJooqProvider class)
+            final Select select = node.getSelect();
+            if (!(select.getSelectItems().size() == 1 && select.getSelectItems().get(0) instanceof AllColumns)) {
+                throw new RuntimeException("Statement not supported: " + sql);
+            }
+            assert node.getFrom().isPresent() && node.getFrom().get() instanceof com.facebook.presto.sql.tree.Table;
```
mbudiu-vmw, line 140: this is a bit inconsistent, sometimes you throw, sometimes you assert.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -24,78 +42,226 @@
...
+                final Values values = (Values) node.getQuery().getQueryBody();
+                final String tableName = node.getTarget().toString();
+                final List<Field<?>> fields = tables.get(tableName.toUpperCase());
+                final int tableId = dDlogAPI.getTableId(ddlogRelationName(tableName));
+                for (final Expression row: values.getRows()) {
+                    assert row instanceof Row;
+                    final List<Expression> items = ((Row) row).getItems();
+                    assert items.size() == fields.size();
+                    final DDlogRecord[] recordsArray = new DDlogRecord[items.size()];
+                    for (int i = 0; i < items.size(); i++) {
```
mbudiu-vmw, line 183: I have some helper function Linq.map/zip/any that you may want to occasionally use.

mbudiu-vmw (approved):

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
lalithsuresh: Done!

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
lalithsuresh: I could use Java streams too but this looks simple enough.

PR #815, opened 2020-11-16
Windows compatibility.

ryzhyk:
Windows support with limited amount of testing.

@Kixiron

### DIFF
--- a/tools/install-flatbuf.sh
+++ b/tools/install-flatbuf.sh
@@ -7,20 +7,34 @@ retry() {
   $cmd || (sleep 2 && $cmd) || (sleep 10 && $cmd)
 }
 
-fetch_flatbuf() {
+fetch_flatbuf_unix() {
     rm -rf flatbuffers
     mkdir flatbuffers
     curl -L https://github.com/google/flatbuffers/archive/v1.11.0.tar.gz | tar -zx -C flatbuffers --strip-components 1
 }
 
+fetch_flatbuf_windows() {
+    curl -L https://github.com/google/flatbuffers/releases/download/v1.11.0/flatc_windows_exe.zip > fb.zip && unzip fb.zip
+}
+
+echo "Installing Flatbuf"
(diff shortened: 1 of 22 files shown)


tools/install-flatbuf.sh
```diff
@@ -7,20 +7,34 @@ retry() {
...
-    make
-    cd ..
+    if ( [ "$(uname)" = "Darwin" ] || [ "$(uname)" = "Linux" ] ); then
+        retry fetch_flatbuf_unix
+        cd flatbuffers
+        cmake -G "Unix Makefiles"
+        make
+        cd ..
+    else
+        retry fetch_flatbuf_unix
```
mbudiu-vmw, line 29: is this a bug?

tools/install-flatbuf.sh
```diff
@@ -7,20 +7,34 @@ retry() {
...
 
-fetch_flatbuf() {
+fetch_flatbuf_unix() {
     rm -rf flatbuffers
     mkdir flatbuffers
     curl -L https://github.com/google/flatbuffers/archive/v1.11.0.tar.gz | tar -zx -C flatbuffers --strip-components 1
 }
 
+fetch_flatbuf_windows() {
+    curl -L https://github.com/google/flatbuffers/releases/download/v1.11.0/flatc_windows_exe.zip > fb.zip && unzip fb.zip
```
mbudiu-vmw, line 17: windows comes with curl?

mbudiu-vmw (approved):

tools/install-flatbuf.sh
ryzhyk: Just bad naming.

tools/install-flatbuf.sh
ryzhyk: Travis runs CI script in an environment that does. This is probably not going to work on all flavors of windows.

Kixiron:
It builds and runs, thank you!

PR #817, opened 2020-11-17
jooq-ddlog: migrate away from dumpTables and use incremental API

lalithsuresh:
This patch materializes records in Java and uses ddlog's incremental API instead of dumpTables.

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
+++ b/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
@@ -382,6 +382,7 @@ public class DDlogJooqProvider implements MockDataProvider {
         if (isStruct) {
             final String structName = record.getStructName();
             if (structName.equals(DDLOG_NONE)) {
+                jooqRecord.setValue(field,null);
                 return;
             }
             if (structName.equals(DDLOG_SOME)) {
--- a/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
+++ b/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
@@ -144,7 +144,7 @@ public class DDlogJooqProvider implements MockDataProvider {
     private void onChange(final DDlogCommand<DDlogRecord> command) {
         final int relationId = command.relid();
         final String relationName = dDlogAPI.getTableName(relationId);
-        final String tableName = relationName.substring(1).toUpperCase();
+        final String tableName = relationNameToTableName(relationName);
         final List<Field<?>> fields = tablesToFields.get(tableName);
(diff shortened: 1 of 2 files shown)


mbudiu-vmw:
So, does having a SqlRecord still pay off?

lalithsuresh:
This patch isn't using it (it's based off of master). I'll need to retry it with the `SqlRecords`.

mbudiu-vmw:
I expect it still would save you some time.

mbudiu-vmw:
Is this ready for review?

lalithsuresh:
yes it's ready for review

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -343,6 +358,13 @@ private static String ddlogRelationName(final String tableName) {
         return "R" + tableName.toLowerCase();
     }
 
+    /*
+     * This corresponds to the naming convention followed by the SQL -> DDlog compiler
+     */
+    private static String relationNameToTableName(final String relationName) {
```
mbudiu-vmw, line 364: I think these should become methods in the compiler.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -362,23 +384,37 @@ private static DDlogRecord maybeOption(final Boolean isNullable, final DDlogReco
...
+            if (structName.equals(DDLOG_SOME)) {
+                toValue(field, record.getStructField(0), jooqRecord);
+                return;
+            }
         }
+        toValue(field, record, jooqRecord);
+    }
+
+    private static void toValue(final Field<?> field, final DDlogRecord record, final Record jooqRecord) {
+        final Class<?> cls = field.getType();
```
mbudiu-vmw, line 404: Reflection is bound to be very slow. This will probably dominate the other conversion costs.
You should consider caching the field types map.

mbudiu-vmw (approved):

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
lalithsuresh: This is not using reflection. This returns a member of type Class<?> from JOOQ's Field<> type.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
lalithsuresh: In the sql-ddlog compiler right? I'll send a separate PR for that.

PR #819, opened 2020-11-18
Support SQL binding variables for inserts and deletes

lalithsuresh:
This patch supports binding variables in inserts and deletes.

Most SQL databases support static queries as well as queries where parameters are replaced with '?', where the bindings for these parameters are supplied separately. The latter is typically used with prepared statements. By default, using the JOOQ DSL to construct a query results in prepared statements, so this patch adds support for them in both inserts and deletes.

On the way, this patch also adds support for multi row inserts (`insert into T values (x1, x2, x3), (y1, y2, y3)...`), both with and without bindings.

Lastly, the tests have been refactored to have the ddlog compiler executed just once on the given schema, and re-used through various smaller tests. There are tests for both static and prepared statements.

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
+++ b/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
@@ -137,13 +137,12 @@ public final class DDlogJooqProvider implements MockDataProvider {
                 mock[i] = executeOne(context);
             }
             dDlogAPI.transactionCommitDumpChanges(this::onChange);
-        } catch (final DDlogException e) {
+        } catch (final DDlogException | RuntimeException e) {
             try {
                 dDlogAPI.transactionRollback();
             } catch (DDlogException rollbackFailed) {
                 throw new RuntimeException(rollbackFailed);
             }
-            throw new RuntimeException(e);
         }
         return mock;
     }
--- a/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
+++ b/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
@@ -86,6 +86,7 @@ import static org.jooq.impl.DSL.field;
 public final class DDlogJooqProvider implements MockDataProvider {
     private static final String DDLOG_SOME = "ddlog_std::Some";
     private static final String DDLOG_NONE = "ddlog_std::None";
+    private static final Object[] DEFAULT_BINDING = new Object[0];
     private final DDlogAPI dDlogAPI;
     private final DSLContext dslContext;
     private final Field<Integer> updateCountField;
@@ -132,7 +133,7 @@ public final class DDlogJooqProvider implements MockDataProvider {
             dDlogAPI.transactionStart();
             final Object[][] bindings = ctx.batchBindings();
             for (int i = 0; i < batchSql.length; i++) {
-                final Object[] binding = bindings != null && bindings.length > i ? bindings[i] : new Object[0];
+                final Object[] binding = bindings != null && bindings.length > i ? bindings[i] : DEFAULT_BINDING;
                 final QueryContext context = new QueryContext(batchSql[i], binding);
                 mock[i] = executeOne(context);
             }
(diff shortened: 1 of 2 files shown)


sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -130,11 +131,17 @@ public DDlogJooqProvider(final DDlogAPI dDlogAPI, final List<String> sqlStatemen
...
+            try {
+                dDlogAPI.transactionStart();
+                final Object[][] bindings = ctx.batchBindings();
+                for (int i = 0; i < batchSql.length; i++) {
+                    final Object[] binding = bindings != null && bindings.length > i ? bindings[i] : new Object[0];
+                    final QueryContext context = new QueryContext(batchSql[i], binding);
+                    mock[i] = executeOne(context);
+                }
+            } finally {
+                    dDlogAPI.transactionCommitDumpChanges(this::onChange);
```
mbudiu-vmw, line 141: inconsistent indentation

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -130,11 +131,17 @@ public DDlogJooqProvider(final DDlogAPI dDlogAPI, final List<String> sqlStatemen
...
         final MockResult[] mock = new MockResult[batchSql.length];
         try {
-            dDlogAPI.transactionStart();
-            for (int i = 0; i < batchSql.length; i++) {
-                mock[i] = executeOne(batchSql[i]);
+            try {
+                dDlogAPI.transactionStart();
+                final Object[][] bindings = ctx.batchBindings();
+                for (int i = 0; i < batchSql.length; i++) {
+                    final Object[] binding = bindings != null && bindings.length > i ? bindings[i] : new Object[0];
```
mbudiu-vmw, line 136: You can cache Object[0].

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -164,68 +171,85 @@ private void onChange(final DDlogCommand<DDlogRecord> command) {
...
 
-    private MockResult executeOne(final String sql) throws SQLException {
-        final Statement statement = parser.createStatement(sql, options);
-        final MockResult result = queryVisitor.process(statement, sql);
+    private MockResult executeOne(final QueryContext context) throws SQLException {
+        final Statement statement = parser.createStatement(context.sql(), options);
+        final MockResult result = queryVisitor.process(statement, context);
         if (result == null) {
-            throw new SQLException("Could not execute SQL statement " + sql);
+            throw new SQLException("Could not execute SQL statement " + context);
```
mbudiu-vmw, line 176: May be nice to figure out why this would happen.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -164,68 +171,85 @@ private void onChange(final DDlogCommand<DDlogRecord> command) {
...
+                        final String error = String.format("Incorrect row size for insertion into table %s. " +
+                                                   "Please specify all the table's fields in their declared order: %s",
+                                                    tableName, context.sql());
+                        throw new RuntimeException(error);
                     }
                     final DDlogRecord[] recordsArray = new DDlogRecord[items.size()];
-                    for (int i = 0; i < items.size(); i++) {
-                        recordsArray[i] = parseLiterals.process(items.get(i), fields.get(i).getDataType().nullable());
+                    if (context.hasBinding()) {
+                        // Is a statement with bind variables
```
mbudiu-vmw, line 237: bound

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -164,68 +171,85 @@ private void onChange(final DDlogCommand<DDlogRecord> command) {
...
+                        for (int i = 0; i < items.size(); i++) {
+                            final boolean isNullableField = fields.get(i).getDataType().nullable();
+                            final DDlogRecord record = toValue(fields.get(i), context.binding()[bindingIndex]);
```
mbudiu-vmw, line 240: Should this counter be maintained in the context and provided by a context.nextBinding() method?

mbudiu-vmw (approved):

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
lalithsuresh: That's a good idea. Every use of the binding seems to bump an index. So it's better to keep it behind the Context.

lalithsuresh: Done.

PR #822, opened 2020-11-21
Add checked conversions to DDValue

Kixiron:
- Refactored `ddval.rs`
- Removed the `decl_ddval_convert!()` macro and replaced it with a blanket implementation of `DDValConvert`
- Added the `type_id` function to get a `DDValue`'s `TypeId` (This is used for checking the types of `DDValue`s)
- Made `eq`, `cmp` and `partial_cmp` in `DDValMethods` unsafe functions since they're UB if applied to `DDVal`s of different types
- Added the ability to inspect values during `apply_updates()`
- Added the `type_id()` function to `Relations`
- Use the inspect function + `Relations::type_id()` to check that the proper values are being inserted in ddlog updates

### DIFF
--- a/src/Language/DifferentialDatalog/Compile.hs
+++ b/src/Language/DifferentialDatalog/Compile.hs
@@ -1824,7 +1824,7 @@ mkGroupBy d filters input_val rl@Rule{..} idx = do
                else openTuple d vALUE_VAR group_vars
     let project = "{fn __f(" <> vALUE_VAR <> ": &DDValue) -> " <+> mkType d False (exprType d ctx rhsProject) $$
                   (braces' $ open $$ mkExpr d ctx rhsProject EVal)                                            $$
-                  "::std::rc::Rc::new(__f)}"
+                  "::std::sync::Arc::new(__f)}"
     -- * Create 'struct Group'
     -- * Apply filters following group_by.
     -- * Return variables still in scope after the last filter.
--- a/src/Language/DifferentialDatalog/Compile.hs
+++ b/src/Language/DifferentialDatalog/Compile.hs
@@ -612,7 +612,7 @@ compileLib d specname modules rs_code = (typeLib, mainLib)
     -- Add 'DDValConvert' impls to 'types' crate.
     -- Since multiple DDlog types can map to the same Rust type, we convert
     -- types to Rust and fold over the resulting set.
-    typeLib = S.foldl (\libs t -> M.adjust ($+$ "::differential_datalog::decl_ddval_convert!{" <> pp t <> "}") mainModuleName libs) typeLibAllFuncs
+    typeLib = S.foldl (\libs _ -> M.adjust ($+$ "") mainModuleName libs) typeLibAllFuncs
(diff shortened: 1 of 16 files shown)


src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -610,7 +612,7 @@ compileLib d specname modules rs_code = (typeLib, mainLib)
     -- Add 'DDValConvert' impls to 'types' crate.
     -- Since multiple DDlog types can map to the same Rust type, we convert
     -- types to Rust and fold over the resulting set.
-    typeLib = S.foldl (\libs t -> M.adjust ($+$ "::differential_datalog::decl_ddval_convert!{" <> pp t <> "}") mainModuleName libs) typeLibAllFuncs
+    typeLib = S.foldl (\libs _ -> M.adjust ($+$ "") mainModuleName libs) typeLibAllFuncs
                       $ S.map (render . mkType d True) $ S.filter (typeNeedsDDValConvert d) $ cTypes cstate
```
Kixiron, line 616: I couldn't figure out how to totally remove the `($+$ "")`, so help there would be appreciated

ryzhyk:
This looks amazing, but I notice that Travis is still unhappy. @Kixiron, are you still working on this?

ryzhyk:
The gitlab pipeline is failing too (https://gitlab.com/ddlog/differential-datalog-kixiron/-/pipelines/219518067).  I haven't looked into the failures in detail. Let me know if I can help.

Kixiron:
I'm not really sure how to fix it, it seems some types don't implement `Record: Mutator<T>`

ryzhyk:
Ok, I'll have a look in a couple of days.

ryzhyk:
I fixed compilation issues in this PR (will push some commits shortly). The blanket implementation of `DDValConvert` is a definite improvement.  But the checked conversion stuff is slowing things down at runtime quite significantly.  I will investigate if we can keep some of the checks without killing the performance.

Kixiron:
Hum, that's not something I considered, but it makes a lot of sense. I'd suggest adding `unsafe fn from_ddval{_ref}_unchecked()` methods, the generated code is already type checked (by the ddlog compiler itself) so runtime checking shouldn't be needed at all. At most we could add a `debug_asser_eqt!(this.type_id(), other.type_id())` within those unchecked methods, but in general it should be fine and the checked methods are more for consumers since they aren't infallible machines

Kixiron:
While I agree that the performance thing is an issue and should be fixed, I don't agree with how it was done. These implementations of `PartialOrd`, `Ord` and `PartialEq` are very unsound, as they can trivially cause UB from within safe code. The MCVE is as simple as this:

```rust
let x = Vec::new().as_ddvalue();
let y = 0u8.as_ddvalue();

x == y
```

Instead, I think adding and using unsafe `from_ddval{_ref}_unchecked()` methods within generated code for conversions (still with a debug assertion to catch any bugs) as well as unsafe equality functions is a much better option, as it makes clear the danger at play. This way we get the both of best worlds, fast no-overhead conversions and explicit notation of where unsafety occurs

Issue #823, opened 2020-11-22
Concurrent transactions


Kixiron:
My application is heavily concurrent, and at data-introduction time there's a number of threads attempting to dump data into ddlog. With the current transaction system I have transactions protected by a mutex and I only lock the mutex for the `.apply_valupdates()` portion, but still with those measures in place larger sets of data being introduced will effectively make the system progress in serial since other threads are waiting to be able to introduce their data. A transaction system that allows multiple concurrent transactions would solve this, allowing for a higher level of parallelism

ryzhyk:
There are several issues intertwined here.  Let me try to break things up.  First, there is the problem of scaling DDlog with the number of threads (see, e.g., #820). Ideally, given a large enough transaction, DDlog should be able to process it faster on multiple cores using multiple DD workers.  Note that this does not require support for concurrent transactions, rather this is a way to achieve parallelism inside a transaction.

Internal parallelism aside, I agree that a more concurrent API would make sense. I think such an API can be either built on top of what we've got today or added using relatively minor modifications to DDlog internals, but one has to be careful in defining the semantics of concurrent transactions, as it is not completely intuitive and if not approached carefully may lead to surprising behavior. The first question is whether you want transactions to be truly concurrent, i.e., changes made by concurrent transactions logically belong to the same timestamp. This design probably won't work well, as (1) no transaction can commit until all concurrent transactions have committed, as there is no way to isolate outputs produced by different transactions, (2) when program inputs are declared as `relations` (i.e., sets) as opposed to `streams` of `multisets`,  transactions can conflict, and I really don't want to worry about conflict resolution, rollbacks due to conflicts and all that.

Another option is to run logically serialized transactions in parallel.  When a transaction starts, it obtains a monotonically increasing serial number, which corresponds to a DD epoch. Any number of transactions can push data in parallel.  The data gets cached either inside DDlog or inside DD (the latter will require changing DDlog internals slightly).  When a transaction calls commit, it blocks until this ***and all transactions with smaller timestamps*** commit.   Is this what had in mind?

Kixiron:
Ideally the first would be implemented since each of my transactions are logically separate in all aspects (that is, data does not cross between them), but either could work. My primary purpose is speed since as of now serial transactions are massively crippling execution speed. Internal scaling would also be a huge boon, I'd be happy to help with getting that implemented if you gave some pointers

ryzhyk:
[...]

PR #825, opened 2020-11-22
Made transactions start using transaction handles

Kixiron:
Made transactions (and operations that require an active transaction) start using handles.
A first step towards #823 with the idea that concurrent transactions can more easily be implemented when handles already exist


ryzhyk:
@Kixiron , shall we close this PR? I do not object to supporting concurrent transactions, but it does not look like this fits into our mid-term strategy.

Kixiron:
That's good with me, it'll probably be easier to do once the ddflow section is a little more modular anyways

PR #827, opened 2020-11-24
Syntax Highlighting for DDlog vscode

desharchana19:


mbudiu-vmw:
Thank you!

tools/vscode/README.md
```diff
@@ -0,0 +1,17 @@
+## [License](https://github.com/vmware/differential-datalog/blob/master/LICENSE)
```
mbudiu-vmw, line 1: We avoid adding binary files to the repository.
So instead of adding the vsix files, you should add the source files that you used to produce the vsix files, and a script that produces the vsix files from the sources. Then the source files can also be reviewed.

mbudiu-vmw (changes requested):

mbudiu-vmw:
You should remove the single quote `'` from the autoClosingPairs and surroundingPairs in language-configuration.json.
We don't have `#region` markers, so remove these from folding.
I also think that single quotes do not belong to the indentationRules or to the wordPattern.

tools/vscode/README.md
desharchana19: Added source code 
https://github.com/vmware/differential-datalog/pull/829

mbudiu-vmw:
I will assume that #829 has all the code necessary and close this PR. If that is incorrect please reopen.

PR #832, opened 2020-11-30
Removed excessive indirections from function pointers

Kixiron:
* Changed all `&'static <Rule/Arrangement Function Pointer>`s to hold a function pointer directly, this removes a lot of indirection from a fairly hot path
* Added a root `Cargo.toml` to the repo with the rust code as workspace members, this allows IDEs to analyze the rust code while having the root dir open
* Fixed some random clippy lints

### DIFF
--- a/rust/template/.cargo/config.toml
+++ b/rust/template/.cargo/config.toml
@@ -1,3 +1,2 @@
 [build]
-
 incremental = true
--- a/rust/template/cmd_parser/parse.rs
+++ b/rust/template/cmd_parser/parse.rs
@@ -680,7 +680,7 @@ fn test_int() {
     );
     assert_eq!(
         record(br"-5 "),
-        Ok((&br""[..], Record::Int(-5_i32.to_bigint().unwrap())))
+        Ok((&br""[..], Record::Int((-5_i32).to_bigint().unwrap())))
     );
     assert_eq!(
         hex_val(br"0xabcd "),
--- a/rust/template/differential_datalog/program/mod.rs
+++ b/rust/template/differential_datalog/program/mod.rs
@@ -275,14 +275,14 @@ pub enum XFormArrangement {
     /// FlatMap arrangement into a collection
     FlatMap {
         description: String,
-        fmfun: &'static FlatMapFunc,
+        fmfun: FlatMapFunc,
         /// Transformation to apply to resulting collection.
         /// `None` terminates the chain of transformations.
         next: Box<Option<XFormCollection>>,
     },
     FilterMap {
         description: String,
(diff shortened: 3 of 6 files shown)


Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[workspace]
+members = [
```
ryzhyk, line 2: This will conflict with the workspace code you added to `Compile.hs` earlier.  We actually used to have this section here some time ago, but ended up removing it.  I believe the reason was that `cargo` complained when adding a dependency on the generated Rust project from external crates, as it does not like seeing workspaces anywhere other than in the main crate.

Also, this breaks CI build.

Kixiron: It shouldn't conflict with the `Compile.hs` code since generated code should be out-of-tree, I may be able to fix the test errors by adding `exclude = ["test/*"]`

ryzhyk: Ok, let's work on the workspace thing then, but I think it can be a separate PR.  I like the other things you did here and would like to merge, but I feel like doing workspaces properly may take some more time.

ryzhyk: > It shouldn't conflict with the `Compile.hs` code since generated code should be out-of-tree except in the case of tests, and I may be able to fix those errors by adding `exclude = ["test/*"]` to the root `Cargo.toml`.

Yes, but the generated code will then contain two workspace sections, won't it?

Kixiron: It won't since this `Cargo.toml` only affects the source directory of `differential-datalog` itself, while the generated code (generally elsewhere) has its own, separate workspace

ryzhyk: My bad, I only now realized this is a separate `.toml`. In that case I have no objections as long as we can fix the CI.

Kixiron: You were indeed correct and it was trickier than I anticipated, so I'll do it separately

Kixiron:
I removed the workspace changes from this pr, it's ready for a re-review

rust/template/distributed_datalog/src/lib.rs
```diff
@@ -5,7 +5,7 @@
     future_incompatible,
     illegal_floating_point_literal_pattern,
     improper_ctypes,
-    intra_doc_link_resolution_failure,
+    broken_intra_doc_links,
```
ryzhyk, line 8: Can you please undo this change too?  It upsets older `rustc`. I have a backwards compatible fix in an upcoming PR.

Other than this the PR looks great.

ryzhyk:
Thanks, @Kixiron !

Issue #523, opened 2020-02-14
Add modify function to c api


mbudiu-vmw:

ryzhyk:
This requires adding C API to create records with named fields, since this is the format expected by the modify command (it cannot rely on positional arguments, as an update may touch only a subset of the fields).

mbudiu-vmw:
I think we can close this.

PR #833, opened 2020-12-01
Expose 'Modify' command in Rust/C/Java API 

mbudiu-vmw:
@lalithsuresh : is this sufficient for your needs? See the Java example for a use of the Modify API.
There are a bunch of other APIs we could consider adding, variants of these calls.

### DIFF
--- a/java/ddlogapi.c
+++ b/java/ddlogapi.c
@@ -792,6 +792,70 @@ JNIEXPORT jlong JNICALL Java_ddlogapi_DDlogAPI_ddlog_1struct(
     return (jlong)result;
 }
 
+JNIEXPORT jlong JNICALL Java_ddlogapi_DDlogAPI_ddlog_1named_1struct(
+    JNIEnv *env, jclass obj, jstring s, jobjectArray names, jlongArray handles) {
+    // All the following objects are deallocated on cleanup if not NULL
+    jlong *handlesRef = NULL;
+    const char* sRef = NULL;
+    jlong result = -1;
+    ddlog_record** fields = NULL;
+    const char** fieldNames = NULL;
+
+    jsize len = (*env)->GetArrayLength(env, handles);
+    handlesRef = (*env)->GetLongArrayElements(env, handles, NULL);
+    if (handlesRef == NULL)
+        goto cleanup;
+    sRef = (*env)->GetStringUTFChars(env, s, NULL);
+    if (sRef == NULL)
+        goto cleanup;
+    jsize nameLen = (*env)->GetArrayLength(env, names);
+    if (nameLen != len) {
+        throwDDlogException(env, "number of names does not match number of values");
+        goto cleanup;
+    }
+    fields = malloc(len * sizeof(ddlog_record*));
+    if (fields == NULL) {
+        throwOutOfMemException(env, "Could not allocate buffer for %d fields.", len);
+        goto cleanup;
+    }
+    for (size_t i = 0; i < len; i++)
+        fields[i] = (ddlog_record*)handlesRef[i];
+
+    fieldNames = calloc(len, sizeof(char*));
+    if (fieldNames == NULL) {
+        throwOutOfMemException(env, "Could not allocate buffer for %d names.", len);
+        goto cleanup;
+    }
+    for (size_t i = 0; i < len; i++) {
+        jobject jname = (*env)->GetObjectArrayElement(env, names, i);
+        if (jname == NULL) {
+            throwDDlogException(env, "Field name cannot be null");
+            goto cleanup;
+        }
+        const char* name = (*env)->GetStringUTFChars(env, jname, NULL);
+        fieldNames[i] = name;
+    }
+
+    ddlog_record* rec = ddlog_named_struct(sRef, fieldNames, fields, len);
+    result = (jlong)rec;
+ cleanup:
+    if (handlesRef)
+        (*env)->ReleaseLongArrayElements(env, handles, handlesRef, 0);
+    if (fields)
+        free(fields);
+    if (fieldNames) {
+        for (size_t i = 0; i < len; i++) {
+            if (fieldNames[i] == NULL) break;
+            jobject jname = (*env)->GetObjectArrayElement(env, names, i);
+            (*env)->ReleaseStringUTFChars(env, jname, fieldNames[i]);
+        }
+        free(fieldNames);
+    }
+    if (sRef)
+        (*env)->ReleaseStringUTFChars(env, s, sRef);
+    return result;
+}
+
 JNIEXPORT jboolean JNICALL Java_ddlogapi_DDlogAPI_ddlog_1is_1bool(
     JNIEnv *env, jclass obj, jlong handle) {
     return (jboolean)ddlog_is_bool((ddlog_record*)handle);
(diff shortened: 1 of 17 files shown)


lalithsuresh:
@mbudiu-vmw I'll give this a spin and let you know.

ryzhyk:
Resolves #523

java/ddlogapi/DDlogRecord.java
```diff
@@ -380,12 +397,20 @@ public String getStructName() {
...
     }
 
-    public DDlogRecord getStructField(int index) {
+    public DDlogRecord getStructField(int index) throws DDlogException {
         if (!this.isStruct())
             throw new RuntimeException("Value is not a struct");
         return fromSharedHandle(DDlogAPI.ddlog_get_struct_field(this.handle, index));
     }
 
+    public DDlogRecord getStructFieldNoEx(int index) {
```
lalithsuresh, line 406: Better to call this getStructFieldUnchecked.

lalithsuresh (commented):
I'll continue later after testing it.

java/ddlogapi/DDlogRecord.java
mbudiu-vmw: It's checked, but not a checked exception.

java/ddlogapi.c
```diff
@@ -792,6 +792,55 @@ JNIEXPORT jlong JNICALL Java_ddlogapi_DDlogAPI_ddlog_1struct(
...
+    const char** fieldNames = malloc(len * sizeof(char*));
+    if (fieldNames == NULL) {
+        free(fields);
+        throwOutOfMemException(env, "Could not allocate buffer for %d names.", len);
+        return -1;
+    }
+    for (size_t i = 0; i < len; i++) {
+        jobject jname = (*env)->GetObjectArrayElement(env, names, i);
+        if (jname == NULL) {
+            free(fields);
```
lalithsuresh, line 824: might be cleaner to just use a goto to jump to the end of the function and call the frees you need.

ryzhyk (commented):
Will read the rest and send more feedback later today.

mbudiu-vmw:
I will wait for more feedback before sending a new version of the PR.

rust/template/ddlog.h
```diff
@@ -1507,4 +1556,24 @@ extern ddlog_cmd* ddlog_delete_val_cmd(table_id table, ddlog_record *rec);
  */
 extern ddlog_cmd* ddlog_delete_key_cmd(table_id table, ddlog_record *rec);
 
+/*
+ * Create a modify-by-key command.
+ *
+ * `table` - input table to delete from.
+ * `key` - key to modify.  The function takes ownership of this record.
+ * `values` - values to modify.  The function takes ownership of this record.
```
ryzhyk, line 1564: Not  a good choice of name, imo.  This argument describes a change to apply to the value.  `delta` coule be a better name, but it's already used with a different meaning in this API. Perhaps `patch`?

rust/template/differential_datalog/record.rs
```diff
@@ -1,6 +1,6 @@
 //! An untyped representation of DDlog values and database update commands.
 
-#![allow(unknown_lints)]
+#![allow(unknown_lints,clippy::needless_range_loop)]
```
ryzhyk, line 3: Actually, Rust is correct, you should probably use an iterator, not a range loop.  I would do that rather than disable the warning.

ryzhyk (approved):

lalithsuresh:
I'm running into some issues using this.

The ddlog program generated by the sql-ddlog compiler is:

```
import fp
import time
import sql
import sqlop

typedef Thosts = Thosts{id:Option<string>, capacity:Option<signed<64>>, up:Option<bool>}

output relation Rgood_hosts[Thosts]
Rgood_hosts[v0] :- Rhosts[v],unwrapBool(a_lt_NR(v.capacity, 64'sd10)),var v0 = v.
```

My update statement is: `update hosts set capacity = 11 where id = 'n1'`

My code to create the Modify command is like this:

```
final DDlogRecord key = matchExpressionFromWhere(where, pkFields, context);
final int numColumnsToUpdate = updateCall.getTargetColumnList().size();

final DDlogRecord[] updatedValues = new DDlogRecord[numColumnsToUpdate];
final String[] columnsToUpdate = new String[numColumnsToUpdate];
for (int i = 0; i < numColumnsToUpdate; i++) {
    final String columnName = ((SqlIdentifier) update.getTargetColumnList().get(i)).getSimple()
                                                                                    .toLowerCase();
    final Field<?> field = getFieldByName(allFields, columnName);
    final boolean isNullableField = field.getDataType().nullable();
    final DDlogRecord valueToUpdateTo = update.getSourceExpressionList().accept(PARSE_LITERALS);
    final DDlogRecord maybeWrapped = maybeOption(isNullableField, valueToUpdateTo);
    updatedValues[i] = maybeWrapped;
    columnsToUpdate[i] = columnName;
}
final DDlogRecord updateRecord = DDlogRecord.makeNamedStruct("", columnsToUpdate, updatedValues);
final int tableId = dDlogAPI.getTableId(ddlogRelationName(tableName));
final DDlogRecCommand command = new DDlogRecCommand(DDlogCommand.Kind.Modify, tableId, key, updateRecord);
dDlogAPI.applyUpdates(new DDlogRecCommand[]{command});
```

I'm noticing some odd behaviors.

1. When creating the DDlog Modify command, if I specify the field name as "CAPACITY" instead of "capacity", nothing crashes, but no row gets updated. So it silently fails.

2. When I correct the field name to "capacity", I get a different error depending on whether I pass "Some{11}" or "11" as the command.

When passing Some{11} (i.e., `maybeWrapped`):

```
Caused by: ddlogapi.DDlogException: ddlog_apply_updates(): error: Cannot use positional struct as mutator
	at ddlogapi.DDlogAPI.ddlog_apply_updates(Native Method)
	at ddlogapi.DDlogAPI.applyUpdates(DDlogAPI.java:497)
	at com.vmware.ddlog.DDlogJooqProvider$QueryVisitor.visitUpdate(DDlogJooqProvider.java:252)
	... 39 more
```

When passing 11 (i.e., `valueToUpdateTo`).

```
Caused by: ddlogapi.DDlogException: ddlog_apply_updates(): error: not a struct Int(BigInt { sign: Plus, data: BigUint { data: [11] } })
	at ddlogapi.DDlogAPI.ddlog_apply_updates(Native Method)
	at ddlogapi.DDlogAPI.applyUpdates(DDlogAPI.java:497)
	at com.vmware.ddlog.DDlogJooqProvider$QueryVisitor.visitUpdate(DDlogJooqProvider.java:252)
	... 39 more
```

The error message when I pass 11 makes sense, as the type does not match. So I suspect something is wrong about the behavior with passing Some{11}.

lalithsuresh:
@mbudiu-vmw you can try my code above in the https://github.com/lalithsuresh/differential-datalog/tree/updates-with-ddlog-modify-command branch. The test to run isJooqProviderTest.testUpdates().

rust/template/ddlog.h
lalithsuresh: Or modification.

mbudiu-vmw:
@lalithsuresh hopefully this fixes the issue you have discovered.
It looks like my attempt to squash commits was a bit too successful - we only have the message from the previous commit, which I didn't make.

ryzhyk (commented):
Looks great. Just need to untangle it from the old PR. Can I help with that?

mbudiu-vmw:
@lalithsuresh : once this passes the CI tests we'll merge and hopefully you'll be good to go.

PR #834, opened 2020-12-01
Added root cargo workspace

Kixiron:


Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[workspace]
```
ryzhyk, line 1: Would it make more sense to keep this file in `rust/` rather than in the project root?

Kixiron: Either will work, it's up to you

ryzhyk: In that case `rust` is a better location.

Kixiron: Ah, I forgot the entire point of the file, whoops. Being in the root of the project allows language servers to detect it and subsequently check the rust files we want it to, being nested within a folder requires opening that directly with the editor

ryzhyk: Fair enough. Merging as is.  Thanks!

PR #835, opened 2020-12-01
Removed some of the allocations in running programs

Kixiron:


ryzhyk:
This is not performance-critical code, but your improvements still make things cleaner.  If panic in Travis though.  Do you know what went wrong or would you like me to investigate?

Kixiron:
I'm incredibly confused as to how/why this is panicking now, none of my changes changed behavior. My best guess is that something's awry in [`dependencies()`](https://github.com/vmware/differential-datalog/blob/da69fc29349860df3607d8f100120e3841a0ed48/rust/template/differential_datalog/program/mod.rs#L1625-L1640), but I don't actually know what could have

ryzhyk:
Ok, I'll try to fix this, but it will have to wait a few days.

rust/template/differential_datalog/program/mod.rs
```diff
@@ -1595,39 +1601,41 @@ impl Program {
...
-            .collect()
+
+                None
+            }
+        })
     }
 
     /// Return all relations required to compute rels, excluding recursive dependencies on rels
-    fn dependencies(rels: &[&Relation]) -> FnvHashSet<Dep> {
+    fn dependencies<'a, R>(mut rels: R) -> FnvHashSet<Dep>
```
ryzhyk, line 1625: This where this PR breaks things.  The original version takes a slice and traverses it multiple times.  The new version takes iterator.  It exhausts the iterator in the first loop, so subsequent call to `rels.all` always returns `true`, so `filter` does nothing.

ryzhyk (changes requested):

rust/template/differential_datalog/program/mod.rs
ryzhyk: @Kixiron , let me know if you'll have time to fix this; otherwise I can do it.

ryzhyk: > This is not performance-critical code, but your improvements still make things cleaner.

I take it back. It turns out that `struct Program` can grow huge for large programs due all the strings that are further bloated by type annotations injected by the type checker. I really like this PR now as it avoids cloning this struct. I am going to fix the bug and merge.

ryzhyk: Actually, it's the other PR that does this, and it still copies the program ones.

Kixiron: I believe that `self: Arc<Self>` is valid in stable rust, that would allow not cloning the physical program any

ryzhyk (approved):

rust/template/differential_datalog/program/mod.rs
ryzhyk: Good point. I'll try that too.  I am also going to change types of all strings in `Program` to `Cow<String>`, as these are typically static strings, so we don't need to allocate them even once.

PR #836, opened 2020-12-02
Refactor Program::run()

Kixiron:
* Split off the timely worker runtime to be separate from `Program` itself
* Started splitting up the monolith that is `Program::run()`, more to come later
* Made workers hold an `Arc<Program>` instead of each cloning their own `Program` instance
* Shoved all the profiling-related stuff (the cpu/timely atomics and the profiling channel) into a struct that each worker holds
* Removed the channels to send worker `Thread`s back to the main thread, now `RunningProgram` just holds onto `WorkerGuards`
* Removed the extra thread that did nothing but spin up a timely computation
* Moved the timely/differential logger initialization into its own function
* Moved the actual `.dataflow()` runtime into `DDlogWorker::session_dataflow()`
* Changed static methods on `Program` that dealt with the timely runtime to be methods on `DDlogWorker`

### DIFF
--- a/rust/template/differential_datalog/program/mod.rs
+++ b/rust/template/differential_datalog/program/mod.rs
@@ -30,9 +30,10 @@ use std::{
     ops::Deref,
     sync::{
         atomic::{AtomicBool, Ordering},
-        mpsc, Arc, Barrier, Mutex,
+        mpsc::{self, Receiver, Sender, TryRecvError},
+        Arc, Barrier, Mutex,
     },
-    thread,
+    thread::{self, JoinHandle, Thread},
     time::Duration,
 };
 use timestamp::{TSAtomic, ToTupleTS};
@@ -50,8 +51,10 @@ use differential_dataflow::trace::wrappers::enter::TraceEnter;
 use differential_dataflow::trace::{BatchReader, Cursor, TraceReader};
 // use differential_dataflow::trace::cursor::CursorDebug;
 use differential_dataflow::Collection;
-use timely::communication::initialize::Configuration;
-use timely::communication::Allocator;
+use timely::communication::{
+    initialize::{Configuration, WorkerGuards},
+    Allocator,
+};
 use timely::dataflow::operators::*;
 use timely::dataflow::scopes::*;
 use timely::dataflow::ProbeHandle;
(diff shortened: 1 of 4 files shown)


rust/template/differential_datalog/program/mod.rs
```diff
@@ -640,25 +637,22 @@ pub struct RunningProgram {
...
     /// deadlocks when one of the workers has died, but `recv` blocks instead
     /// of failing, since the channel is still considered alive.
-    reply_recv: Vec<mpsc::Receiver<Reply>>,
+    reply_recv: Vec<Receiver<Reply>>,
     relations: FnvHashMap<RelId, RelationInstance>,
-    /// Join handle of the thread running timely computation.
-    thread_handle: Option<thread::JoinHandle<Result<(), String>>>,
-    /// Timely worker threads.
-    worker_threads: Vec<thread::Thread>,
+    worker_guards: Option<WorkerGuards<Result<(), String>>>,
```
ryzhyk, line 647: Looks like this causes a couple of problems.  First, HDDlog is no longer `Send`, which breaks the server_api test: https://gitlab.com/ddlog/differential-datalog-kixiron/-/jobs/887145084
(This has been fixed in the latest timely: https://github.com/TimelyDataflow/timely-dataflow/issues/296, but we are still using an older version from crates.io).

Second, removing the timely thread seems to have altered the error handling path causing the nested panic error in another test: https://gitlab.com/ddlog/differential-datalog-kixiron/-/jobs/887145073

I need more time to investigate these issues, i.e., find a workaround for the former and debug the latter.

rust/template/differential_datalog/program/worker.rs
```diff
@@ -0,0 +1,853 @@
...
+    }
+
+    /// Get the index of the current worker
+    pub fn worker_index(&self) -> usize {
+        self.worker.index()
+    }
+
+    /// Set the current transaction frontier's timestamp
+    fn set_frontier_timestamp(&self, timestamp: TS) {
+        self.frontier_timestamp.store(timestamp, Ordering::Relaxed);
```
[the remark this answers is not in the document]
Kixiron: I actually changed these to `Acquire`/`Release`, I'll push the changes later today

rust/template/differential_datalog/program/mod.rs
Kixiron: The former can be fixed with an `unsafe impl Send for HDDlog {}` if you're comfortable with it, the latter is more tricky though

ryzhyk: - I think `unsafe impl` is ok.
- I haven't looked closely into the second issue yet. I suspect it's just a matter of more careful error handling now that we don't have a separate thread to catch panic's.

PR #841, opened 2020-12-04
Capture stdout/stderr of processes invoked

### LINKED ISSUE #838: A way to read messages from Standard Error in Java API
Currently during compilation is something goes wrong it just logs the error in Standard Err and terminates. We should have a mechanism to return these error messages to the caller

mbudiu-vmw:
Fixes #838
The class CompilationOutput is now available. It is passed as an argument to functions that compile generated programs.
After the compilation is finished the function gives access to the exit code and to stdout and stderr of the compilation processes.

### DIFF
--- a/java/ddlogapi/DDlogAPI.java
+++ b/java/ddlogapi/DDlogAPI.java
@@ -659,33 +659,85 @@ public class DDlogAPI {
         return ddlogHome;
     }
 
+    public static class CompilationResult {
+        public int exitCode;
+        StringBuilder stdout;
+        StringBuilder stderr;
+        final boolean verbose;  // if true echo messages
+
+        public CompilationResult(boolean verbose) {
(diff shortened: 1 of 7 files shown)


java/ddlogapi/DDlogAPI.java
```diff
@@ -659,33 +659,85 @@ public static String ddlogInstallationPath() throws DDlogException {
         return ddlogHome;
     }
 
+    public static class CompilationResult {
```
ryzhyk, line 662: Do some of the methods in this class need to be synchronized because they can be invoked by the client as well as by the output reader threads this class spawns?

ryzhyk (approved):
@mbudiu-vmw, please merge if it's ready.

Gitlab CI is failing due to unrelated souffle test issue.

PR #842, opened 2020-12-05
Support for base64 encoding and decoding

### LINKED ISSUE #837: Support to encode and decode base64 strings
Need utility methods to perform base64 encoding and decoding for a given string.

mbudiu-vmw:
Fixes #837

### DIFF
--- a/lib/base64.dl
+++ b/lib/base64.dl
@@ -1,2 +1,5 @@
-extern function decode(s: string): Result<string, string>
-extern function encode(s: string): string
+// Decodes a base64 encoding; if the input is not a valid encoding it
+// produces an error describing the problem.
+extern function decode(s: string): Result<Vec<u8>, string>
+// Encode a byte array as a base64 string.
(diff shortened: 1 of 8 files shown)


lib/base64.dl
```diff
@@ -0,0 +1,2 @@
+extern function decode(s: string): Result<string, string>
```
ryzhyk, line 1: A comment here would be great.

test/datalog_tests/base64_test.dl
```diff
@@ -0,0 +1,5 @@
+import base64
+
+output relation Base64(str: string, encoding: string)
+Base64("TEST", decode(encode("TEST")).unwrap_or("")).
```
ryzhyk, line 4: Would you consider adding tests for the error cases?

ryzhyk (approved):

ryzhyk (approved):

PR #845, opened 2020-12-06
Report dep cycle in stratified negation errors.

ryzhyk:
Enhance stratified negation error messages by printing the dependency
cycle that caused the error:

```
Relation R3 is mutually recursive with R2 and therefore cannot appear negated in this rule.
Dependency cycle: R3 -> R2 -> R3 // This part of the message is new.
```


mbudiu-vmw (approved):
I think the message could be even simple, as you describe it in the comments : a relation cannot depend on itself negated.

PR #850, opened 2020-12-09
Decompose generated Rust project into crates. 

ryzhyk:
Large DDlog projects get compiled to Rust projects that can take a long
time to compile to binary code.  One way to remediate this is to make
sure that small changes to DDlog code produce small changes to Rust code
that can be quickly re-compiled by only re-compiling the affected
modules.  Unfortunately, Rust's incremental compilation only really
works at the granularity of a whole crate (there is also some amount
incremental intra-crate compilation going on, but it is quite
unpredictable and fragile).  So we need to break up the generated
project into crates to achieve incremental compilation.

The crate decomposition algorithm preserves the structure of the DDlog
project, so that each declaration can be found at the same path in the
Rust code as in the DDlog code.  It starts with placing each module in
a separate crate and then merges crates with circular depencies (induced
by `import` statements).  In addition, if two modules belong to the same
crate, then the subtree of the project containing these modules and
their ancestors up to the lowest common ancestor must be in the same
crate.

Each generated module contains the following Rust declarations:
- Functions and types declared in the corresponding DDlog module
- Arrangements and keys for relations declared in the corresponding
  DDlog module.
- Rules, facts, transformer applications declared in the corresponding
  DDlog module.
- The main program crate (that gets recompiled every time) only
  assembles all of the above into a dataflow graph to be passed to the
  `differential-datalog` crate.

The rest of this patch adjusts the generated code to correctly
cross-reference various declarations across generated crates.

One side effect of the new Rust project structure is that Rust type
declarations are scattered across crates, which are generated
automatically and whose boundaries can change.  To expose a stable Rust
interface to the DDlog program, the main crate reexports all function
and type declarations precisely mirroring the structure of the DDlog
program.


lib/ddlog_rt.rs
```diff
@@ -275,65 +343,12 @@ impl<Args: 'static + Clone, Output: 'static + Clone> Abomonation
...
 #[cfg(test)]
 mod tests {
-    use crate::closure::Closure;
-    use crate::closure::ClosureImpl;
-    use ::serde::Deserialize;
-    use ::serde::Serialize;
+    use super::Closure;
+    use super::ClosureImpl;
+    use serde::Deserialize;
+    use serde::Serialize;
```
Kixiron, line 351:
```suggestion
    use super::{Closure, ClosureImpl};
    use serde::{Deserialize, Serialize};
```

lib/internment.rs
```diff
@@ -138,66 +126,7 @@ where
...
-{
-    type Target = T;
-
-    fn to_flatbuf_vector_element(&self, fbb: &mut fbrt::FlatBufferBuilder<'b>) -> Self::Target {
-        self.as_ref().to_flatbuf_vector_element(fbb)
-    }
-}
-
-pub fn istring_join(strings: &crate::ddlog_std::Vec<istring>, sep: &String) -> String {
+pub fn istring_join(strings: &ddlog_std::Vec<istring>, sep: &String) -> String {
```
Kixiron, line 129:
```suggestion
pub fn istring_join(strings: &[istring], sep: &str) -> String {
```
Deref coercion should allow `&String` -> `&str` and `&ddlog_std::Vec<T>` -> `&[T]`, may need to implement `Deref<Target = Vec<T>>` and `Deref<Target = [T]>` for `ddlog_std::Vec<T>` though

lib/ddlog_bigint.rs
```diff
@@ -20,6 +16,14 @@ use std::ops::*;
...
+/* This module is designed to be imported both as a standard DDlog library and as a normal Rust
+ * module, e.g., from `differential_datalog_test`.  We therefore need to import these traits
+ * so that they are available in the latter case and rename them so that they don't cause duplicate
+ * import error in the former case. */
+use differential_datalog::record::FromRecord as FromRec;
+use differential_datalog::record::IntoRecord as IntoRec;
+use differential_datalog::record::Record;
+
 #[derive(Eq, PartialOrd, PartialEq, Ord, Clone, Hash)]
 pub struct Int {
```
Kixiron, line 28: Implementing `Deref<Target = BigInt>` would add to ergonomics a lot (Same with `Uint`)

lib/ddlog_bigint.rs
```diff
@@ -323,3 +327,263 @@ impl num::Zero for Int {
...
+        }
+    }
+    pub fn parse_bytes(buf: &[u8], radix: u32) -> Uint {
+        Uint {
+            x: BigUint::parse_bytes(buf, radix).unwrap(),
+        }
+    }
+}
+
+#[no_mangle]
```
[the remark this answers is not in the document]
ryzhyk: Good idea!

lib/allocate.rs
```diff
@@ -3,11 +3,11 @@ use std::collections::BTreeSet;
 use std::ops;
 
 pub fn allocate<B: Ord + Clone, N: num::Num + ops::Add + cmp::Ord + Copy>(
-    allocated: &crate::ddlog_std::Set<N>,
-    toallocate: &crate::ddlog_std::Vec<B>,
+    allocated: &ddlog_std::Set<N>,
```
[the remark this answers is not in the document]
ryzhyk: Thanks for reviewing, this is really valuable!

All Rust code, including libraries used to live in the same namespace.  Fortunately, this is no longer the case, so yes, at least as far as native library code goes, imports should work. I sometimes prefer qualified names, especially when the actual type or function name is too generic.  I noticed that you feel different about these things :)

I'll fix this library as you suggest.  Overall, I feel that our native library code has been tortured to death by all the refactorings and looks terrible in many ways, including those identified in your review.

lib/ddlog_log.rs
```diff
@@ -32,6 +31,7 @@ static LOG_CONFIG: Lazy<sync::RwLock<LogConfig>> =
 
 /// Logging API exposed to the DDlog program.
 /// (see detailed documentation in `log.dl`)
+#[allow(clippy::ptr_arg, clippy::trivially_copy_pass_by_ref)]
```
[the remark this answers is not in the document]
ryzhyk: This is probably doable at least for read-only function arguments, but will probably require changes to the code generator. Definitely not in this PR.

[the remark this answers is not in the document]
Kixiron: Yes, that's definitely out of scope for this pr

lib/allocate.rs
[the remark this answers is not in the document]
Kixiron: I generally prefer importing things since it becomes nicer to work within, for overtly generic names (such as `Vec`) I'd just alias it to `DDlogVec` or something

lib/ddlog_bigint.rs
ryzhyk: As long as it doesn't mess up the generated code. There's a bit of a tension here between ergonomics and surprising side effects in automatically generated code where fewer magic conversions is better. The same is true for many other library types. I don't think we have a consistent approach here. I'll try implementing these `Deref` and see if anything breaks.

lib/allocate.rs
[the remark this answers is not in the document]
ryzhyk: Heh, we can agree to disagree.  In case of our native libraries, though, any consistent style would be an improvement :)

lib/ddlog_bigint.rs
Kixiron: If it ends up that there's a multitude of unwanted behavior happening, we can delay that and I can make a pr that makes you ye generated code use more qualified function calls, but hopefully it doesn't come to that

lib/ddlog_bigint.rs
```diff
@@ -69,7 +73,7 @@ impl Int {
     pub fn from_i128(v: i128) -> Int {
         Int { x: BigInt::from(v) }
     }
-    pub fn from_Uint(v: uint::Uint) -> Int {
+    pub fn from_Uint(v: Uint) -> Int {
```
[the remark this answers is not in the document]
ryzhyk: Turns out, we rely on `from_<typename>` functions in generated code. Sure, we could use something like `to_snake_case`, but I prefer to keep things simple.  I will add a comment before the `impl` block to clarify the choice of naming.

src/Language/DifferentialDatalog/Crate.hs
```diff
@@ -0,0 +1,235 @@
...
+
+import Language.DifferentialDatalog.Module
+import Language.DifferentialDatalog.Syntax
+import Language.DifferentialDatalog.Util
+
+-- A crate is a set of modules.
+type Crate = S.Set ModuleName
+
+-- Returns the top-level module of a crate.  This is well-defined, as by
+-- construction crate modules form a crate with a unique root.
```
mbudiu-vmw, line 56: typo?

src/Language/DifferentialDatalog/Crate.hs
```diff
@@ -0,0 +1,235 @@
...
+-- crate1' = {m1, m3, m4, m6, m8}.
+addLowestCommonAncestor :: Crate -> Crate
+addLowestCommonAncestor crate =
+    S.foldl (\mods m -> foldl' (\mods' mname -> S.insert (ModuleName mname) mods') (S.insert m mods)
+                               $ drop lca_len $ inits $ modulePath m)
+            S.empty crate
+    where
+    -- Lowest common ancestor path length.
+    lca_len = length $ lcp $ map modulePath $ S.toList crate
+    -- Longest common prefix.
```
mbudiu-vmw, line 146: these could be utility functions in some library, but I don't know if it's worth moving them

src/Language/DifferentialDatalog/Expr.hs
```diff
@@ -527,6 +528,14 @@ exprIsPure d ctx e | isNothing funcs = False
                    | otherwise = all (\f -> not $ funcGetSideEffectAttr d f) $ fromJust funcs
     where funcs = exprFuncsRec d ctx e
 
+-- | Expression can be evaluated statically.
```
mbudiu-vmw, line 531: aren't literals in this category too? or constant expressions?

src/Language/DifferentialDatalog/Module.hs
```diff
@@ -109,9 +115,14 @@ emptyModule mname = DatalogModule {
     moduleDefs = emptyDatalogProgram
 }
 
--- Standard library module name.
+-- Standard library module names.
 stdLibs :: [ModuleName]
-stdLibs = [ModuleName [mOD_STD], ModuleName ["internment"]]
+stdLibs = [ ModuleName [mOD_RT]
```
mbudiu-vmw, line 120: Could this be read from ddlog_std.dl?

src/Language/DifferentialDatalog/Module.hs
```diff
@@ -121,57 +132,59 @@ stdImports :: [Import]
...
-          mapM ((\mod -> do let tomlfile = addExtension (dropExtension $ moduleFile mod) "toml"
-                            exists <- doesFileExist tomlfile
-                            if exists
-                               then do toml_code <- readFile tomlfile
-                                       return $ Just $ pp toml_code
-                               else return Nothing))
+    -- Collect '.rs' and '.toml' files associated with each module.
+    rs <- M.fromList <$>
+          mapM (\mod -> do
+                   let rsfile = addExtension (dropExtension $ moduleFile mod) "rs"
```
mbudiu-vmw, line 157: common subexpression `(dropExresion $ moduleFile mod)`

src/Language/DifferentialDatalog/Module.hs
```diff
@@ -121,57 +132,59 @@ stdImports :: [Import]
...
+                   let fbfile = addExtension (dropExtension $ moduleFile mod) ".flatbuf.rs"
+                   let tomlfile = addExtension (dropExtension $ moduleFile mod) "toml"
+                   rs_exists <- doesFileExist rsfile
+                   flatbuf_exists <- doesFileExist fbfile
+                   toml_exists <- doesFileExist tomlfile
+                   rs_code <- if rs_exists
```
mbudiu-vmw, line 163: this also looks like a little function

test/antrea/test-antrea.sh
```diff
@@ -4,7 +4,15 @@
 
 set -e
 
-stack install
+# When running in CI, the DDlog compiler should be preinstalled by the build stage.
```
mbudiu-vmw, line 7: does this save much?

test/datalog_tests/json_test.dl
```diff
@@ -163,7 +163,7 @@ JsonTest(struct_with_map1(),
 
 typedef U64FromString = U64FromString {
     x: string,
-    #[rust="serde(with=\"crate::json::serde_string\")"]
+    #[rust="serde(with=\"types__json::serde_string\")"]
```
mbudiu-vmw, line 166: this is a subtle user-facing change. I wonder where this has to be documented.

mbudiu-vmw (approved):

lib/ddlog_std.dl
```diff
@@ -291,7 +288,7 @@ function map_err(res: Result<'V, 'E1>, f: function('E1): 'E2): Result<'V, 'E2> {
  * Range; all values starting at 'from' and ending at 'to'
  * inclusively, in steps of 'step'.  'step' must be positive.
  */
-extern function range(from: 'A, to: 'A, step: 'A): Vec<'A>
+//extern function range(from: 'A, to: 'A, step: 'A): Vec<'A>
```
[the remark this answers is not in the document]
ryzhyk: DDlog no longer allows unimplemented extern functions (as it reexports them in the Rust API, which does not compile if the function isn't actually there).  Some time ago we decided to keep it there to make it explicit that it is now deprecated. I can remove it.

lib/ddlog_std.dl
```diff
@@ -806,7 +803,6 @@ extern function vec_push_imm(v: Vec<'X>, x: 'X): Vec<'X>
 extern function vec_contains(v: Vec<'X>, x: 'X): bool
 extern function vec_is_empty(v: Vec<'X>): bool
 extern function vec_nth(v: Vec<'X>, n: usize): Option<'X>
-extern function vec_set_nth(v: Vec<'X>, n: usize, value: 'X)
```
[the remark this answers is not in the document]
ryzhyk: I think it lacked the implementation and shouldn't have been there in the first place. There's  a `vec_update_nth`.

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -568,87 +601,316 @@ compile d_unoptimized specname modules rs_code toml_code dir crate_types = do
...
+mkCargoToml :: (?cfg::Config, ?specname::String, ?crate_graph::CrateGraph, ?modules::[DatalogModule])
+    => M.Map ModuleName (Doc, Doc, Doc) -> Crate -> (FilePath, Doc)
+mkCargoToml rs_code crate =
+    (filepath, code)
+    where
+    filepath = "types" </> (crateDirPath crate) </> "Cargo.toml"
+    code = "[package]"                                                                     $$
+           "name = \"" <> pp crate_name <> "\""                                            $$
+           "version = \"0.1.0\""                                                           $$
+           "edition = \"2018\""                                                            $$
```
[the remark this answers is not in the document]
ryzhyk: Yes, we are on the 2018 Rust edition, which is currently the latest.

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -568,87 +601,316 @@ compile d_unoptimized specname modules rs_code toml_code dir crate_types = do
...
+           ""                                                                              $$
+           "[features]"                                                                    $$
```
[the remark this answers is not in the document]
ryzhyk: Problem is, this is a function of the output of the crate decomposition algorithm and must be generated.

src/Language/DifferentialDatalog/Expr.hs
ryzhyk: For now we only convert function calls to statics.  Most literals and constant expressions are statically evaluated anyway by the Rust compiler.

test/antrea/test-antrea.sh
ryzhyk: Big time!

PR #851, opened 2020-12-10
Support FlatMap over groups.

ryzhyk:
`FlatMap(g)`, where `g` is of type `Group<>` crashed the compiler.
Implementing it required fixing the compiler, but also implementing
`IntoIterator` for `Group` in `ddlog_std.rs`.


mbudiu-vmw (approved):

mbudiu-vmw:
Some day we will have iterators and perhaps this won't be a special case anymore.

PR #852, opened 2020-12-10
Include transitive dependencies in Cargo.toml.

ryzhyk:
The new structure of generated Rust code precisely follows the DDlog
project structure, meaning that each Rust module only needs to import
modules imported by the corresponding DDlog module and hence each Rust
crate must import crates containing at least on dependency of a module
from this crate.

Turns out theere is one exception to this rule: the type inference
engine adds type annotations to generic functions, which may refer to
types defined in modules not directly imported by the current module.
To handle such cases without a more elaborate dependency analysis, we
simply add a transitive closure of dependencies, not just direct
dependencies, to each crate.  I don't think this will affect compilation
time, including incremental compilation, significantly.

### DIFF
--- a/src/Language/DifferentialDatalog/Util.hs
+++ b/src/Language/DifferentialDatalog/Util.hs
@@ -181,6 +181,37 @@ updateFile path content = do
                 renameFile tmppath path
        else writeFile path content
 
+-- | Update an entire directory tree, updating files whose contents has changed,
+-- creating new files and deleting files not in `files` list, but keeping
+-- files that haven't changed unmodified.
+updateDirectory :: FilePath -> M.Map FilePath String -> IO ()
+updateDirectory dir files = do
+    -- Prepend directory path to all file names.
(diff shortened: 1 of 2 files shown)


src/Language/DifferentialDatalog/Util.hs
```diff
@@ -181,6 +181,34 @@ updateFile path content = do
...
+    existing_files <- traverseDir dir
+    -- Files not in the tree.
+    let to_delete = filter (\f -> M.notMember f files') existing_files
+    mapM_ removeFile to_delete
+    -- Update or create files in the tree.
+    mapM_ (\(mpath, mtext) -> updateFile mpath mtext)
+          $ M.toList files'
+
+-- | Traverse directory tree, returning the list of files in it.
+traverseDir :: FilePath -> IO [FilePath]
```
mbudiu-vmw, line 201: no library function doing this?
anyway, it's small.

mbudiu-vmw (approved):

PR #858, opened 2020-12-11
Use `Cow<str>` in `struct Program`.

ryzhyk:
Reduce the memory cost of creating and cloning `struct Program` by
representing rule and relation descriptions as `Cow<str>` instead of
`String`.  We currently only store static strings in `Program`, so there
is no need to allocate them on the heap.  By using `Cow` we eliminate
the memory waste in our current use case, while making it possible to
use dynamically allocated strings in the future.

### DIFF
--- a/rust/template/differential_datalog/program/mod.rs
+++ b/rust/template/differential_datalog/program/mod.rs
@@ -25,6 +25,7 @@ use crate::{ddval::*, profile::*, record::Mutator};
 use arrange::{antijoin_arranged, ArrangedCollection, Arrangements, A};
 use fnv::{FnvHashMap, FnvHashSet};
 use std::{
+    borrow::Cow,
     collections::{hash_map, BTreeSet},
     fmt::{self, Debug, Formatter},
     iter,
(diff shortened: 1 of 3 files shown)


Kixiron:
This can also be applied to the `Vec`s [stored within the `Program` struct](https://github.com/vmware/differential-datalog/blob/master/rust/template/differential_datalog/program/mod.rs#L134-L135) as well as [`ProgNode::SCC`](https://github.com/vmware/differential-datalog/blob/master/rust/template/differential_datalog/program/mod.rs#L158) and the [`Relation` struct](https://github.com/vmware/differential-datalog/blob/master/rust/template/differential_datalog/program/mod.rs#L231-L234)

mbudiu-vmw (approved):

ryzhyk:
> This can also be applied to the `Vec`s [stored within the `Program` struct](https://github.com/vmware/differential-datalog/blob/master/rust/template/differential_datalog/program/mod.rs#L134-L135) as well as [`ProgNode::SCC`](https://github.com/vmware/differential-datalog/blob/master/rust/template/differential_datalog/program/mod.rs#L158) and the [`Relation` struct](https://github.com/vmware/differential-datalog/blob/master/rust/template/differential_datalog/program/mod.rs#L231-L234)

Good point.

PR #864, opened 2020-12-14
Module.hs: Fix self-import check.

### LINKED ISSUE #860: DDlog reports incorrect errors on self-importation
If a module imports itself, instead of giving a "recursive module" error ddlog gives a "duplicated relation, defined in foo.dl and foo.dl" error

ryzhyk:
A module is not allowed to import itself (otherwise, all local declarations
will be duplicated).  The corresponding check was performed too late in
the control flow and was never reached.

This fixes #860.

### DIFF
--- a/src/Language/DifferentialDatalog/Module.hs
+++ b/src/Language/DifferentialDatalog/Module.hs
@@ -53,7 +53,7 @@ import System.FilePath.Posix
 import Data.List
 import Data.List.Split
 import Data.Maybe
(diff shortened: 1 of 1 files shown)


mbudiu-vmw (approved):
No test case?

ryzhyk:
> No test case?

Unfortunately, we don't have an infra for multi-file negative tests (which is how this bug was not detected in the first place).  We should do something shell-based.

PR #869, opened 2020-12-16
Refactored internment module and fixed some clippy lints

Kixiron:
* Adds some docs and `From` implementations to the `internment` module
* Adds `#[automatically_derived]` attributes to some macros to disable some lints

### DIFF
--- a/lib/internment.rs
+++ b/lib/internment.rs
@@ -10,12 +10,6 @@ use std::{
 
 /// An atomically reference counted handle to an interned value
 ///
-/// While this type is thread safe, it may cause logical bugs
-/// due to the underlying internment cache being a global variable,
-/// so don't rely on the existence or pre-internment of a variable
-/// since there's no way to know whether it will or will not exist
-/// within a threaded context.
-///
 /// The `PartialOrd` and `Ord` implementations for this type do not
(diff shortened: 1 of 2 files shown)


lib/internment.rs
```diff
@@ -1,190 +1,281 @@
...
+use serde::{de::Deserializer, ser::Serializer};
+use std::{
+    cmp::{self, Ordering},
+    fmt::{Debug, Display, Formatter, Result as FmtResult},
+    hash::Hash,
+};
+
+/// An atomically reference counted handle to an interned value
+///
+/// While this type is thread safe, it may cause logical bugs
```
ryzhyk, line 13: Thanks for the PR!

I don't understand this paragraph. How can one rely on a variable being pre-interned? What kind of usage pattern are you trying to prevent here?

Kixiron: It's just letting the user know that it's a global store of all interned items

ryzhyk: How can this lead to logical bugs?

The rest of the PR looks great, but I feel that this comment needs to be removed or rephrased. Perhaps just pointing to the [`arc_interner`](https://crates.io/crates/arc-interner) crate documentation is good enough.

ryzhyk (approved):

PR #872, opened 2020-12-17
Split tuples, arrays and the C API off from the records file and refactored them

Kixiron:
* Made `differential_datalog/records.rs` into a module
* Pulled the C API out of it, refactored a bit and put it into a feature gated file
* Rewrote the tuple macro, added 1-tuples and moved it to a file
* Refactored the array macro and moved it to a file
* Added some utility functions to the `Record` enum


ryzhyk (approved):
This is great, thanks for all the cleanups! I fixed regressions and rebased.  In addition to using a 1.48 feature your code is using some other post 1.41 features, so I decided to advance the earliest supported version of Rust to 1.47. I will push that change to this branch once I'm done uploading a new 1.47-based docker image (my upload speed sucks), and testing with it.

ryzhyk:
PS. I noticed that git diff is useless when renaming files is involved.  It'd make reviewing a lot easier if changes that involve renaming files were submitted as separate commits to make it easier to identify what's actually changed in the code.

Kixiron:
I'm sorry about that, I thought I had split it into a separate commit

ryzhyk:
C API in `record/c_api.rs` does not get exported in the generated `.so` file, most likely due to this bug: https://github.com/rust-lang/rust/issues/50007. It does get exported if I merge the C API back into `record/mod.rs`. Any objections to that change?

Kixiron:
That's very unfortunate, no objections from me

PR #875, opened 2020-12-18
Added the FromRecord and IntoRecord derive macros

Kixiron:
* Created a derive macro for the `FromRecord` and `IntoRecord` traits to allow end users to easily implement `FromRecord` and `IntoRecord` automatically without any boilerplate. Declarative macros exist within `differential_datalog`, but they're difficult to learn, use and must be manually updated and so are not ideal as a user-facing API (Whether the derives should be used in generated code I'll leave to your judgment)

These are the first two of (hopefully) 3 macros if all goes well: `FromRecord`, `IntoRecord` and `Mutator`


ryzhyk:
Apologies about the delay. Will try to review this and other PRs in the next couple of days.

Cargo.toml
```diff
@@ -1,9 +1,11 @@
 [workspace]
 members = [
     "rust/template",
+    "rust/template/ovsdb",
     "rust/template/cmd_parser",
-    "rust/template/differential_datalog",
+    "rust/template/ddlog_derive",
```
ryzhyk, line 6: Why does this have to be a separate crate?

ryzhyk (commented):
Looks awesome.  I have no experience with procedural macros, so can't really write a competent review, but I notice that there are no tests here. I can add some tests, and also change `Compile.hs` to use this instead of my existing macros to implement `FromRecord` and friends.  However, I won't be able to fix any issues that come up in the process.

Kixiron:
I'd wait on that, I've been working on tests using [`trybuild`](https://github.com/dtolnay/trybuild) as well as the remaining two derives

ryzhyk:
> I'd wait on that, I've been working on tests using [`trybuild`](https://github.com/dtolnay/trybuild) as well as the remaining two derives. Procs just tend to be a bit tedious to write, so things move a bit slower than normal

I converted the PR to a draft reflecting that it is not yet ready for review.  Thanks for continuing to work on this!

Cargo.toml
Kixiron: Procedural macros have to be separate crates (that can only export proc/derive macros) since they're compiled and ran at compile time

Kixiron:
Are there any other traits that require boilerplate for users to implement besides `FromRecord`, `IntoRecord` and `Mutator`?

And on the generated code, would you like tuple structs/variants to generate positional records or named records, and do you want `FromRecord` derives to generate the code for both `positional -> struct` and `named -> struct` or just `implemented by IntoRecord -> struct` (`FromRecord` on a tuple struct only generating the code for positional records, for instance)

Kixiron:
And how should `Record::Serialized` be handled, should that be an option via attribute or something? (I think it should actually be `Record::Serialized(Name, Vec<u8>)` instead of a `String`, but that's tangential to this). The problem I see occurring is people not wanting to depend on `serde_json` + whatever future serialization formats are implemented, one method I see working is something like this

```rust
#[derive(FromRecord)]
#[ddlog(deserialize = "\"json\", serde_json::from_str")]
struct Foo {}
```

This is a bit tricky of a problem to attack in a modular way, so I'm open to suggestions

This may also be a bit much for the initial pr and I don't think it should block, so I'd prefer brainstorming a bit here and then implementing it at a later time

ryzhyk:
> Are there any other traits that require boilerplate for users to implement besides `FromRecord`, `IntoRecord` and `Mutator`?

There's also `DDValConvert`, but you already took care of that.  We currently implement `Serialize` and `Deserialize` for all types using procedural macros from serde. So I guess `From/IntoRecord` and `Mutator` cover everything.

> And on the generated code, would you like tuple structs/variants to generate positional records or named records, and do you want `FromRecord` derives to generate the code for both `positional -> struct` and `named -> struct` or just `implemented by IntoRecord -> struct` (`FromRecord` on a tuple struct only generating the code for positional records, for instance)

- I guess it's ok for tuple structs to only generate positional records.
- `FromRecord` on normal structs should work for both positional and named records
- `FromRecord` on tuple structs should only work for positional records

ryzhyk:
> And how should `Record::Serialized` be handled, should that be an option via attribute or something? 

At the moment, all use cases for `Record::Serialized` only require support `FromRecord`, i.e., so I don't think we need an attribute.  The idea is that any type can be deserialized from any format specified at runtime.

> (I think it should actually be `Record::Serialized(Name, Vec<u8>)` instead of a `String`, but that's tangential to this).

`Record` is meant to be a human-readable representation, hence `String`, but I guess you're right, once we start using serialized representations they can just as well be binary.

>  The problem I see occurring is people not wanting to depend on `serde_json` + whatever future serialization formats are implemented, one method I see working is something like this
> 
> ```rust
> #[derive(FromRecord)]
> #[ddlog(deserialize = "\"json\", serde_json::from_str")]
> struct Foo {}
> ```

Associating a single representation with a type does not seem right. We already have use cases where the same type is deserialized from both JSON and bincode (although the latter does not currently go through `Record`).  The current design keeps serialization format orthogonal to the type signature, while making it possible to introduce new formats in the future in the spirit of `serde`.  As for depending on `serde_json`, we can in principle add compiler switches to specify what formats should be compiled into the DDlog binary.  E.g., if the user only needs bincode, they don't need to depend on `serde_json`. This will require some form of dynamic dispatch, e.g., we can have a static map from format names ("json", "bincode", etc), to their corresponding deserializers.

> This is a bit tricky of a problem to attack in a modular way, so I'm open to suggestions
> 
> This may also be a bit much for the initial pr and I don't think it should block, so I'd prefer brainstorming a bit here and then implementing it at a later time

Agreed. It'll be great if the current functionality can be preserved by procedural macros.  Hopefully future improvements shouldn't be too difficult.

Kixiron:
The idea behind the attribute was that you could add an additional one for each additional format you wanted
```rust
#[derive(FromRecord)]
#[ddlog(
    deserialize = "\"json\", serde_json::from_str",
    deserialize = "\"cbor\", cbor::from_slice",
)]
struct Foo {}
```

ryzhyk:
> The idea behind the attribute was that you could add an additional one for each additional format you wanted
> 
> ```rust
> #[derive(FromRecord)]
> #[ddlog(
>     deserialize = "\"json\", serde_json::from_str",
>     deserialize = "\"cbor\", cbor::from_slice",
> )]
> struct Foo {}
> ```

Sure, but that still implies that the implementer of the data type knows what formats it will need to support, which is unnecessary and be a serious limitations, e.g., for library types. `serde` enables clean (at least in theory) separation of concerns between type definition and its representation.  Why would you want to give that up?

Kixiron:
This still fully utilizes serde, it just allows opting into data formats

ryzhyk:
> This still fully utilizes serde, it just allows opting into data formats

Which is a good idea, I just think this has to be a build-time option, not an attribute of a data type.

Kixiron:
As I've been writing the `Mutator` derive, I'm not sure that the previous implementation is correct
In a simplified form, this is the `decl_mutator_struct!` macro:
```rust
impl Mutator<$Type> for Record
where
    Record: Mutator<$Generics>,
    $Generics: FromRecord,
{
    fn mutate(&self, _x: &mut $Type) -> Result<(), String> {
        match self {
            Record::PosStruct(_, _args) => {
                #[allow(unused_mut)]
                let mut index = 0;

                $(
                    if index == _args.len() {
                        return Err(format!("Positional struct mutator does not contain all elements"));
                    }

                    let arg_upd = &_args[index];
                    index += 1;

                    <dyn Mutator<$FieldType>>::mutate(arg_upd, &mut _x.$Field)?;
                )*

                if index != _args.len() {
                    return Err(format!("Positional struct mutator has too many elements"));
                }
            },

            Record::NamedStruct(_, _args) => {
                $(
                    if let Some(arg_upd) = arg_find(_args, stringify!($Field)) {
                        <dyn Mutator<$FieldType>>::mutate(arg_upd, &mut _x.$Field)?;
                    }
                )*
            },

            _ => {
                return Err(format!("not a struct {:?}", *self));
            }
        }

        Ok(())
    }
}
```
This seems incorrect for a number of reasons:
- It does not check that `constructor` is the same as the struct's constructor, meaning that two different structs with the same fields can mutate each other
- It does not check for malformed arg numbers until *after* mutations have been applied to the `&mut _` parameter, meaning that the passed-in element can be left in a malformed state

Should I fix these issues?

ryzhyk:
>     * It does not check that `constructor` is the same as the struct's constructor, meaning that two different structs with the same fields can mutate each other

The mutate semantics is a bit subtle.  Imagine a type with many constructors where all constructors have some common fields.  The mutate API allows the client to change the common fields without specifying the actual constructor name, which they may not even know (just like they do not need to know the values of other fields they don't want to modify).

>     * It does not check for malformed arg numbers until _after_ mutations have been applied to the `&mut _` parameter, meaning that the passed-in element can be left in a malformed state

Yep, you're right.

>     * Missing fields from `NamedStruct` are simply ignored

This is not a bug.  The idea is that the caller may want to modify only some of the fields and does not want to touch the rest (and may not even know their value).
 
> Additionally, the `arg_extract()` function used within `FromRecord` uses `Default::default()` to create a dummy value in the case that an argument by the given name can't be found, which seems very wrong to me

I guess the idea is that the user only needs to specify fields whose values are non-zero.  But I agree, this seems way too liberal.  I don't object to a stricter implementation.

> Should I fix these issues?

If you have time, thanks!

Kixiron:
So you'd prefer record mutation to have a sort of nominal type system where for named records the number of fields must be equal?

ryzhyk:
> So you'd prefer record mutation to have a sort of nominal type system where 

Right, Records in general behave that way.  You can also think of mutations as functions represented as data (in case of structs, every field of the mutator struct describes a change to the corresponding field of the mutated value).

> for named records the number of fields must be equal?

For named records, mutator field names must match a subset of fields in the mutated values, the number of fields does not necessarily have to be equal.  For positional records, the number of fields must be equal, i.e., a positional record cannot selectively mutate only a subset of fields.  At least, this is the current semantics.

rust/template/ddlog_derive/src/mutator.rs
```diff
@@ -0,0 +1,250 @@
...
+    quote! {
+        match self {
+            differential_datalog::record::Record::PosStruct(constructor, args)
+                if args.is_empty() => {},
+
+            differential_datalog::record::Record::NamedStruct(constructor, args)
+                if args.is_empty() => {},
+
+            error => {
+                return std::result::Result::Err(std::format!("not a struct {:?}", error));
```
ryzhyk, line 66: The error message is not correct if `self` is a struct with >0 fields.

rust/template/ddlog_derive/src/mutator.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    let mutator = quote! {
+        match self {
+            differential_datalog::record::Record::PosStruct(constructor, args)
+                if args.len() == #num_fields => {
+                    #( #field_mutations )*
+                },
+
+            error => {
+                return std::result::Result::Err(std::format!("not a struct {:?}", error));
```
ryzhyk, line 107: same issue here

ryzhyk (commented):
This is some hardcore stuff! Let me try using this in the compiler.

PR #881, opened 2020-12-24
Use the latest timely and DD code.

ryzhyk:
Upgrade DDlog to use the latest master branch of timely and DD.
Surprisingly, the required change is very small: a slight change in the
trace API, plus, since DD now natively supports 16-bit timestamps, we no
longer need the custom TS16 type.


mbudiu-vmw (approved):
These urls are repeated many times, would be nice to have them in some config.

PR #886, opened 2020-12-30
Concurrent improvements

Kixiron:
* Switches ddlog channels to use `crossbeam_channel` (mainly done to make `Sender<T>` `Send`, will likely also bring performance improvements)
* Remove `Mutex<Box<_>>` from `Relation::change_cb`
* Made callback closures now use `Fn()` instead of `FnMut()`, this more accurately reflects what should be happening within them and allows us to locklessly interact with them
* Removed a couple `Mutex`s from within various update handlers, less locks should lead to less runtime overhead


ryzhyk:
Neat!

PR #887, opened 2020-12-31
Hygiene improvements to Mutator

Kixiron:
* Made the names used within the `Mutator` macro use mixed spans so that they're hygienic, otherwise any fields with the name `args` would clash with the `args` ident declared in the macro, causing compile errors
* Added tests for this as well as some code ripped straight from what ddlog actually generates just for a real-world test case


ryzhyk (approved):
Thank you!  Do we have the same issue with `FromRecord` and `IntoRecord`?

Kixiron:
Thankfully not since there's no binding of variables we use exclusively, only binding to struct fields in construction or destruction

PR #888, opened 2021-01-03
Optimized string interning and removed some extra mutexes

Kixiron:
* Removed some unneeded mutexes from the startup routine that are no longer needed after the switch to `crossbeam-channel`
* Switched from a DIY string interner to `lasso` with a `fxhash` hasher, this removes some mutex contention while also improving our memory footprint and interning strings faster
* Implemented `Copy` for `Option<T>` where `T` is `Copy`

### DIFF
--- a/lib/ddlog_std.rs
+++ b/lib/ddlog_std.rs
@@ -200,6 +200,8 @@ pub fn bigint_pow32(base: &ddlog_bigint::Int, exp: &u32) -> ddlog_bigint::Int {
 }
 
 // Option
+impl<T: Copy> Copy for Option<T> {}
+
 pub fn option2std<T>(x: StdOption<T>) -> Option<T> {
     match x {
         StdOption::None => Option::None,
--- a/lib/intern.rs
+++ b/lib/intern.rs
@@ -1,104 +1,85 @@
-use differential_datalog::record;
-use differential_datalog::record::*;
+use differential_datalog::record::{self, Record};
+use fxhash::FxBuildHasher;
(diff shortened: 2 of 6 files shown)


ryzhyk:
Thanks, @Kixiron !
Converted to draft, as tests are still failing.

PS. I am in the process of refactoring `differential_datalog/record` and `api` modules.  It's probably best not to touch those parts in the next few days to avoid conflicts (just a heads up, I am not referring to this PR).

ryzhyk:
It's nice to see some ugly mutex-related code disappear, thanks! 

Regarding string interner. I don't mind switching to lasso (although I won't pretend to understand any of the equestrian jargon in the description of the crate), but just to be clear, this module is only used in one set of tests right now and is generally not very useful in real applications as it never deallocates strings (I think it's the same for lasso, right?), so I'm just curious why you are interested in improving it?

Kixiron:
RSlint interns a lot of strings for things like variables, and lasso gives a nice little perf boost for that

Kixiron:
I haven't benchmarked it, but I would imagine it's faster due to lasso's memory allocation scheme (a sharded arena), fast hashing scheme and the lack of reference counting, atomic or otherwise (it also has the bonus of using `NonZeroU32`s for keys which allows niche optimization). However as you said, it does lack the ability to free strings incrementally in the way `istring` does. An incrementally freeing arena of some sort has been on my bucket list for a while since it's invaluable within a persistent scenario like a LSP

Issue #794, opened 2020-10-30
RFC: API to run DDlog instance in a separate process.


ryzhyk:
Motivation
------------

- Run multiple instances of the same program on behalf of multiple tenants while enforcing resource isolation between tenants.
- Run different DDlog programs for the same or multiple tenants (DDlog-generated '.so' libraries export the same set of symbols; therefore multiple libraries cannot be easily loaded in the same process).

Design
---------

- This will probably be implemented using the D3log infrastructure.
- We may even want to support remote DDlog instances.

**details TBD**

Kixiron:
It may be awkward in the actual implementation, but I think basing the api off of an in-process model is the best route for having an api that feels good to work with. Ideally you could totally swap backends from ddlog to d3log with little to no changes in code, even though that's a slightly unrealistic goal to have

mbudiu-vmw:
>  statically typed API will require the client to additionally link against a DDlog-generated library.
Do you intend to support the case of client that loads multiple such libraries at once?

mbudiu-vmw:
> Starting a DDlog instance.

Maybe use ssh for this? Are there any requirements for the environment of a ddlog process? I can't really think of any myself - if you can run a rust executable you should be able to run ddlog.

mbudiu-vmw:
> Split trait DDlog into statically and dynamically typed APIs.

You should probably start with this piece, since it has nothing to do with the distributed execution.

mbudiu-vmw:
> The DDlog protocol, implemented over the ZeroMQ protocol

ZeroMQ seems to offer much more than you need. For example, I think you will never need multicast.

> heartbeat messages

If you use an existing framework it should take care of this. You don't want to implement this by hand.

ryzhyk:
> > statically typed API will require the client to additionally link against a DDlog-generated library.
> 
> Do you intend to support the case of client that loads multiple such libraries at once?

Not at the moment, but this is in any case orthogonal to this RFC.

mbudiu-vmw:
How about having two separate traits for producers (who supply deltas) and consumers (who implement the callbacks)? This way the consumers don't have to be in the same process and you can make a chain of processes.

ryzhyk:
> > Starting a DDlog instance.
> 
> Maybe use ssh for this? Are there any requirements for the environment of a ddlog process? I can't really think of any myself - if you can run a rust executable you should be able to run ddlog.

[...]

PR #889, opened 2021-01-07
DDlog API refactoring Part 1.

ryzhyk:
This is the first step towards making the DDlog API generic so that we
can have multiple implementations, including one that talks to a DDlog
server over the network (#794).

In this commit:

- Split `trait DDlog` into `trait DDlogUntyped` and `trait DDlogTyped`.
  The former works with data represented as records, the latter
  exports methods that work with `DDValue`'s.
- Move a bunch of methods that were in `HDDlog`, but not in `trait DDlog`
  to either one of the above traits or into one of several new traits:
  `DDlogProfiling`, `DDlogDump`, `DDlogInventory`.  This way the entire
  DDlog API is now abstracted away.
- Change method signatures to eliminate any generics.  This way we will
  be able to implement dynamic dispatch for the DDlog API (i.e., pass
  references to a DDlog program as `&dyn DDlogXXX`) in the future.

### DIFF
--- a/rust/template/differential_datalog/src/ddlog.rs
+++ b/rust/template/differential_datalog/src/ddlog.rs
@@ -1,7 +1,11 @@
+//! Traits that encapsulate a running DDlog program.
+
 use std::collections::btree_set::BTreeSet;
-use std::fmt::Debug;
+use std::collections::BTreeMap;
+#[cfg(feature = "c_api")]
+use std::ffi::CStr;
+use std::io;
 use std::iter::Iterator;
-use std::ops::Deref;
 
 use crate::ddval::DDValue;
 use crate::program::IdxId;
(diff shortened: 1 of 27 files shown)


rust/template/src/api/c_api.rs
```diff
@@ -35,45 +35,88 @@ use std::os::unix::io::{FromRawFd, IntoRawFd, RawFd};
...
-    if tname.is_null() {
+pub unsafe extern "C" fn ddlog_get_table_id(
+    prog: *const HDDlog,
+    tname: *const raw::c_char,
+) -> libc::size_t {
+    if prog.is_null() || tname.is_null() {
         return libc::size_t::max_value();
     }
 
+    let prog = Arc::from_raw(prog);
```
Kixiron, line 46: `ManuallyDrop::new(Arc::from_raw(..))` is less error prone and more clear than calling from_raw and later into_raw

rust/template/src/api/c_api.rs
```diff
@@ -35,45 +35,88 @@ use std::os::unix::io::{FromRawFd, IntoRawFd, RawFd};
...
-    match HDDlog::get_table_cname(tid) {
+pub unsafe extern "C" fn ddlog_get_table_name(
+    prog: *const HDDlog,
+    tid: libc::size_t,
+) -> *const raw::c_char {
+    if prog.is_null() {
+        return ptr::null();
+    }
+
+    let prog = Arc::from_raw(prog);
```
Kixiron, line 66: Same as above

rust/template/src/api/c_api.rs
```diff
@@ -35,45 +35,88 @@ use std::os::unix::io::{FromRawFd, IntoRawFd, RawFd};
...
-pub unsafe extern "C" fn ddlog_get_index_id(iname: *const raw::c_char) -> libc::size_t {
+pub unsafe extern "C" fn ddlog_get_index_id(
+    prog: *const HDDlog,
+    iname: *const raw::c_char,
+) -> libc::size_t {
+    if prog.is_null() {
+        return libc::size_t::max_value();
+    }
+
+    let prog = Arc::from_raw(prog);
```
Kixiron, line 86: Arcs again

rust/template/src/api/c_api.rs
```diff
@@ -35,45 +35,88 @@ use std::os::unix::io::{FromRawFd, IntoRawFd, RawFd};
 use std::os::windows::io::{FromRawHandle, IntoRawHandle, RawHandle};
 
 #[no_mangle]
-pub unsafe extern "C" fn ddlog_get_table_id(tname: *const raw::c_char) -> libc::size_t {
-    if tname.is_null() {
+pub unsafe extern "C" fn ddlog_get_table_id(
+    prog: *const HDDlog,
```
Kixiron, line 39: To be truly FFI safe while still allowing consumers to take advantage of ddlog we probably want the type used by consumers to be an `Arc<Box<dyn DDlog>>` or similar so that any ddlog runtime can be used while keeping it thread safe since dynamically dispatched objects use fat pointers which aren't ffi safe

Kixiron, line 39: If you want to be clear about the contract that functions expect, you can use `NonNull<HDDlog>` to signify non-null pointers and `Option<NonNull<T>>` for nullable ones. This can be done in both parameter and return positions and allows the backing rust code niceties like `?`

ryzhyk:
@Kixiron , thanks for the review. I will incorporate your suggestions in the next commit, which will switch the C API over to use dynamic dispatch.

Regarding `Arc`'s, I actually think that we can get rid of `from_raw` altogether.  We should just cast `*HDDlog` to `&HDDlog` (using `&*`) and work with that, since most APIs don't touch the reference count.  I guess I just did not fully understand how `into_raw` works for `Arc`s when writing that code.

Kixiron:
Sounds good to me!

mbudiu-vmw (approved):

rust/template/src/api/c_api.rs
Kixiron: It's undefined behavior, same as what happens with the current code. If nulls are expected (which I don't see why why would be) then `Option<NonNull<T>>` does the same thing with the ability to take a null value as well

ryzhyk:
@convolvatron
@mbudiu-vmw 
@blp

rust/template/differential_datalog/src/ddlog.rs
```diff
@@ -60,34 +93,33 @@ pub trait DDlog: Debug {
     fn transaction_rollback(&self) -> Result<(), String>;
 
     /// Apply a set of updates.
-    fn apply_updates<V, I>(&self, upds: I) -> Result<(), String>
-    where
-        V: Deref<Target = UpdCmd>,
-        I: Iterator<Item = V>;
+    fn apply_updates_untyped(&self, upds: &mut dyn Iterator<Item = UpdCmd>) -> Result<(), String>;
```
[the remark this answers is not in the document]
blp: Dynamic?

rust/template/differential_datalog/src/replay.rs
```diff
@@ -48,254 +54,339 @@ where
...
-    C: DDlogConvert,
-{
-    match rel_ident {
-        RelIdentifier::RelName(rname) => Some(rname.as_ref()),
-        RelIdentifier::RelId(id) => C::relid2name(*id),
+/// DDlog API implementation that records each command passing through it and
+/// forwards it to the next handler in the chain.
+///
+/// Error handling:
+/// If the next handler is specified, we invoke it, whether recording succeeded
```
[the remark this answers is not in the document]
blp: I can't spot a missing verb.

ryzhyk:
I will close this PR and re-open it from another branch. I incorporated most feedback.  Most importantly, I renamed

- `trait DDlogUntyped` -> `trait DDlogDynamic`.
- `trait DDlogTyped` -> `trait DDlog` (this trait extends `DDlogDynamic`).

PR #895, opened 2021-01-14
Fixes

### LINKED ISSUE #893: Confusing error message due to type constraint ordering
Consider this DDlog program:
```
import json

function f(): JsonValue {
   var v: Map<string, JsonValue> = map_empty();
   JsonObject{v}
}
```

Here is the produced error:
```
ddlog: error: x.dl:4:4-4:9: Type mismatch:
expected type: string
actual type: internment::Intern<string>
in
expected type: ddlog_std::Map<string,json::JsonValue<>>
actual type: ddlog_std::Map<internment::istring<>,json::JsonValue<>>
in
expression 'var v'
   var v: Map<string, JsonValue> = map_empty();
   ^^^^^
```

I suspect that this happens because the type constraints are solved in reverse order with respect to the program source ordering, and the first unification that fails is for the first statement.

mbudiu-vmw:
A set of small fixes. See the individual commits.
Also fixes #893 for a major usability improvement in my opinion.


sql/install-ddlog-jar.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 
+stack install
```
ryzhyk, line 3: Please make this conditional on `${IS_CI_RUN}`

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -320,7 +320,7 @@ deref (x, _, _)          = x
 mutref :: (Doc, EKind, ENode) -> Doc
 mutref (x, EReference, _)  = x
 mutref (x, ENoReturn, _)   = x
-mutref (x, _, _)           = parens $ "&mut" <> x
+mutref (x, _, _)           = parens $ "&mut" <+> x
```
ryzhyk, line 323: oops

ryzhyk (commented):
I'll try to figure out what's going on with the vec test.

PR #896, opened 2021-01-14
Support for ephemeral streams.

ryzhyk:
This commit adds support for ephemeral relations whose contents gets
discarded at the end of each transaction.  This is useful in streaming
applications that operate on a mix or persistent tables whose contents
accumulates over the lifetime of the application and transient data
that gets transformed and/or enriched and sent out.  Such transient data
must then be discarded; otherwise it accumulates unboundedly, eventually
causing an out-of-memory situation.

One way to support such streaming applications is to clear ephemeral
relations at the start of each transaction.  While this approach works,
it is expensive (DD ends up doing the same computation twice on insertion
and deletion, as well as arranging transient data just to discard it
immediately) and cumbersome, as the client has to issue additional
`clear` commands and handle delete notifications during commit.

We therefore introduce direct support for ephemeral data.  We do so by
extending the stream abstraction (in a non-backwards compatible way, but
I don't think it has many users yet, so this should not be an issue).
So far strems were just multisets whose contents is not tracked by the
DDlog runtime, so the `clear` command does not work for streams.  We
change the definition of streams so that streams store transient data
that gets discarded right after commit.

Internally the implementation uses the `lookup_map` operator from the
`dogsdogsdogs` sub-crate of DD.  `lookup_map` is a special form of
"streaming" `join` that joins an arrangement with a collection without
arranging the collection first.  The memory footprint of the operator in
steady state is 0.

We use `lookup_map` to implement relation-to-stream joins in DD.  Since
`lookup_map` does not track the accumulated state of the collection, but
only its changes in the last timestamp, it does not produce delete
outputs (i.e., outputs with negative weights); hence we must be careful
not to arrange the output collection produced by `lookup_map`, as such
an arrangement will grow unboundedly.  We enforce this by imposing
restrictions on how streams can be used in DDlog: (1) streams cannot be
aggregated, negated, or used in a recursive fragment of the program, (2)
an output of a rule that uses a stream in its body must be a stream.

See the new tutorial section on streams for more details.

Here is an example program that uses streams:

```
    input relation ZipCodes(zip: u32, city: string)
    input stream Parcel(zip: u32, weight: usize)

    // Add the name of the destination city to each parcel.
    output stream ParcelCity(zip: u32, city: string, weight: usize)
    ParcelCity(zip, city, weight) :-
        Parcel(zip, weight),
        ZipCodes(zip, city).
```

```
    echo Populating the zip code table.;
    start;
    insert ZipCodes(94022, "Los Altos"),
    insert ZipCodes(94035, "Moffett Field"),
    insert ZipCodes(94039, "Mountain View"),
    insert ZipCodes(94085, "Sunnyvale"),
    commit dump_changes;

    echo Feeding data to the Parcel stream;
    start;
    insert Parcel(94022, 2),
    insert Parcel(94085, 6),
    commit dump_changes;

    # DDlog output:
    #   Populating the zip code table.
    #   Feeding data to the Parcel stream
    #   ParcelCity:
    #   ParcelCity{.zip = 94022, .city = "Los Altos", .weight = 2}: +1
    #   ParcelCity{.zip = 94085, .city = "Sunnyvale", .weight = 6}: +1

    echo Send more parcels!;
    start;
    insert Parcel(94039, 10),
    insert Parcel(94022, 1),
    commit dump_changes;

    # DDlog output:
    #   Send more parcels!
    #   ParcelCity:
    #   ParcelCity{.zip = 94022, .city = "Los Altos", .weight = 1}: +1
    #   ParcelCity{.zip = 94039, .city = "Mountain View", .weight = 10}: +1
```


mbudiu-vmw (approved):
This was pretty quick.

doc/tutorial/tutorial.md
```diff
@@ -1859,23 +1857,133 @@ Output `multiset`s are more memory-efficient than `relation`s: internally the
...
+
+A better solution is to use the DDlog **stream** abstraction, intended to
+simplify working with ephemeral data, i.e., data that gets discarded after being
+used.  A stream is a relation that only contains records added during the
+current transaction.  When the transaction completes, all previously inserted
+records as well as any records derived from them disappear without a trace and
+the memory footprint of all streams in the program drops to zero.
+
+We re-declare the `Parcel` relation as a stream (by using the `stream` keyword
+instead of `relation`).  Any rule that uses a stream yields a stream; hence
```
Kixiron, line 1899: This seems incredibly limiting, since streams also can't be used in recursive rules and such this makes streams only applicable to a very small subset of use cases

doc/tutorial/tutorial.md
```diff
@@ -1859,23 +1857,133 @@ Output `multiset`s are more memory-efficient than `relation`s: internally the
...
+Note that although the stream becomes empty at the start of every transaction,
+DDlog does not output delete records (i.e., records with negative weights) for
+streams.
+
+In general, streams can be used in rules just like normal relations.  They can
+be joined with other relations, filtered, and flat-mapped.  There are also some
+restrictions on streams.  Most importantly, streams cannot be part of a
+recursive computation.  In addition streams currently cannot be aggregated using
+the `group_by` operator, or used as one of the inputs to an antijoin.  These
+constraints may be removed in future versions of DDlog.
```
Kixiron, line 1960: Again, very limited. Recursion and negation are some of the cornerstones of datalog so this means streams are only available to a small set of use cases

rust/template/differential_datalog/Cargo.toml
```diff
@@ -13,6 +13,7 @@ c_api = []
 [dependencies]
 #differential-dataflow = "0.11.0"
 differential-dataflow = { git = "https://github.com/ddlog-dev/differential-dataflow", branch = "ddlog-2" }
+dogsdogsdogs = { git = "https://github.com/ddlog-dev/differential-dataflow", branch = "ddlog-2" }
```
Kixiron, line 16: Unrelated to this pr, d3 looks very promising for a lot of the joins that ddlog does, are there plans to use it elsewhere within the runtime?

src/Language/DifferentialDatalog/Rule.hs
```diff
@@ -222,3 +224,13 @@ ruleHeadIsRecursive d Rule{..} head_idx =
 ruleIsRecursive :: DatalogProgram -> Rule -> Bool
 ruleIsRecursive d rl@Rule{..} =
     any (ruleHeadIsRecursive d rl) [0 .. length ruleLHS - 1]
+
+-- | True iff the prefix of the rule of length 'n' produces a stream.
```
Kixiron, line 228:
```suggestion
-- | True if the prefix of the rule of length 'n' produces a stream.
```

src/Language/DifferentialDatalog/Rule.hs
```diff
@@ -222,3 +224,13 @@ ruleHeadIsRecursive d Rule{..} head_idx =
...
+rulePrefixIsStream :: DatalogProgram -> Rule -> Int -> Bool
+rulePrefixIsStream d rl n =
+    any (\rhs -> rhsIsLiteral rhs && (relIsStream $ getRelation d $ atomRelation (rhsAtom rhs)))
+        $ take n $ ruleRHS rl
+
+-- | True iff the body of the rule yields a stream.
```
Kixiron, line 234:
```suggestion
-- | True if the body of the rule yields a stream.
```

src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -356,8 +357,16 @@ atomValidate d ctx atom = do
...
+ruleRHSValidate d rl (RHSLiteral polarity atom) idx = do
     atomValidate d (CtxRuleRAtom rl idx) atom
+    let rel = getRelation d $ atomRelation atom
+    when (rulePrefixIsStream d rl idx) $ check d (relSemantics rel /= RelStream) (pos atom)
+        $ "Illegal join of two streams: stream relation '" ++ name rel ++ "' and the stream produced by the prefix of the rule before this literal. Streams can only be joined with regular non-stream relations."
+    when (relIsStream rel) $ check d polarity (pos atom)
+        $ "Illegal negation of a stream relation '" ++ name rel ++ "'. Only non-stream relations can be negated."
+    -- This is an implementation constraint.  We should be able to antijoin streams.
+    when (polarity == False) $ check d (not $ rulePrefixIsStream d rl idx) (pos atom)
+        $ "Unsupported antijoin. The prefix of the rule before this literal produces a stream. Antijoins with streams are currently not supported."
```
Kixiron, line 369: Nit, the wording is a bit harsh, maybe something like "attempted to join streams {stream1} and {stream2} which is not currently supported" would be better

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -1067,15 +1185,82 @@ impl Program {
...
+                        |(k, _), key| *key = k.clone(),
+                        move |v1, w1, v2, w2| (jfun(&v1.1, v2), w1 * w2),
+                        ().into_ddvalue(),
+                        ().into_ddvalue(),
+                        ().into_ddvalue(),
+                    )
+                    // Filter out `None`'s.
+                    // FIXME: We wouldn't need this if `lookup_map` allowed `output_func`
+                    // to return `Option`.
+                    .flat_map(|v| v)
```
Kixiron, line 1219: Nit, I think using the `identity` function is clearer but it's a matter of personal preference

doc/tutorial/tutorial.md
ryzhyk: The whole point of streams is that they have zero memory footprint.  Recursive computations in DD fundamentally require indexing stuff, so this is not an easy limitation to lift.  Having said that, you can have as much recursion as you like in your program, and then join the result of the recursive computation with a stream.  We also plan to add more things you can do with streams in the near future.  For example, we might support rolling up a stream into a "normal" relation that you can then use with recursion, etc.

rust/template/differential_datalog/Cargo.toml
ryzhyk: I have support for d3 implemented in a branch, which has now bit rotted and will need to be refreshed.  Unfortunately, while it works and the performance is basically the same as with regular DD operators, I never got much improvement in terms of memory footprint out of it.  The main reason is that, while you don't need to arrange every prefix of the rule anymore, you need to construct additional arrangements of individual relations in the rule.  In addition, d3 does not work with recursion, and I don't think I figured out how to express antijoins with it.  But this is all very use case-specific, so I agree it would be nice to support it.

doc/tutorial/tutorial.md
ryzhyk: PS. Limited as they are, streams speed things up a lot when they are applicable. We have some internal use cases that benefit from this.

ryzhyk:
@Kixiron , @mbudiu-vmw , thanks for your reviews.  I'll cleanup and merge this when I'm done putting out fires elsewhere.

src/Language/DifferentialDatalog/Rule.hs
mbudiu-vmw: "iff" is a standard abbreviation for "if and only if"

rust/template/differential_datalog/Cargo.toml
ryzhyk: > d3 does not work with recursion

Actually this is no longer true.  DD now contains differentiation and integration operators that can be used to implement delta queries inside recursive fragments of DD: [blog](https://github.com/frankmcsherry/blog/blob/master/posts/2020-02-15.md)

PR #898, opened 2021-01-15
Keep track of source position for constructor fields

### LINKED ISSUE #894: Error indicator points to the wrong source code
Consider this DDlog program:
```
typedef T = T{ x: bit<32> }

input relation Dummy(t: ())
Dummy(()).
output relation R(t: T)
R(x) :- Dummy(y), var x = T{ .z = 1 }.
```

Compiling it produces the following error:

```
ddlog: error: x.dl:6:35-6:37: Unknown field z
R(x) :- Dummy(y), var x = T{ .z = 1 }.
                                  ^^
```

Note that the error marker points to the initializer and not to the wrong field. This is particularly confusing when the initializer is itself a field of some other object.

mbudiu-vmw:
Fixes #894 
I added a new IdentifierWithPos object in the compiler, which may be useful in other contexts where today String is used, but I have only used it to fix the issue specified above.

### DIFF
--- a/src/Language/DifferentialDatalog/Parse.hs
+++ b/src/Language/DifferentialDatalog/Parse.hs
@@ -485,12 +485,12 @@ atom is_head = withPos $ do
                           else val
        return $ Atom nopos rname $ maybe val' (\b -> E $ EBinding (p1, p2) b val') binding
 
-anonarg = ((makeIdentifierWithPos ""),) <$> expr
-namedarg = (,) <$> (dot *> posIdent) <*> (reservedOp "=" *> expr)
+anonarg = ((IdentifierWithPos nopos ""),) <$> expr
+namedarg = (,) <$> (dot *> posVarIdent) <*> (reservedOp "=" *> expr)
 
 -- Identifier with position
-posIdent :: ParsecT String () Identity IdentifierWithPos
-posIdent = withPos $ makeIdentifierWithPos <$> varIdent
+posVarIdent :: ParsecT String () Identity IdentifierWithPos
+posVarIdent = withPos $ makeIdentifierWithPos <$> varIdent
 
 typeSpec = withPos $
             bitType
@@ -649,8 +649,8 @@ pterm = (withPos $
        <|> enumber)
       <?> "match term"
 
-anonpat = (makeIdentifierWithPos "",) <$> pattern
-namedpat = (,) <$> (dot *> posIdent) <*> (reservedOp "=" *> pattern)
+anonpat = (IdentifierWithPos nopos "",) <$> pattern
+namedpat = (,) <$> (dot *> posVarIdent) <*> (reservedOp "=" *> pattern)
 
 lhs = (withPos $
            eTuple <$> (parens $ commaSepEnd lhs)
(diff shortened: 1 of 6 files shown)


src/Language/DifferentialDatalog/Parse.hs
```diff
@@ -485,8 +485,12 @@ atom is_head = withPos $ do
...
        return $ Atom nopos rname $ maybe val' (\b -> E $ EBinding (p1, p2) b val') binding
 
-anonarg = ("",) <$> expr
-namedarg = (,) <$> (dot *> varIdent) <*> (reservedOp "=" *> expr)
+anonarg = ((makeIdentifierWithPos ""),) <$> expr
+namedarg = (,) <$> (dot *> posIdent) <*> (reservedOp "=" *> expr)
+
+-- Identifier with position
+posIdent :: ParsecT String () Identity IdentifierWithPos
+posIdent = withPos $ makeIdentifierWithPos <$> varIdent
```
ryzhyk, line 493: This should probably be called `posVarIdent`, you may want to create `IdentifierWithPos` for things like relation or type names.

src/Language/DifferentialDatalog/Syntax.hs
```diff
@@ -230,6 +232,22 @@ instance WithPos ArgType where
...
+instance Ord IdentifierWithPos where
+    compare (IdentifierWithPos _ n1) (IdentifierWithPos _ n2) = compare n1 n2
+
+instance WithPos IdentifierWithPos where
+    pos = idPos
+    atPos i p = i{idPos = p}
+
+instance PP IdentifierWithPos where
+    pp IdentifierWithPos{..} = pp idName
+
```
ryzhyk, line 250: I think we need `instance PP Name` and use `name` instead of `idName` everywhere.

src/Language/DifferentialDatalog/Syntax.hs
```diff
@@ -230,6 +232,22 @@ instance WithPos ArgType where
 instance PP ArgType where
     pp ArgType{..} = (if atypeMut then "mut" else empty) <+> pp atypeType
 
+data IdentifierWithPos = IdentifierWithPos { idPos :: Pos, idName :: String }
+makeIdentifierWithPos str = IdentifierWithPos nopos str
```
ryzhyk, line 236: I don't think it's a very good function name. It's more like `makeIdentifierWithOUTPos` :)

For better or worse, I tend not to use this type of helpers throughout the compiler. So perhaps just write `IdentifierWithPos nopos name`. I won't insist though.

ryzhyk:
@mbudiu-vmw , can we finalize this?  I think all that's left is to rebase this on master and re-run tests if you haven't done it yet.

mbudiu-vmw:
We could also possibly use the IdentifierWithPos in a few more places in the AST.

ryzhyk:
Let's merge this first

mbudiu-vmw:
Should we merge this?

ryzhyk:
Yes, assuming it has been rebased and tests are passing

PR #899, opened 2021-01-17
Workaround for OS X linker error.

ryzhyk:
`extern C` functions declared in `ddlog_log.rs` do not get exported on
Mac OS unless the program explicitly does `import log`.  This is most
likely due to Rust's arbitrary treatment of extern functions that are
not directly visible in the main crate of the program.  The fix is to
import `ddlog_log` in `template/src/lib.rs`.

### DIFF
--- a/rust/template/src/lib.rs
+++ b/rust/template/src/lib.rs
@@ -61,6 +61,12 @@ use serde::Deserializer;
 use serde::Serialize;
 use serde::Serializer;
 
+// This import is only needed to convince the OS X compiler to export
+// `extern C` functions declared in ddlog_log.rs in the generated lib.
+#[doc(hidden)]
+#[cfg(feature = "c_api")]
+pub use ddlog_log as hidden_ddlog_log;
+
 /// A default implementation of `DDlogConvert` that just forwards calls
 /// to generated functions of equal name.
 #[derive(Debug)]


rust/template/src/lib.rs
```diff
@@ -61,6 +61,10 @@ use serde::Deserializer;
 use serde::Serialize;
 use serde::Serializer;
 
+// This import is only needed to convince the OS X compiler to export
+// `extern C` functions declared in ddlog_log.rs in the generated lib.
+pub use ddlog_log as hidden_ddlog_log;
```
Kixiron, line 66: Maybe feature gate this under the `c_api` flag so it's not exported for non-ffi users

Kixiron, line 66:
```suggestion
#[doc(hidden)]
pub use ddlog_log;
```
A much better way of hiding extraneous documentation, it'll clean up the rustdoc output and will help to not confuse users

ryzhyk: good point

ryzhyk: Thanks, I'll add `doc(hidden)`, but renaming is still necessary to avoid conflict with another import below.

mbudiu-vmw (approved):

PR #900, opened 2021-01-18
Added ddlog benchmarks

Kixiron:
Added the `citations` benchmark, which is a real-world ddlog use case, it tests ddlog's parallelism and throughput under very large workloads. I also set up the framework for adding more benchmarks in the future, which I plan to do

Current Results:
```
citations/1 thread/100000 records
time:   [28.364 s 28.846 s 29.341 s]
thrpt:  [3.5355 Kelem/s 3.5716 Kelem/s 3.6108 Kelem/s]

citations/2 threads/100000 records
time:   [26.284 s 26.793 s 27.179 s]
thrpt:  [4.1461 Kelem/s 4.1977 Kelem/s 4.2499 Kelem/s]

citations/3 threads/100000 records
time:   [28.088 s 28.492 s 28.880 s]
thrpt:  [3.7379 Kelem/s 3.7988 Kelem/s 3.8680 Kelem/s]

citations/4 threads/100000 records
time:   [28.033 s 28.593 s 29.278 s]
thrpt:  [3.4992 Kelem/s 3.5749 Kelem/s 3.6543 Kelem/s]
```

### DIFF
--- a/rust/ddlog_benches/README.md
+++ b/rust/ddlog_benches/README.md
@@ -22,8 +22,8 @@ For info on writing benchmarks see the [criterion user guide] and the [criterion
 
 ### Other supported commands
 
-- `cargo make benchmark-twitter`: Only run the Twitter benchmarks
-- `cargo make benchmark-livejournal`: Only run the LiveJournal benchmarks
+- `cargo make bench-twitter`: Only run the Twitter benchmarks
+- `cargo make bench-livejournal`: Only run the LiveJournal benchmarks
 - `cargo make build-ddlog`: Only build the generated ddlog code
 - `cargo make download-data`: Download the datasets required for benchmarking
 
--- a/rust/ddlog_benches/README.md
+++ b/rust/ddlog_benches/README.md
@@ -1,0 +1,32 @@
+# ddlog_benches
+
+DDlog has a unique benchmarking scenario since it's a compiler, so benchmarking the programs
+it generates is a little more involved than normal, which is why we use [`cargo-make`] to automate
+the setup, build and bench process in a cross-platform way. To run the benchmarks in this crate,
+first install `cargo-make` with the following command (`--force` makes cargo install the latest version)
+
+```sh
+cargo install --force cargo-make
+```
+
+To run all benchmarks use the following command
+
+```sh
+cargo make benchmarks
+```
+
+When adding a benchmark, add the required ddlog code to `/ddlog`, any datasets to `/data` (preferably in a
+compressed format), add a module within `/src` containing your support code and the actual benchmark within
+`/benches` and finally add a `[[bench]]` entry into the `Cargo.toml` along with any required dependencies.
+For info on writing benchmarks see the [criterion user guide] and the [criterion docs].
+
+### Other supported commands
+
+- `cargo make benchmark-twitter`: Only run the Twitter benchmarks
+- `cargo make benchmark-livejournal`: Only run the LiveJournal benchmarks
+- `cargo make build-ddlog`: Only build the generated ddlog code
+- `cargo make download-data`: Download the datasets required for benchmarking
+
+[`cargo-make`]: https://github.com/sagiegurari/cargo-make
+[criterion user guide]: https://bheisler.github.io/criterion.rs/book/index.html
+[criterion docs]: https://docs.rs/criterion
--- a/rust/ddlog_benches/benches/twitter.rs
+++ b/rust/ddlog_benches/benches/twitter.rs
@@ -14,8 +14,8 @@ fn record_counts() -> impl Iterator<Item = usize> {
     (50_000..=200_000).step_by(50_000)
 }
 
-fn twitter(c: &mut Criterion) {
-    let mut group = c.benchmark_group("twitter");
+fn twitter_micro(c: &mut Criterion) {
+    let mut group = c.benchmark_group("twitter-micro");
     group.sampling_mode(SamplingMode::Flat);
     group.sample_size(10);
 
@@ -73,5 +73,5 @@ fn twitter_macro(c: &mut Criterion) {
     }
 }
 
-criterion_group!(benches, twitter, twitter_macro);
+criterion_group!(benches, twitter_micro, twitter_macro);
 criterion_main!(benches);
--- a/rust/ddlog_benches/benches/twitter.rs
+++ b/rust/ddlog_benches/benches/twitter.rs
@@ -1,0 +1,77 @@
+use criterion::{
+    black_box, criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion, SamplingMode,
+};
+use ddlog_benches::twitter;
+use std::ops::RangeInclusive;
+
+/// Benchmark all targets using 1, 2, 3, and 4 threads
+const DDLOG_WORKERS: RangeInclusive<usize> = 1..=4;
+
+const MACRO_SAMPLES: Option<usize> = Some(10_000_000);
+
+// Step by 50k records from 50k to 200k records
+fn record_counts() -> impl Iterator<Item = usize> {
+    (50_000..=200_000).step_by(50_000)
+}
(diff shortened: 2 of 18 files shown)


rust/ddlog_benches/README.md
```diff
@@ -0,0 +1,12 @@
+# ddlog_benches
+
+DDlog has a unique benchmarking scenario since it's a compiler, so benchmarking the programs
+it generates is a little more involved than normal, which is why we use [`cargo-make`] to automate
+the setup, build and bench process in a cross-platform way. To run the benchmarks in this crate,
+first install `cargo-make` with the following command (`--force` makes cargo install the latest version)
```
ryzhyk, line 6: How do you run the benchmark after installing `cargo-make`?

ryzhyk:
This sounds awesome!  I don't see the `.dl` file though.

Also, do your perf results include the recent improvements you've been working on or is this on master?

rust/ddlog_benches/README.md
Kixiron: `cargo make benchmark-all` at the moment, I'll probably add settings later so you can choose which benchmarks and so on, but there's no point in that now. I'll add the command to the readme, thanks for reminding me!

Kixiron:
The `.dl` file was somehow not included which is annoying. These results are on master, I didn't want to do a combined pr since that'd be too much, plus this gives a nice baseline and a good test bench that's decoupled from anything to do with rslint so it's focused entirely on ddlog

rust/ddlog_benches/benches/citations.rs
```diff
@@ -0,0 +1,82 @@
+use criterion::{
+    black_box, criterion_group, criterion_main, BatchSize, BenchmarkId, Criterion, Throughput,
+};
+use ddlog_benches::citations;
+use std::{ops::RangeInclusive, time::Duration};
+
+/// Benchmark all targets using 1, 2, 3, and 4 threads
+const DDLOG_WORKERS: RangeInclusive<usize> = 1..=4;
+
+// Increment the number of records taken by 100,000 until we're taking 1 million records
```
ryzhyk, line 10: Looks like there's just 1 increment happening here.  Does it make sense to go 100,000 in increments of 20,000? Actually, I'm not sure I see a lot of value in benchmarking loads of different dataset sizes. 2 or 3 is probably enough.

rust/ddlog_benches/src/citations.rs
```diff
@@ -0,0 +1,71 @@
...
+
+/// Creates a new ddlog instance with the specified number of workers
+pub fn init(workers: usize) -> DDlog {
+    let (ddlog, _) = DDlog::run(workers, false).expect("failed to create DDlog instance");
+    ddlog
+}
+
+/// Runs the test case, returning the ddlog instance so that the dropping of the ddlog
+/// instance isn't recorded in the benchmark
+pub fn run(ddlog: DDlog, dataset: Vec<Update<DDValue>>) -> DDlog {
```
ryzhyk, line 59: Would be nice to extend the benchmark to compare incremental vs non-incremental evaluation.  This currently does everything in one transaction.

ryzhyk (commented):
As always, this is very nice and useful, thanks!  In addition to the inline comments, I was wondering if it'd be possible to add correctness tests to the benchmarks or should we run them as separate tests?  Once this PR is ready, I will integrate it into our test suite, so we can run benchmarks on each commit.

rust/ddlog_benches/README.md
```diff
@@ -0,0 +1,25 @@
...
+cargo install --force cargo-make
+```
+
+To run all benchmarks use the following command
+
+```sh
+cargo make benchmark-all
+```
+
+When adding a benchmark, add the required ddlog code to `/ddlog`, any datasets to `/data` (preferably in a
```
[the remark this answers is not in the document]
Kixiron: What I was actually considering was just downloading the sources if they weren't already there (and providing a command to re-download them as well) since they're available as part of the snap dataset

rust/ddlog_benches/src/citations.rs
Kixiron: Yes, I'm looking at using one of the temporal datasets for exactly that

rust/ddlog_benches/benches/citations.rs
Kixiron: So, this is my current conundrum: Criterion is amazing and gives statistically correct results, but even at the minimum sample size of 10 benchmarks still take upwards of 5 minutes *each*. The alternative would be some sort of "macro-benchmarking" rig that just runs each case 1 or 2 times and averages them, but we lose the sense of security that criterion provides. An alternative to both of these would be a hybrid approach, running smaller datasets under criterion while larger go under the macro rig, thoughts?

rust/ddlog_benches/README.md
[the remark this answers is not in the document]
ryzhyk: Sounds good, there's no reason to create a new repo if we can use an external source (as long as the external source does not change and is likely to stay there for a long time). Do you plan to integrate the download step in Makefile.toml?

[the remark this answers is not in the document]
Kixiron: Yes, the [snap dataset](http://snap.stanford.edu/data/index.html) shouldn't change and the downloading will be in the makefile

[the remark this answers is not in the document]
Kixiron: Considering the task a wee bit more, I may opt for an [xtask](https://github.com/matklad/cargo-xtask/) rig to minimize code duplication, as much as I like cargo-make it's not as modular as I'd like

rust/ddlog_benches/benches/citations.rs
ryzhyk: It sounds like we want to stick to criterion.  Can the test suite be controlled from the command line? E.g., can we set things up so that on every commit we run a scaled down version (say, 5 iterations of shorter 2-minute tests) in CI, and then we can run longer, more reliable, benchmarks on demand, e.g., on a dedicated server or just locally on our machines?.

Kixiron: I'll look into it, I think I should be able to make things controlled via cli so that we can scale things along to where they're needed (be it large & long benchmarks locally or small ones on CI)

rust/ddlog_benches/README.md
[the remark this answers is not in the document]
ryzhyk: I'm not familiar with either of those, but xtask does not seem to have many users or be particularly well maintained.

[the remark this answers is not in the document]
Kixiron: Ah, it's a pattern as opposed to an actual crate, it's just a pattern of using rust as a medium for setup/teardown scripts. I don't actually think it's needed though

rust/ddlog_benches/benches/twitter.rs
```diff
@@ -14,8 +14,8 @@ fn record_counts() -> impl Iterator<Item = usize> {
     (50_000..=200_000).step_by(50_000)
 }
 
-fn twitter(c: &mut Criterion) {
-    let mut group = c.benchmark_group("twitter");
+fn twitter_micro(c: &mut Criterion) {
```
Kixiron, line 17: You probably want to do this same thing to the live-journal benchmark

ryzhyk: I did.

PS. I rebased this branch locally and am still tweaking it.  Will force-push later today.

ryzhyk:
@Kixiron , I squashed your commits to make sure `.csv` files are not stuck in the history.  It did not look like individual commit boundaries mattered in this case.

I love this benchmarking framework.  We should definitely grow it in the future and probably get some dedicated hardware to run it on.

Kixiron:
Yah, I'm definitely happy to look into expanding the benches in the future with more complex benches for testing incrementality and concurrency

PR #905, opened 2021-01-27
Add OS X warning fix to README

debnil:
I added instructions to work around the "unverified developer" security warning when using the precompiled binary on OS X. If not fixed, the user will encounter the warning when first using `ddlog`, likely in the tutorial.
The desktop popup will have the following text:
```
"ddlog" cannot be opened because the developer cannot be verified.
macOS cannot verify that this app is free from malware.

Chrome downloaded this file ... from `github.com`
```

An image of the popup is below for additional reference. 
OS X unverified developer security warning


mbudiu-vmw (changes requested):
Can you also give an example of the error you would get? This will make it easier for people to pattern-match or search for it.

mbudiu-vmw:
I meant for the changes to be in the README, not in the PR message.

mbudiu-vmw (approved):

PR #907, opened 2021-01-27
Script to install DDlog vscode extension

mbudiu-vmw:
The DDlog syntax highlighting is not perfect, but it is still useful.


Kixiron:
I don't think the script should install the theme (and I'm not sure why we have it in the first place), a language doesn't offer a theme for itself, it just gives vscode the tools to apply the user's own theme to the language

ryzhyk:
@mbudiu-vmw , shall we merge this?

PR #910, opened 2021-02-02
Support insert_or_update and delete_by_key using flatbufs 

mbudiu-vmw:
Expose them in the Java API.

I tried to have an enum in flatbuf.fbs with operation names, but I am not sure how to import it from the rust code in flatbuf.rs, since the namespace is only known at compile-time. So for now I hardwired the constant values instead of using an enum.

### DIFF
--- a/rust/template/src/flatbuf.rs
+++ b/rust/template/src/flatbuf.rs
@@ -5,6 +5,8 @@ use differential_datalog::program;
 use differential_datalog::program::{Response, Update};
 use differential_datalog::DeltaMap;
 use flatbuffers as fbrt;
+use enum_primitive_derive::Primitive;
+use num_traits::{FromPrimitive, ToPrimitive};
 
 /// Trait for types that can be de-serialized from FlatBuffer-embedded objects.
 pub trait FromFlatBuffer<T>: Sized {
@@ -464,8 +466,8 @@ impl<'a> FromFlatBuffer<fb::__Command<'a>> for DDValueUpdate {
         let val_table = cmd.val().ok_or_else(|| {
             format!("Update::from_flatbuf: invalid buffer: failed to extract value")
         })?;
-        match cmd.operation() {
-            Operation_InsertOrDelete => {
+        match FBOperation::from_i8(cmd.operation()) {
+            Some(FBOperation::InsertOrDelete) => {
                 let val = relval_from_flatbuf(relid, val_table)?;
                 match cmd.weight() {
                     1 => Ok(DDValueUpdate(Update::Insert { relid, v: val })),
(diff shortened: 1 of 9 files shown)


src/Language/DifferentialDatalog/FlatBuffer.hs
```diff
@@ -1484,6 +1510,12 @@ rustValueFromFlatbuf =
     "        _ => Err(format!(\"DDValue::idxkey_from_flatbuf: invalid idxid {}\", idxid))"      $$
     "    }"                                                                                     $$
     "}"                                                                                         $$
+    "fn relkey_from_flatbuf(relid: program::RelId, v: fbrt::Table) -> Response<DDValue> {"      $$
+    "    match relid {"                                                                         $$
+    (nest' $ nest' $ vcat relkey_enums)                                                         $$
+    "        _ => Err(format!(\"DDValue::rekkey_from_flatbuf: invalid relid {}\", relid))"      $$
```
ryzhyk, line 1516: typo

rust/template/src/flatbuf.rs
```diff
@@ -464,15 +464,38 @@ impl<'a> FromFlatBuffer<fb::__Command<'a>> for DDValueUpdate {
...
+                o
+            )),
         }
     }
 }
 
+// These should be ideally imported from flatbuf, but
+// I don't know how to do this.  In the meantime keep in sync
+// with Flatbuffer.hs
+const Operation_InsertOrDelete: i8 = 0;
```
ryzhyk, line 495: I think this should be an enum
```
enum FBOperation {
    InsertOrDelete = 0,
    InsertOrUpdate = 1,
    DeleteByKey = 2,
}
```

You can cast it like so: `FBOperation::InsertOrUpdate as i8`.

ryzhyk (commented):
I think you also need to modify `relTypesToSerialize` in `FlatBuffer.hs` to include types used in the primary key of the relation if any.  I did not try compiling this:

```
relTypesToSerialize :: (?d::DatalogProgram) => Relation -> [Type]
relTypesToSerialize rel@Relation{..} =
nub $ rel_types ++ key_types
where
    rel_types = execState (typeSubtypes [] relType)
              [typeNormalize ?d relType] -- Relation type must appear in 'union Value'; therefore we
                                         -- generate a table for it even if it's a primitive
                                         -- type.
    key_types = maybe [] (\key_type -> execState (typeSubtypes [] key_type)
              [typeNormalize ?d key_type]) $ relKeyType ?d rel
```

This requires adding a new function `relKeyType` to `Relation.hs`:

```
relKeyType :: DatalogProgram -> Relation -> Maybe Type
relKeyType d rel =
    case relPrimaryKey d rel of
            Nothing -> Nothing
            Just KeyExpr{..} -> Just $ exprType d (CtxKey rel) keyExpr
```

To test this, we need to add a relation with a primary key whose type doesn't occur in any other relation, e.g., a new struct only used in that key.  Without the above fix the test should fail. (In fact, I think that tests are currently failing in CI because of this).

rust/template/src/flatbuf.rs
```diff
@@ -464,26 +466,29 @@ impl<'a> FromFlatBuffer<fb::__Command<'a>> for DDValueUpdate {
...
                 let val = relval_from_flatbuf(relid, val_table)?;
                 Ok(DDValueUpdate(Update::InsertOrUpdate { relid, v: val }))
             }
-            Operation_DeleteByKey => {
+            Some(FBOperation::DeleteByKey) => {
                 let val = relkey_from_flatbuf(relid, val_table)?;
                 Ok(DDValueUpdate(Update::DeleteKey { relid, k: val }))
             }
-            o => Err(format!(
+            Some(o) => Err(format!(
```
ryzhyk, line 486: You cover all cases, so this should be unreachable. We should be able to remove this case in the next PR.

PR #913, opened 2021-02-06
Use GCC-like error message format that Emacs can parse.

blp:
This isn't right, because I couldn't figure out how to do the equivalent of "fprintf(stderr, ...); exit(1);" in Haskell, but it's a start.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -29,6 +29,7 @@ import System.FilePath.Posix
 import System.Console.GetOpt
 import System.Console.ANSI
 import Control.Exception
+import Control.Monad.Trans.Except
 import Control.Monad
 import Data.List
 import Data.Time.Clock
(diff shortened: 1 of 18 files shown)


mbudiu-vmw (approved):
I think you will need to rerun some tests and regenerate their expected outputs. Look for the tests with "fail" in test.sh

blp:
> I think you will need to rerun some tests and regenerate their expected outputs. Look for the tests with "fail" in test.sh

I'm sure that's true, but it's it's not ready yet; I don't know how to do it right.

ryzhyk:
I got rid of the `ddlog:` prefix. 

@blp, can you confirm that emacs like the error messages now?

blp:
> I got rid of the `ddlog:` prefix.
> 
> @blp, can you confirm that emacs like the error messages now?

Yes, it does.

I still see the ddlog prefix and it's the error prefix that seems to have gone away, but Emacs is happy:

```
ddlog: /home/bpfaff/nicira/ovn/northd/lswitch.dl:49.1-49.18: Unknown relation 'Warning'
```

ryzhyk:
> > I got rid of the `ddlog:` prefix.
> > @blp, can you confirm that emacs like the error messages now?
> 
> Yes, it does.
> 
> I still see the ddlog prefix and it's the error prefix that seems to have gone away, but Emacs is happy:
> 
> ```
> ddlog: /home/bpfaff/nicira/ovn/northd/lswitch.dl:49.1-49.18: Unknown relation 'Warning'
> ```

Right, there are several classes of errors. I only fixed one of them. Let me work on the other two.

Issue #784, opened 2020-10-22
Allow pattern matches in `for`, `FlatMap`, closures


ryzhyk:
We currently don't allow any of the following:

```
for ((k,v) in map)
```

```
(var x, var y) = FlatMap(expr)
```

```
|(x,y)| x+y
```

mbudiu-vmw:
Also `if ((k,v) = z)`, which returns true only if the pattern matching succeeds.

Kixiron:
I would prefer a rust-like `if let (k, v) = z {}` so it's clear that pattern matching is happening

ryzhyk:
I am with @Kixiron on this one. `if ((k,v) = z)` is very unorthodox.  I haven't seen this in any other language.

`if let (k, v) = z {}` on the other hand is just syntactic sugar around `match`.

PR #914, opened 2021-02-11
Improved handling of iteration in the compiler.

ryzhyk:
Introduce the `#[dyn_alloc]` attribute (see commit message for detail).

 This PR also improves the handling of iteration in a couple of ways.
    
- First, it enables pattern matching in `for` loops and `FlatMap` (#784).  One
  can now write:
        ```
        for ((k,v) in map) {}
        ```
        instead of
        ```
        for ((kv in map) { var k = kv.0}
        ```
        and
        ```
        (var x, var y) = FlatMap(expr)
        ```
        instead of:
        ```
        var xy = FlatMap(expr),
        var x = xy.0
        ```
      In general, the pattern can contain constructor names, tuples,
      placeholders, and variable declarations.  The pattern must be
      irrefutable, i.e., any constructor name used in the pattern must be
      the unique constructor for the given type.

- Second, we remove the hardwired knowledge about iterable types from
      the compiler.  Until now the compiler only knew how to iterate over
      `Vec`, `Set`, `Map`, `Group`, and `TinySet` types.  Instead we now
      allow the programmer to label any extern type as iterable, meaning
      that it implements `iter()` and `into_iter()` methods, that return
      Rust iterators using one of two attributes:
      ```
      #[iterate_by_ref=iter:<type>]
      ```
      or
      ```
      #[iterate_by_val=iter:<type>]
      ```
      where `type` is the type yielded by the `next()` method of the iterator
      The former indicates that the `iter()` method returns a by-reference
      iterator, the latter indicates that `iter()` returns a by-value
      iterator.
    
    Example:

     ```
     #[iterate_by_val=iter:('K,'V)]
     extern type Map<'K,'V>
     ```
    
     This feature will enable us to introduce new container types, e.g.,
     immutable sets and maps in the future.
    
     Unfortunately, it introduces one, hopefully minor, usability
     regression.  Since we can no longer distinguish a map from any other
     container whose iterator is a 2-tuple, we cannot reliably convert
     DDlog maps to/from Java maps in the FlatBuf-based API.  They must
     therefore be passed around as arrays of key-value tuples and converted
     to/from Java maps by the client.
    
Future todos:
- Support pattern matching in closure arguments and the LHS of `group_by` clauses.
- Should refutable patterns be supported for `FlatMap`?  This would have
  the effect of filtering the flattened collection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,50 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Language improvements
+
+- Enable pattern matching in `for` loops and `FlatMap`. One can now write:
+    ```
+    for ((k,v) in map) {}
+    ```
+    instead of
+    ```
+    for ((kv in map) { var k = kv.0}
+    ```
+    and
+    ```
+    (var x, var y) = FlatMap(expr)
+    ```
+    instead of:
+    ```
+    var xy = FlatMap(expr),
+    var x = xy.0
+    ```
+
+- Remove the hardwired knowledge about iterable types from the compiler.
+  Until now the compiler only knew how to iterate over `Vec`, `Set`,
+  `Map`, `Group`, and `TinySet` types.  Instead we now allow the programmer
+  to label any extern type as iterable, meaning that it implements `iter()`
+  and `into_iter()` methods, that return Rust iterators using one of two
+  attributes:
+  ```
+  #[iterate_by_ref=iter:<type>]
+  ```
+  or
+  ```
+  #[iterate_by_val=iter:<type>]
+  ```
+  where `type` is the type yielded by the `next()` method of the iterator
+  The former indicates that the `iter()` method returns a by-reference
+  iterator, the latter indicates that `iter()` returns a by-value
+  iterator.
+
+  As a side effect of this change, the compiler no longer distinguishes
+  maps from other cotnainers that over 2-tuples.  Therefore maps are
+  represented as lists of tuples in the Flatbuf-based Java API.
+
 ## [0.36.0] - Feb 7, 2021
 
 ### API changes
--- a/doc/language_reference/language_reference.md
+++ b/doc/language_reference/language_reference.md
@@ -497,7 +497,7 @@ cons_term    ::= (* positional arguments *)
                             ("," "." field_name "=" expr)* "}"]
 var_term     ::= var_name
 ite_term     ::= "if" term term [ "else" term ]
-for_term     ::= "for" "(" var_name "in" expr ")" term
+for_term     ::= "for" "(" for_pattern "in" expr ")" term
 return_term  ::= "return" [expr]
 vardecl_term ::= "var" var_name
 
@@ -524,6 +524,20 @@ pattern ::= (* tuple pattern *)
           | string_literal  (* matches specified string value *)
           | int_literal     (* matches specified integer or bitvector value *)
           | "_"             (* wildcard, matches any value *)
+```
+
+```EBNF
+(* pattern that binds loop variables in a for-loop *)
+for_pattern ::= (* tuple pattern *)
+                "(" [pattern (,pattern)* ")"
+                (* constructor pattern with positional arguments. The constructor must be the unique constructor for the type.*)
+              | cons_name ["{" [pattern (,pattern)*] "}"]
+                (* constructor pattern with named arguments. The constructor must be the unique constructor for the type. *)
+              | cons_name "{" ["." field_name "=" pattern
+                               ("," "." field_name "=" pattern)*] "}"
+              | vardecl_term    (* binds variable to a field inside the matched value or the entire value. *)
+              | var_term        (* binds variable to a field inside the matched value (shorthand for vardecl_term) *)
+              | "_"             (* wildcard, matches any value *)
 ```
 
 ### Constraints on expressions
@@ -647,9 +661,20 @@ rhs_clause ::= atom                                      (* 1.atom *)
              | "not" atom                                (* 2.negated atom *)
              | expr                                      (* 3.condition *)
              | expr "=" expr                             (* 4.assignment *)
-             | "var" var_name "=" "FlatMap" "(" expr ")" (* 5.flat map *)
+             | flatmap_pattern "=" "FlatMap" "(" expr ")" (* 5.flat map *)
              | "var" var_name = expr "." "group_by"      (* 6.grouping; in general a *)
                                 "(" expr ")"             (*   group_by clause can be any expression containing `expr.group_by(expr)` subexpression. *)
+
+(* pattern that binds variables in the left-hand side of a FlatMap clause *)
+flatmap_pattern ::= (* tuple pattern *)
+                    "(" [pattern (,pattern)* ")"
+                    (* constructor pattern with positional arguments.  The constructor must be the unique constructor for a type. *)
+                  | cons_name ["{" [pattern (,pattern)*] "}"]
+                    (* constructor pattern with named arguments. The constructor must be the unique constructor for the type. *)
+                  | cons_name "{" ["." field_name "=" pattern
+                                  ("," "." field_name "=" pattern)*] "}"
+                  | vardecl_term    (* binds variable to a field inside the matched value or the entire value. *)
+                  | "_"             (* wildcard, matches any value *)
 ```
 
 An atom is a predicate that holds when a given value belongs to a relation.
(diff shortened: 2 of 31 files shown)


java/test_flatbuf1/Test.java
```diff
@@ -57,11 +57,11 @@ private String printNI(NIReader v) {
     }
 
     private String printQI(QIReader v) {
-        Map<Long,String> vs = v.m();
+        List<Tuple2__bit_32___stringReader> vs = v.m();
```
mbudiu-vmw, line 60: this is an API change.
I assume lists are more efficient, but this is somewhat unintuitive. You'd think that a map in DL maps to a map in Java.
Also, the usability of the Map is better, since you don't have to figure out this Tuple name.

src/Language/DifferentialDatalog/Attribute.hs
```diff
@@ -84,11 +88,20 @@ typedefValidateAttr d TypeDef{..} attr = do
                 $ "'sharef_ref' attribute is only applicable to extern types."
             check d (length tdefArgs == 1) (pos attr)
                 $ "Types annotated with 'shared_ref' must have exactly one type argument, e.g., \"Ref<'T>\"."
+         "dyn_alloc" -> do
+            check d (isNothing tdefType) (pos attr)
```
mbudiu-vmw, line 92: Maybe you should have a helper function isExternType. Would make this more readable and abstract.

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -2782,6 +2740,20 @@ mkPatExpr' d inkind ctx e               varkind _    = do
                    (deref (vname, varkind, undefined) <+> "==" <+> mkExpr d ctx (E e) EVal)
                    []
 
+-- Irrefutable patterns are patterns that are quaranteed to match any expression
```
mbudiu-vmw, line 2743: typo "guaranteed"

java/test_flatbuf1/Test.java
mbudiu-vmw: I guess you are doing this because you represent collections by iterating over their elements.

src/Language/DifferentialDatalog/FlatBuffer.hs
```diff
@@ -385,22 +386,17 @@ typeTableName x =
...
+                       -> "__Table_Vec_" <> (mkTypeIdentifier $ fst $ typeIterType ?d t)
          t             -> error $ "typeTableName: Unexpected type " ++ show t
 
 -- True if type is serialized into a vector inside FlatBuffer.
 typeIsVector :: (WithType a, ?d::DatalogProgram) => a -> Bool
-typeIsVector x =
-    case typeNormalizeForFlatBuf x of
-         TOpaque{..} -> elem typeName sET_TYPES || typeName == mAP_TYPE
-         _ -> False
+typeIsVector x = typeIsIterable ?d $ typeNormalizeForFlatBuf x
```
mbudiu-vmw, line 399: this function name is a bit misleading. Perhaps it should be typeIsSerializedAsVector

src/Language/DifferentialDatalog/FlatBuffer.hs
```diff
@@ -824,8 +810,8 @@ jConvObjType in_array_context rw x =
          _                  -> jConvType rw x
 
 
-jConvObjTypeW :: (WithType a, ?d::DatalogProgram) => a -> Doc
-jConvObjTypeW = jConvObjType False Write
+--jConvObjTypeW :: (WithType a, ?d::DatalogProgram) => a -> Doc
```
mbudiu-vmw, line 813: why not delete this?

src/Language/DifferentialDatalog/FlatBuffer.hs
```diff
@@ -892,6 +876,7 @@ jConv2FBType fbctx e t =
 
 -- Like jConv2FBType, but assumes that input is a Java object type, not a
 -- primitive.
+{-
```
mbudiu-vmw, line 879: and this?

CHANGELOG.md
```diff
@@ -3,6 +3,46 @@ All notable changes to this project will be documented in this file.
...
+  #[iterate_by_ref=iter:<type>]
+  ```
+  or
+  ```
+  #[iterate_by_val=iter:<type>]
+  ```
+  where `type` is the type yielded by the `next()` method of the iterator
+  The former indicates that the `iter()` method returns a by-reference
+  iterator, the latter indicates that `iter()` returns a by-value
+  iterator.
```
mbudiu-vmw, line 44: i think you have to mention that you change the API for Java serialization of Maps. It is a breaking change for users.

src/Language/DifferentialDatalog/Rust.hs
```diff
@@ -0,0 +1,82 @@
...
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+-}
+
+{-# LANGUAGE RecordWildCards, FlexibleContexts, LambdaCase, TupleSections, OverloadedStrings, TemplateHaskell, QuasiQuotes, ImplicitParams, NamedFieldPuns #-}
+
+{- |
+Module     : Rust
+Description: Rust language definitions used in the DDlog compiler.
```
mbudiu-vmw, line 28: I imagine this file will grow.

src/Language/DifferentialDatalog/TypeInference.hs
```diff
@@ -374,11 +374,21 @@ deIsSharedRef de tv = do
...
+                         $ "expression '" ++ show de ++ "' must be of an iterable type, e.g., 'Set<>', 'Map<>', 'Vec<>', or 'Group<>', " ++
+                           "but its type '" ++ n ++ "' is not iterable"
+                 Just (titer, _) ->
+                     do let tmap = M.fromList $ zip (tdefArgs t) targs
+                        let te_titer = typeToTExpr titer
+                        let titer_concrete = teSubstTypeArgs tmap te_titer
+                        return $ Just [tv ==== titer_concrete]
         expand te = err ?d (pos de)
-                    $ "expression '" ++ show de ++ "' must be of an iterable type, e.g., 'Set<>', 'Map<>', 'Vec<>', or 'Group<>', but its type is " ++ show te
+                    $ "expression '" ++ show de ++ "' must be of an iterable type, e.g., 'Set<>', 'Map<>', 'Vec<>', or 'Group<>', " ++
```
mbudiu-vmw, line 390: or annotated with....

test/datalog_tests/simple2.dl
```diff
@@ -471,3 +471,129 @@ FlatGroup(x) :-
...
+    f1: 'T,
+    f2: istring
+}
+
+typedef ForStruct2 = ForStruct2 {
+    f1: (bool, double),
+    f2: ForStruct1<(usize, usize)>
+}
+
+function testMap(): Map<(usize, string), ForStruct2> {
```
mbudiu-vmw, line 489: what is the difficulty in creating a `const` construct in DDlog that is mapped to a global constant in Rust?

mbudiu-vmw (approved):

doc/language_reference/language_reference.md
```diff
@@ -647,9 +661,20 @@ rhs_clause ::= atom                                      (* 1.atom *)
              | "not" atom                                (* 2.negated atom *)
              | expr                                      (* 3.condition *)
              | expr "=" expr                             (* 4.assignment *)
-             | "var" var_name "=" "FlatMap" "(" expr ")" (* 5.flat map *)
+             | flatmap_pattern "=" "FlatMap" "(" expr ")" (* 5.flat map *)
              | "var" var_name = expr "." "group_by"      (* 6.grouping; in general a *)
                                 "(" expr ")"             (*   group_by clause can be any expression containing `expr.group_by(expr)` subexpression. *)
+
+(* pattern that binds variables in the left-hand side of a FlatMap clause *)
```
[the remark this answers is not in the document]
ryzhyk: No fundamental reason, but in for-loops the `var` prefix can be omitted: `for (x in y)` is equivalent to `for (var x in y)`, while FlatMap's always require it: `var x = FlatMap(y)`.

java/test_flatbuf1/Test.java
ryzhyk: The compiler simply does not know that something behaves like a map any longer. We can add another attribute for this or just live with this and let the client convert between maps and arrays of tuples themselves.

test/datalog_tests/simple2.dl
ryzhyk: Shouldn't be too hard.  One caveat is that the parse won't be able to disambiguate between constructor names and constants in some cases, so we'll have to resolve this at a later stage, but we already have similar logic for variable name vs function name resolution, so this is not a problem.

PR #912, opened 2021-02-04
Stream aggregation.

ryzhyk:
Implements RFC #904.

TODOs:
- [x] Explicit syntax for differentiation (currently applied automatically when the body of the rule is a stream, but the head is relation)
- [x] Program rewriting pass to eliminate arrangements of delayed relations.
- [x] More tests
- [x] Documentation
- [x] Commit message
- [x] Merge conflicts
- [x] Changelog

PR #918, opened 2021-02-13
Iterate over groups with weights.

ryzhyk:
See individual commits for details.

Note: the plan is to rebase #912 on top of this PR, so we can test weighted aggregation before merging both PRs.

### DIFF
--- a/doc/tutorial/tutorial.md
+++ b/doc/tutorial/tutorial.md
@@ -1482,7 +1482,7 @@ function best_vendor_string(g: Group<string, (string, bit<64>)>): string
 {
     var min_vendor = "";
     var min_price = 'hffffffffffffffff;
-    for ((vendor, price) in g) {
+    for (((vendor, price), w) in g) {
         if (price < min_price) {
             min_vendor = vendor;
             min_price = price;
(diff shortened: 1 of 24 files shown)


doc/tutorial/tutorial.md
```diff
@@ -1497,6 +1497,26 @@ BestDeal(best) :-
...
 
+Similar to vectors, sets, and other container types, groups are iterable
+objects, and can be iterated over in a for-loop of flattened by a `FlatMap`
+operator.  Let us have a closer look at the for-loop in the last code snippet:
+
+```
+for (((vendor, price), w) in g) { .. }
+```
+
+The iterator yields a 2-tuple, where the first component `(vendor, price)` of
```
mbudiu-vmw, line 1508: This is tied to projection. You can implement projection in multiple ways: return a tuple multiple times with different weights, or return it once with the sum of the weights. 

For relations the internal weights should really be invisible to users, they are an artifact of the differential implementation (logically the set should be distincted before being passed to group-by, and then the weights would all 1). For multisets the weights will be important.

mbudiu-vmw: In fact, I believe that for a relation you should apply "distinct" to the weight before you pass it to the group function, i.e., `w -> if (w>0) then 1 else 0`. For a stream or multiset you should pass the real weight.

mbudiu-vmw (approved):
I will approve, but I really think that the weight should be 1 for relations, not the internal DDlog weight.

ryzhyk:
> I will approve, but I really think that the weight should be 1 for relations, not the internal DDlog weight.

It's doable, but we're not quite ready for this change yet.  We don't enforce the distinction between multisets and relations for internal relations, they are basically all multisets.

Also, what if the projection of a relation is a multiset.  Would you still distinct weights?  This is probably not what users want or expect in most cases.

I am going to close this PR now and continue developing it in #912 .

PR #921, opened 2021-02-23
Bug in simplifyDifferentiation.

ryzhyk:
When generating auxiliary relations for differentiation and delay
operators, we incorrectly added the prefix to the fully qualified
relation name instead of the local component of the name.  As a result,
the compiler crashed when such relations were declared anywhere other
than the main module.


src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -624,7 +624,11 @@ simplifyDelayedRels d =
            delayed_rels d'
     where
     mk_delayed_rel_name :: (String, Delay) -> String
-    mk_delayed_rel_name (rel, del) = "__" ++ show (delayDelay del) ++ "_" ++ rel
+    mk_delayed_rel_name (rel, del) = scoped rel_scope $ "__" ++ show (delayDelay del) ++ "_" ++ rel_name
```
mbudiu-vmw, line 627: Can you please add a documentation line in the reference manual that some identifiers are reserved?

mbudiu-vmw (approved):

PR #922, opened 2021-02-27
Functional hashsets

ryzhyk:
See individual commit messages.

@Kixiron, any chance you could have a look at the Rust code?

### DIFF
--- a/lib/hashset.dl
+++ b/lib/hashset.dl
@@ -1,0 +1,200 @@
+/* Immutable hash sets.
+ * This module contains bindings for the `HashSet` type
+ * from the `im` crate. */
+
+#[iterate_by_ref=iter:'A]
+extern type HashSet<'A>
+
+extern function hashset_singleton(x: 'X): HashSet<'X>
+extern function hashset_empty(): HashSet<'X>
+
+function size(s: HashSet<'X>): usize {
+    hashset_size(s)
+}
+
+function insert(s: mut HashSet<'X>, v: 'X) {
+    hashset_insert(s, v)
+}
+
+function insert_imm(s: HashSet<'X>, v: 'X): HashSet<'X> {
+    hashset_insert_imm(s, v)
+}
+
+function remove(s: mut HashSet<'X>, v: 'X) {
+    hashset_remove(s, v)
+}
+
+function remove_imm(s: HashSet<'X>, v: 'X): HashSet<'X> {
+    hashset_remove_imm(s, v)
+}
+
+function contains(s: HashSet<'X>, v: 'X): bool {
+    hashset_contains(s, v)
+}
+
+function is_empty(s: HashSet<'X>): bool {
+    hashset_is_empty(s)
+}
+
+function nth(s: HashSet<'X>, n: usize): Option<'X> {
+    hashset_nth(s, n)
+}
+
+function to_vec(s: HashSet<'A>): Vec<'A> {
+    hashset_to_vec(s)
+}
+
+function to_hashset(v: Vec<'A>): HashSet<'A> {
+    var res = hashset_empty();
+    for (x in v) {
+        res.insert(x);
+    };
+    res
+}
+
+function to_hashset(g: Group<'K, 'A>): HashSet<'A> {
+    var res = hashset_empty();
+    for ((x, _) in g) {
+        res.insert(x);
+    };
+    res
+}
+
+function to_hashset(o: Option<'X>): HashSet<'X> {
+    match (o) {
+        Some{x} -> hashset_singleton(x),
+        None -> hashset_empty()
+    }
+}
+
+function union(s1: HashSet<'X>, s2: HashSet<'X>): HashSet<'X> {
+    hashset_union(s1, s2)
+}
+
+function union(sets: Vec<HashSet<'X>>): HashSet<'X> {
+    hashset_unions(sets)
+}
+
+function union(sets: Group<'K, HashSet<'X>>): HashSet<'X> {
+    group_hashset_unions(sets)
+}
+
+function intersection(s1: HashSet<'X>, s2: HashSet<'X>): HashSet<'X> {
+    hashset_intersection(s1, s2)
+}
+
+function difference(s1: HashSet<'X>, s2: HashSet<'X>): HashSet<'X> {
+    hashset_difference(s1, s2)
+}
+
+/* Applies closure `f` to each element of the set. */
+function map(s: HashSet<'A>, f: function('A): 'B): HashSet<'B> {
+    var res = hashset_empty();
+    for (x in s) {
+        res.insert(f(x))
+    };
+    res
+}
+
+/* Returns the element that gives the minimum value from the specified function.
+ * If several elements are equally minimum, the first element is returned.
+ * If the set is empty, `None` is returned. */
+function arg_min(s: HashSet<'A>, f: function('A): 'B): Option<'A> {
+    hashset_arg_min(s, f)
+}
+
+/* Returns the element that gives the maximum value from the specified function.
+ * If several elements are equally maximum, the first element is returned.
+ * If the set is empty, `None` is returned. */
+function arg_max(s: HashSet<'A>, f: function('A): 'B): Option<'A> {
+    hashset_arg_max(s, f)
+}
+
+/* Returns the first element of the set that satisfies predicate `f` or
+ * `None` if none of the elements satisfy the predicate. */
+function find(s: HashSet<'A>, f: function('A): bool): Option<'A> {
+    for (x in s) {
+        if (f(x)) {
+            return Some{x}
+        }
+    };
+    None
+}
+
+/* Returns a vector containing only those elements in `s` that satisfy predicate
+ * `f`. */
+function filter(s: HashSet<'A>, f: function('A): bool): HashSet<'A> {
+    var res = hashset_empty();
+    for (x in s) {
+        if (f(x)) {
+            res.insert(x)
+        }
+    };
+    res
+}
+
+/* Both filters and maps the set.
+ *
+ * Calls the closure on each element of the set.  If the closure returns
+ * `Some{element}`, then that element is returned. */
+function filter_map(s: HashSet<'A>, f: function('A): Option<'B>): HashSet<'B> {
+    var res = hashset_empty();
+    for (x in s) {
+        match (f(x)) {
+            None -> (),
+            Some{y} -> res.insert(y)
+        }
+    };
+    res
+}
+
+/* Returnds `true` iff all elements of the set satisfy predicate `f`. */
+function all(s: HashSet<'A>, f: function('A): bool): bool {
+    for (x in s) {
+        if (not f(x)) {
+            return false
+        }
+    };
+    true
+}
+
+/* Returnds `true` iff at least one element of the set satisfies predicate `f`. */
+function any(s: HashSet<'A>, f: function('A): bool): bool {
+    for (x in s) {
+        if (f(x)) {
+            return true
+        }
+    };
+    false
+}
+
+/* Iterates over the set is ascending order, aggregating its contents using `f`.
+ *
+ * `f` - takes the previous value of the accumulator and the next element in the
+ * set and returns the new value of the accumulator.
+ *
+ * `initializer` - initial value of the accumulator. */
+function fold(s: HashSet<'A>, f: function('B, 'A): 'B, initializer: 'B): 'B {
+    var res = initializer;
+    for (x in s) {
+        res = f(res, x)
+    };
+    res
+}
+
+extern function hashset_arg_min(s: HashSet<'A>, f: function('A): 'B): Option<'A>
+extern function hashset_arg_max(s: HashSet<'A>, f: function('A): 'B): Option<'A>
+extern function hashset_size(s: HashSet<'X>): usize
+extern function hashset_insert(s: mut HashSet<'X>, v: 'X)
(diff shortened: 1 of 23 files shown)


lib/hashset.dl
```diff
@@ -0,0 +1,200 @@
...
+extern function hashset_remove_imm(s: HashSet<'X>, v: 'X): HashSet<'X>
+extern function hashset_contains(s: HashSet<'X>, v: 'X): bool
+extern function hashset_is_empty(s: HashSet<'X>): bool
+extern function hashset_nth(s: HashSet<'X>, n: usize): Option<'X>
+extern function hashset_to_vec(s: HashSet<'A>): Vec<'A>
+extern function hashset_union(s1: HashSet<'X>, s2: HashSet<'X>): HashSet<'X>
+extern function hashset_unions(sets: Vec<HashSet<'X>>): HashSet<'X>
+extern function group_hashset_unions(sets: Group<'K, HashSet<'X>>): HashSet<'X>
+extern function hashset_intersection(s1: HashSet<'X>, s2: HashSet<'X>): HashSet<'X>
+extern function hashset_difference(s1: HashSet<'X>, s2: HashSet<'X>): HashSet<'X>
```
mbudiu-vmw, line 200: No singleton?

lib/hashset.rs
```diff
@@ -0,0 +1,387 @@
...
+}
+
+pub fn hashset_to_vec<T: Hash + Clone + Ord>(s: &HashSet<T>) -> DDVec<T> {
+    DDVec {
+        vec: s.iter().cloned().collect(),
+    }
+}
+
+pub fn hashset_union<T: Hash + Eq + Clone>(s1: &HashSet<T>, s2: &HashSet<T>) -> HashSet<T> {
+    s1.clone().union(s2.clone())
```
mbudiu-vmw, line 354: Why clone?

rust/template/differential_datalog/src/program/worker.rs
```diff
@@ -428,10 +442,36 @@ impl<'a> DDlogWorker<'a> {
...
+            // Create an `Enabled` relation used to enforce the dataflow termination in the
+            // presence of delayed relations.  A delayed relation can potentially generate an
+            // infinite sequence of outputs for all future timestamps, preventing the
+            // differential dataflow from terminating.  DD can only terminate cleanly when there
+            // is no data in flight, and will keep spinning forever if new records keep getting
+            // injected in the dataflow.  To prevent this scenario, we join all delayed relations
+            // with `Enabled`.  `Enabled` contains a single record (an empty tuple) as long as
+            // the program is running.  We retract this record on shutdown, hopefully enforcing
+            // quiescing the dataflow.
+            let (enabled_session, enabled_collection) = outer.new_collection::<(),Weight>();
```
mbudiu-vmw, line 454: Is this overhead?

test/datalog_tests/hashset_test.dl
```diff
@@ -0,0 +1,82 @@
...
+    test_set().filter_map(|x| if (x.0 == 1) {
+        Some{ {var y = x; y.0 = 100; y} }
+    } else {
+        None
+    })).
+
+// arg_min.
+SetTransforms(
+    "arg_min(min)",
+    hashset_singleton(test_set().arg_min(|x| x.1.arg_min(|x|x)).unwrap_or_default())).
```
mbudiu-vmw, line 40: I see the singleton

test/datalog_tests/test-stream.sh
```diff
@@ -86,6 +88,29 @@ run_rel_query_test() {
...
+        for (( j=1; j<=$3; j++ ))
+        do
+            echo "insert UserSession$1($(($i * $2 + $j))),"
+            #echo "insert UserSession$1($j),"
+        done
+        echo "commit dump_changes;"
+    done
+    ) > unique_bench$1.dat
+    /usr/bin/time ./streams_ddlog/target/release/streams_cli -w 1 --no-store < unique_bench$1.dat > unique_bench$1.dump
+    diff -q unique_bench$1.dump unique_bench$1.dump.expected
```
mbudiu-vmw, line 108: How much faster is it?

mbudiu-vmw (approved):

lib/hashset.rs
```diff
@@ -0,0 +1,387 @@
...
+    }
+}
+
+impl<T: Hash + Eq + Clone> HashSet<T> {
+    pub fn unit(x: T) -> Self {
+        let mut set = IMHashSet::default();
+        set.insert(x);
+        HashSet::from(set)
+    }
+    pub fn insert(&mut self, x: T) -> Option<T> {
```
Kixiron, line 197:
```suggestion
    }

    pub fn insert(&mut self, x: T) -> Option<T> {
```

lib/hashset.rs
```diff
@@ -0,0 +1,387 @@
...
+        self.set.insert(x)
+    }
+    pub fn remove<BT: ?Sized>(&mut self, x: &BT) -> Option<T>
```
Kixiron, line 200:
```suggestion
    }

    pub fn remove<BT: ?Sized>(&mut self, x: &BT) -> Option<T>
```

lib/hashset.rs
```diff
@@ -0,0 +1,387 @@
...
+    where
+        BT: Hash + Eq,
+        T: Borrow<BT>,
+    {
+        self.set.remove(x)
+    }
+    pub fn without(&self, x: &T) -> Self {
```
Kixiron, line 207:
```suggestion
    }

    pub fn without(&self, x: &T) -> Self {
```

lib/hashset.rs
```diff
@@ -0,0 +1,387 @@
...
+        HashSet::from(self.set.without(x))
+    }
+    pub fn update(&self, x: T) -> Self {
```
Kixiron, line 210:
```suggestion
    }

    pub fn update(&self, x: T) -> Self {
```

lib/hashset.rs
```diff
@@ -0,0 +1,387 @@
...
+        HashSet::from(self.set.update(x))
+    }
+    pub fn union(self, other: Self) -> Self {
```
Kixiron, line 213:
```suggestion
    }

    pub fn union(self, other: Self) -> Self {
```

lib/hashset.rs
```diff
@@ -0,0 +1,387 @@
...
+}
+
+// Set update semantics: update contains values that are in one of the sets
+// but not the other.
+impl<T: FromRecord + Hash + Eq + Clone + Ord> Mutator<HashSet<T>> for Record {
+    fn mutate(&self, set: &mut HashSet<T>) -> StdResult<(), String> {
+        let upd = <HashSet<T>>::from_record(self)?;
+        for v in upd.into_iter() {
+            if set.remove(&v).is_none() {
+                set.insert(v);
```
Kixiron, line 294: I don't quite understand the rationale here, why is the element removed and inserted if there wasn't previously a value?

lib/hashset.rs
```diff
@@ -0,0 +1,387 @@
...
+    f: &Box<dyn Closure<*const T, B>>,
+) -> DDOption<T> {
+    DDOption::from(s.iter().min_by_key(|x| f.call(*x)).map(|x| (*x).clone()))
+}
+
+pub fn hashset_arg_max<T: Hash + Clone + Ord, B: Ord>(
+    s: &HashSet<T>,
+    f: &Box<dyn Closure<*const T, B>>,
+) -> DDOption<T> {
+    DDOption::from(s.iter().max_by_key(|x| f.call(*x)).map(|x| (*x).clone()))
```
Kixiron, line 386: By the looks of `Closure<*const T, _>` `Closure::call()` should be an unsafe function

rust/template/differential_datalog/src/program/worker.rs
```diff
@@ -40,15 +43,21 @@ use timely::{
     worker::Worker,
 };
 
-type SessionData = (
-    FnvHashMap<RelId, InputSession<TS, DDValue, Weight>>,
-    BTreeMap<
+// Handles to objects involved in managing the progress of the dataflow.
+struct SessionData {
```
Kixiron, line 47: ❤️

test/datalog_tests/hashset_test.dump.expected
```diff
@@ -0,0 +1,10 @@
+hashset_test::SetTransforms{.description = "all(contains(3))", .s = []}
+hashset_test::SetTransforms{.description = "any(contains(3))", .s = [(2, [4, 5, 6]), (4, [8, 9, 10]), (3, [6, 7, 8]), (1, [2, 3, 4]), (0, [0, 1, 2])]}
+hashset_test::SetTransforms{.description = "arg_max(max)", .s = [(4, [8, 9, 10])]}
+hashset_test::SetTransforms{.description = "arg_min(min)", .s = [(0, [0, 1, 2])]}
+hashset_test::SetTransforms{.description = "filter(contains 2)", .s = [(1, [2, 3, 4]), (0, [0, 1, 2])]}
+hashset_test::SetTransforms{.description = "filter_map(==1, :=100)", .s = [(100, [2, 3, 4])]}
+hashset_test::SetTransforms{.description = "find(contains(3))", .s = [(1, [2, 3, 4])]}
+hashset_test::SetTransforms{.description = "map(push 100)", .s = [(1, [2, 3, 4, 100]), (3, [6, 7, 8, 100]), (2, [4, 5, 6, 100]), (0, [0, 1, 2, 100]), (4, [8, 9, 10, 100])]}
+hashset_test::SetFolds{.description = "fold(+)", .a = 10}
+hashset_test::SetFolds{.description = "fold(fold(+))", .a = 75}
```
Kixiron, line 10: This makes me wonder if we shouldn't make `Collection::assert_eq()` a ddlog expression (similar to how `apply` works)

rust/template/differential_datalog/src/program/worker.rs
```diff
@@ -428,10 +442,34 @@ impl<'a> DDlogWorker<'a> {
...
+            // Create an `Enabled` relation used to enforce the dataflow termination in the
+            // presence of delayed relations.  A delayed relation can potentially generate an
+            // infinite sequence of outputs for all future timestamps, preventing the
+            // differential dataflow from terminating.  DD can only terminate cleanly when there
+            // is no data in flight, and will keep spinning forever if new records keep getting
+            // injected in the dataflow.  To prevent this scenario, we join all delayed relations
+            // with `Enabled`.  `Enabled` contains a single record (an empty tuple) as long as
+            // the program is running.  We retract this record on shutdown, hopefully enforcing
+            // quiescing the dataflow.
+            let (enabled_session, enabled_collection) = outer.new_collection::<(),Weight>();
```
Kixiron, line 454: I'd add a note pointing to [timely-dataflow/#306](https://github.com/TimelyDataflow/timely-dataflow/issues/306) as the (hopefully) ultimate solution to this

lib/hashset.dl
```diff
@@ -0,0 +1,200 @@
...
+        }
+    };
+    res
+}
+
+/* Both filters and maps the set.
+ *
+ * Calls the closure on each element of the set.  If the closure returns
+ * `Some{element}`, then that element is returned. */
+function filter_map(s: HashSet<'A>, f: function('A): Option<'B>): HashSet<'B> {
```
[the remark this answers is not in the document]
ryzhyk: Yes, but this is slightly more efficient, as creating a closure in DDlog involves a dynamic memory allocation.

lib/hashset.rs
ryzhyk: It's basically what the above comment says.  `upd` is a set that describes a set of updates to apply to `set`.  For each value in `upd`, if the value is in `set`, then it must be deleted, and if not, then it must be inserted.  Actually, now I realize that this is just the symmetric difference of `upd` and `set`, and we have a function for it, so that's what I should be using instead.

lib/hashset.rs
ryzhyk: The `union` function consumes both its arguments, so we cannot pass them by reference.

rust/template/differential_datalog/src/program/worker.rs
ryzhyk: yep, but a small one, as `Enabled` contains exactly one record, and we only pay it if there are delayed relations in the program.  For the one benchmark we have for this now, it does not seem to make any difference.

rust/template/differential_datalog/src/program/worker.rs
ryzhyk: Ack, although it doesn't look like it will be implemented any time soon...

test/datalog_tests/hashset_test.dump.expected
ryzhyk: Can you elaborate?  How would this work?  This would have to be part of the command language, not DDlog itself, right?

test/datalog_tests/test-stream.sh
ryzhyk: The version using hashsets is 20x faster than the equivalent implementation using regular sets.  The gap growth with the size of the set. I also had an optimized set-based implementation that was only 5 times slower.

test/datalog_tests/hashset_test.dump.expected
Kixiron: Basically I just see it taking advantage of the fact that vectors implement `.to_stream()`. You'd make a vec of the contents you expected to see and then we'd call ddflow's `assert_eq()` on the collection of expected items and the relation you want to check. Admittedly this only has fairly limited use since the user would have to specify timestamps, but I think it could still be genuinely useful. In a similar vein `.assert_empty()` could also be useful

PR #924, opened 2021-03-02
Formal definition of DDlog

mbudiu-vmw:
This is still incomplete, but has much more than it used to have.


mbudiu-vmw:
Not yet completed

PR #931, opened 2021-03-09
Added named regions for standalone relations

Kixiron:
This is part one of what I suspect will be a saga of similar changes. This pr adds a named region encompassing each standalone (non-recursive) relation. For further changes I'm reasonably sure I'll have to make much more sweeping changes to the dataflow rendering code, which will be painful but highly beneficial in the long term as it'll become significantly easier to reason about and modify, as well as being less prone to fun little lifetime sinkholes.


rust/template/differential_datalog/src/program/worker.rs
```diff
@@ -609,54 +610,79 @@ fn render_relation<'a>(
...
     delayed_vars: &DelayedVarMap<Child<'a, Worker<Allocator>, TS>>,
 ) {
-    // Relation may already be in the map if it was created by an `Apply` node
-    let mut collection = collections.remove(&relation.id).unwrap_or_else(|| {
-        let (session, collection) = scope.new_collection::<DDValue, Weight>();
-        sessions.insert(relation.id, session);
-
-        collection
-    });
+    scope.clone().region_named(relation.name(), |region| {
```
ryzhyk, line 613: So we're using regions to make relation/rule metadata visible to external tools (e.g., debuggers). I didn't realize this would be the case, but I guess it makes sense.  For operators, I was under the impression that they all now have `_named` versions, so we could use those instead, but it looks like there's only a small number of `_named` operators (`threshold_named`, `reduce_named`, and a few more).

Do you know if regions have any runtime cost?

mbudiu-vmw:
We should talk about the long term plans for the compiler.
I believe that the intention was to introduce another layer in the IR, which is closer to the dataflow graph.
This will be a significant refactoring of the code. I hope that some of these changes will not need to be completely reimplemented.

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,14 +85,17 @@ const PROF_MSG_BUF_SIZE: usize = 10_000;
 pub type Response<X> = Result<X, String>;
 
 /// Unique identifier of a DDlog relation.
+// TODO: Newtype this for type-safety
```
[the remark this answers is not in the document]
Kixiron: [`#[repr(transparent)]`](https://doc.rust-lang.org/nomicon/other-reprs.html#reprtransparent) struct wrappers will have the exact same abi is the inner type, so ```rust
#[repr(transparent)]
struct RelationId(usize);
```
will have the same abi as a `usize`

**Comment by Kixiron** [comment]
The issue as of now seems to be [this piece of code](https://github.com/vmware/differential-datalog/blob/master/rust/template/differential_datalog/src/program/mod.rs#L1451) and the associated pointer casts that are basically transmutes
image

**Comment by Kixiron** [comment]
Well that was exceedingly painful and makes me ever more convinced that we need to switch to a `Vec<XFormCollection>` design instead of the linked-list style approach we have now. Hopefully tests pass now

**Comment by ryzhyk** [comment]
> Well that was exceedingly painful 

It certainly looks that way.

> and makes me ever more convinced that we need to switch to a `Vec<XFormCollection>` design instead of the linked-list style approach we have now. Hopefully tests pass now

The only reason we're not doing this is because depending on the kind of transform the next element in the list can be either `XFormCollection` or `XFormArrangement`.

**Comment by Kixiron** [comment]
I'm fairly confident that can be made nice to work with given nicely-segmented structs and all that jazz, at minimum an enum

**Comment by ryzhyk** [comment]
> I'm fairly confident that can be made nice to work with given nicely-segmented structs and all that jazz, at minimum an enum

I think so. It will sacrifice a bit of compile-time type checking, which is still better than having all these hacks.

**Comment by Kixiron** [comment]
I think it's fixed, but now some of the tests time out for some wonderful reason

File: rust/template/Cargo.toml [hunk 3]
```diff
@@ -14,29 +14,15 @@ command-line = ["cmd_parser", "rustop"]
...
-optional = true
-
 [dependencies]
 abomonation = "0.7"
 time = { version = "0.2", features = ["serde"] }
 ordered-float = { version = "2.0.0", features = ["serde"] }
 cpuprofiler = { version = "0.0", optional = true }
 #differential-dataflow = "0.11.0"
-differential-dataflow = { git = "https://github.com/ddlog-dev/differential-dataflow", branch = "ddlog-3" }
+differential-dataflow = { git = "https://github.com/ddlog-dev/differential-dataflow", branch = "ddlog-4" }
```
ryzhyk, line 23: Can this please be a separate commit?

rust/template/differential_datalog/src/program/arrange.rs
```diff
@@ -18,10 +18,22 @@ use differential_dataflow::{
 use fnv::FnvHashMap;
 use std::ops::{Add, Mul, Neg};
 use timely::{
-    dataflow::scopes::{Child, Scope, ScopeParent},
+    dataflow::scopes::{Child, Scope},
     progress::{timestamp::Refines, Timestamp},
 };
 
+#[derive(Clone)]
```
ryzhyk, line 25: Nice!

CHANGELOG.md
```diff
@@ -8,6 +9,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changes
 
 - Upgraded to timely dataflow and differential dataflow dependencies to v0.12.
+- Worked on improving the debuggability of ddlog dataflow graphs
+
+### Bug Fixes
+
+- Fixed crashes due to misused unsafe code
```
ryzhyk, line 16: I don't think this needs to be in the changelog. The code only became buggy when nested scoped were introduced, so there was never a bug in the master branch; it only transiently existed in this PR.

rust/template/differential_datalog/src/program/arrange.rs
```diff
@@ -18,10 +18,22 @@ use differential_dataflow::{
...
+pub enum ArrangementFlavor<S, T>
+where
+    S: Scope,
+    S::Timestamp: Lattice + Refines<T>,
+    T: Lattice + Timestamp,
+{
+    Local(Arrangement<S, Weight, TValAgent<S::Timestamp>, TKeyAgent<S::Timestamp>>),
```
ryzhyk, line 32: Please add some comments, probably along the lines of "arrangement created in the local scope", "arrangement imported from the parent scope".

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -1248,19 +1250,21 @@ impl Program {
...
+    //       combination of traits and `Vec<XFormCollection>`s (as opposed to
+    //       what we do now with a linked list of them)
+    fn xform_collection<'a, S, T, Lookup>(
+        col: Collection<S, DDValue, Weight>,
         xform: &Option<XFormCollection>,
-        arrangements: &Arrangements<'a, 'b, P, T>,
-        lookup_collection: &LC,
+        arrangements: &Arrangements<'a, S, T>,
+        lookup_collection: Lookup,
         is_top_level_scope: bool,
```
ryzhyk, line 1261: I think this flag should no longer be needed.

rust/template/differential_datalog/src/program/worker.rs
```diff
@@ -608,54 +609,79 @@ fn render_relation<'a>(
...
+            .map(|(&arr_id, arr)| (arr_id, arr.enter_region(region)))
+            .collect();
 
-    if rule_collections.len() > 1 {
-        collection = with_prof_context(&format!("concatenate rules for {}", relation.name), || {
-            collection.concatenate(rule_collections)
+        // apply rules
+        // TODO: Regions for rules
+        let rule_collections = relation.rules.iter().map(|rule| {
+            let get_rule_collection = |rule_id| {
```
ryzhyk, line 631: This variable should be called rel_id, not rule_id.

rust/template/differential_datalog/src/program/mod.rs
Kixiron: For strongly connected rules we pass `false` to `is_top_level_collection`, I guess I'd make a `mk_rule_streamless` function for that? I really don't like it, but I'll try and address the code duplication in successive PRs

ryzhyk: Nah, let's not introduce code duplication in the first place. Just ignore my comment.

Kixiron: Actually, it may be 100% fine to allow `mk_rule()` to create recursive streams, I'm pretty sure it's valid within timely

ryzhyk: PS. I am still reviewing this and might have more comments.
PPS. It looks like tests are now passing (yay!) except some linter errors.

Kixiron: I've fixed the linting stuff locally, just wanted to push as many changes at once as possible to reduce CI times since I'm impatient

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -1274,19 +1278,118 @@ impl Program {
...
+
+            Self::xform_collection(
+                xformed.clone(),
+                &*next,
+                arrangements,
+                lookup_collection,
+                is_top_level_scope,
+            )
+        } else {
+            Self::streamless_xform_collection_ref(
```
ryzhyk: This might be wrong.  This assumes that only the first operator in a rule can be a stream xform and will break otherwise.  In practice, this means that if a streaming group_by occurs after some other operators, e.g., streaming join, things will break.  I will try to create a counterexample, but may not have the time until later today.

ryzhyk:
Duplication does suck, and this is what I was trying to avoid with unsafe conversions, but I guess this works.

Kixiron:
I'll address it in a follow up PR, it just requires some groundwork from this one

ryzhyk (approved):
Looks good; I'll fix a small typo and merge (just waiting for other pipelines to complete).

ryzhyk:
I notice that tests run more slowly in this branch. Might be load variations on the CI machine, but I want to investigate.

Kixiron:
Opened [timely-dataflow/#382](https://github.com/TimelyDataflow/timely-dataflow/issues/382)

rust/template/differential_datalog/src/program/config.rs
```diff
@@ -0,0 +1,174 @@
...
+#[derive(Debug, Clone, Copy)]
+pub struct Config {
+    /// The number of timely
+    pub num_timely_workers: usize,
+    /// Whether extra regions should be added to the dataflow
+    ///
+    /// These extra regions *significantly* help with the readability
+    /// of the generated dataflows at the cost of a minor performance
+    /// penalty. Best used with [`ProfilingKind::TimelyProfiling`]
+    /// in order to see the
```
ryzhyk, line 25: Incomplete sentence.

rust/template/differential_datalog/src/program/worker.rs
```diff
@@ -592,90 +627,83 @@ impl<'a> DDlogWorker<'a> {
     }
 }
 
-fn render_relation<'a>(
+// TODO: Add back regions for relations
```
ryzhyk, line 630: So we're basically back to where we were before this PR?

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -988,70 +991,49 @@ enum Reply {
 }
 
 impl Program {
-    /// Instantiate the program with `nworkers` timely threads.
+    /// Instantiate the program with `number_workers` timely threads.
     pub fn run(&self, number_workers: usize) -> Result<RunningProgram, String> {
+        let config = Config {
+            num_timely_workers: number_workers,
+            ..Default::default()
```
ryzhyk, line 998: I might be missing something, but I think this will set profiling mode to `None`, so neither self-profiling, which is currently always enabled by default, nor native timely/differential profiling will work. Since `run_with_config` is currently not callable through the DDlog API, there is basically no way for users to enable any kind of profiling.

rust/template/differential_datalog/src/program/worker.rs
Kixiron: No, it's PR still introduces a bunch of named operators

ryzhyk: Got it. The PR's been around for a while and I'm starting to forget what's in it :)

rust/template/differential_datalog/src/program/mod.rs
ryzhyk: This is my only remaining concern (well, this and the failing pipeline).  Can we address those, merge, and throw a party?

Kixiron: Set the profiling mode to `SelfProfiling` in `Program::run`, now waiting on CI

ryzhyk (approved):

PR #934, opened 2021-03-10
ovsdb2ddlog: Add --intern-table CLI switch.

ryzhyk:
Added `--intern-table` flag to the compiler to declare input tables coming from
OVSDB as `Intern<...>`.  This is useful for tables whose records are copied
around as a whole and can therefore benefit from interning performance- and
memory-wise.  In the past we had to create a separate table and copy records
from the original input table to it while wrapping them in `Intern<>`.  With
this change, we avoid the extra copy and intern records as we ingest them
for selected tables.

I am not pushing any tests for the new feature with this commit.
Testing was done using the master branch of OVN that currently has a
bunch of nondet test failures, presumably due to race conditions in the
test harness.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,17 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   replacing the standard collections (`Set`, `Map`, `Vec`) with functional
   versions.
 
+### ovsdb2ddlog compiler
+
+- Added `--intern-table` flag to the compiler to declare input tables coming from
+  OVSDB as `Intern<...>`.  This is useful for tables whose records are copied
+  around as a whole and can therefore benefit from interning performance- and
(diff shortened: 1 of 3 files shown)


blp (approved):
Seems useful. Thank you! I did not review the Haskell.

Do we have a place to document ovsdb2tool? As is, the CHANGELOG adds useful advice on how to use the new option, but changelogs are not the first place I look for documentation.

ryzhyk:
> Seems useful. Thank you! I did not review the Haskell.
> 
> Do we have a place to document ovsdb2tool? As is, the CHANGELOG adds useful advice on how to use the new option, but changelogs are not the first place I look for documentation.

We don't have any documentation at the moment apart from the usage message printed by the compiler and a detailed comment in the Haskell source.

There should probably be a separate file in the `doc` folder documenting the compiler.

PR #940, opened 2021-03-15
Remove old files

ryzhyk:
Delete all old files in the generated Rust project.  We previously did
this for the `types` subdirectory only.  Now we do this for the rest of
the project.  This is needed to avoid compilation errors when a newer
DDlog compiler version changes the structure of the template.


Kixiron:
Why was this closed?

ryzhyk:
> Why was this closed?

It was buggy.  I will fix and reopen it.  I normally open a PR after the branch has passed Travis and gitlab tests.  Somehow I missed a Travis test failure with this one.

PR #945, opened 2021-03-18
Experimental D3log support.

ryzhyk:
This is initial, incomplete, and experimental compiler support for
D3log.  Syntactically, it enables rules of the form:

```
R(...) @node_id :- ...
```

where `R[T]` is an internal relation (i.e., not an input or output),
and `node_id` specifies the location where the output of the rule
must be sent.

The compiler replaces `R` with a pair of relations:
`input relation R[T]` and
`output relation __out_R[(Option<D3logLocationId>, T)]` and
rewrites all rules

`R(x) :- ...` as `__out_R[(None, x)]` and
`R(x) @node_id :- ...` as `__out_R[(Some(node_id), x)] `.

In order to help the D3log runtime connect outputs to remote inputs,
we extend HDDlog to implement a new trait:

```
pub trait D3log {
    /// Parses a value output by DDlog for a distributed relation.
    /// A distributed relation `R[T]` is a relation whose values can be
    /// produced and consumed on different nodes.  The compiler splits such
    /// relations into an input/output relation pair, where the former has the
    /// same name and record type as `R`, while the latter has an opaque type
    /// that internally contains a value of type `T` and location id of the
    /// node where the value must be sent.
    /// This method extracts this information from an output value and returns
    /// location id, _input_ relation id, and the inner value.
    ///
    /// Returns `Err(val)` if `relid` is not a distributed relation id or `val`
    /// is not a value that belongs to this relation, returns `Err(val)`.
    fn d3log_localize_val(&self, relid: RelId, val: DDValue) -> Result<(Option<D3logLocationId>, RelId, DDValue), DDValue>;
}
```

See `test/datalog_tests/lb.dl` for a toy D3log program and
`test/datalog_tests/lb_test/src/main.rs` for an example Rust client that
uses the new API.


Kixiron:
I can't help thinking that this could be better generalized into a ddlog facade of the `.exchange()` operator, that way even single-machine ddlog programs could distribute their data on a customizable basis and all programs could do the distribution in a programmatic way, like this

```prolog
R(x) :-
    S(x),
    Exchange(hash(x)).
```

ryzhyk:
> I can't help thinking that this could be better generalized into a ddlog facade of the `.exchange()` operator, that way even single-machine ddlog programs could distribute their data on a customizable basis and all programs could do the distribution in a programmatic way, like this
> 
> ```prolog
> R(x) :-
>     S(x),
>     Exchange(hash(x)).
> ```
> 
> This would also allow rule-level distribution granularity instead of clause-level and I think would be a more effective usage of timely's features & design

I think this is precisely what we're doing, just not using `.exchange`.  `lb.dl` example in this PR is similar to your example, just using slightly different syntax.

Whether this could be more easily achieved by building on the timely infrastructure is a good question.  I don't really understand `timely-communication`, but I suspect we'd have to extend it significantly to support the design outlined in  @convolvatron's document.

ryzhyk:
@mbudiu-vmw , @convolvatron , please merge this when/if it gets approved.

@Kixiron , we should discuss the alternative timely-based design at some point.  This is experimental stuff and I expect it to change more than once before it becomes stable.

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -524,13 +526,15 @@ compile d_unoptimized specname modules rs_code dir crate_types = do
...
-    let d = addDummyRel
-            $ simplifyDifferentiation
-            $ simplifyDelayedRels
-            $ optimize d_unoptimized
+    let d_optimized = addDummyRel
+                      $ simplifyDifferentiation
+                      $ simplifyDelayedRels
+                      $ optimize d_unoptimized
+    -- Compile away D3log location annotations.
+    let (d, d3log_rel_map) = processD3logAnnotations d_optimized
```
mbudiu-vmw, line 534: If you do this earlier then later passes may not have to worry about strange relations with annotations anymore.

src/Language/DifferentialDatalog/D3log.hs
```diff
@@ -0,0 +1,124 @@
...
+                   then rule{ ruleLHS = [rlhead'] }
+                   else rule)
+             $ progRules d
+
+-- 'Option<D3logLocationId>'
+locationSpecifierType :: Type
+locationSpecifierType = tUser oPTION_TYPE [tUser lOCATION_TYPE []]
+
+outRelName :: String -> String
+outRelName rel = scoped rel_scope $ "__out_" ++ rel_name
```
mbudiu-vmw, line 112: I would be more comfortable if we had a fresh name allocator

src/Language/DifferentialDatalog/Syntax.hs
```diff
@@ -640,6 +641,29 @@ instance Show Atom where
 atomIsDelayed :: Atom -> Bool
 atomIsDelayed a = atomDelay a /= delayZero
 
+data RuleLHS = RuleLHS { lhsPos         :: Pos
+                         -- D3log only: location to send the output record to.
```
mbudiu-vmw, line 645: should this comment be one line lower?

test/datalog_tests/lb.dl
```diff
@@ -0,0 +1,17 @@
...
+AllWorkers(workers) :-
+    Workers(worker_id),
+    var workers = worker_id.group_by(()).to_vec().
+
+relation BalancedData(dat: Payload)
+
+BalancedData(dat) @worker :-
+    Data(dat),
+    AllWorkers(workers),
+    var worker = workers.nth(dat.hash64() % workers.len()).unwrap_or_default().
```
mbudiu-vmw, line 17: Is there a way to get the hash64 for other data types?

test/datalog_tests/lb_test/src/main.rs
```diff
@@ -0,0 +1,91 @@
+use lb_ddlog::api::HDDlog;
```
mbudiu-vmw, line 1: why write the test in Rust instead of using the CLI?

mbudiu-vmw (approved):

src/Language/DifferentialDatalog/Compile.hs
ryzhyk: The reason this is here is to help keep error messages clear.  This is true for many other bits of syntax that could be desugared early on in the compilation pipeline, but then the user would have to deal with error messages over the desugared code that they did not write.

src/Language/DifferentialDatalog/D3log.hs
ryzhyk: Yes, this is a recurring problem in the compiler. The reason this is not completely trivial is because we'd have to thread state through otherwise purely function code, but it's probably still worth doing. Do you want to submit an issue?

test/datalog_tests/lb.dl
ryzhyk: It should work for any type.

test/datalog_tests/lb_test/src/main.rs
ryzhyk: The goal is to illustrate how the D3log runtime (that will be written in Rust) can extract location information from output.

PR #946, opened 2021-03-18
c_api.rs: Remove spurious assertion.

ryzhyk:
An assertion in c_api.rs was checking that the weight of each record in
a delta is +1 or -1.  This invariant no longer holds: streams and
multisets can have arbitrary multiplicities.


mbudiu-vmw (approved):

mbudiu-vmw:
We could keep it for relations

ryzhyk:
> We could keep it for relations

Right, but this wouldn't be the place to do it.

PR #950, opened 2021-03-26
tutorial: Describe performance impact of dereferencing pointers.

blp:
This is based on a commit message written by Leonid.

### DIFF
--- a/doc/tutorial/tutorial.md
+++ b/doc/tutorial/tutorial.md
@@ -2736,6 +2736,48 @@ TopScore(school, top_score) :-
     var top_score = Aggregate((school), group_max(student.sat_score)).
 ```
 
+### Performance pitfalls with pattern matching references
+
+`&` can pattern match a value stored by reference, so it is tempting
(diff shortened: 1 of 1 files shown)


doc/tutorial/tutorial.md
```diff
@@ -2736,6 +2736,48 @@ TopScore(school, top_score) :-
     var top_score = Aggregate((school), group_max(student.sat_score)).
 ```
 
+### Performance pitfalls with pattern matching references
```
mbudiu-vmw, line 2739: The term "smart pointer" appears for the first time in this section.
Perhaps a more extensive discussion of the Rust representation of data in the generated code is necessary.

blp: I was following the terminology that Leonid used in his commit message, but it does not exactly match the terminology used in the tutorial.  I think that the intended meaning is `Ref<>` or `Intern<>`.

Perhaps I should just write "reference" here.

blp: I replaced this by a version that eliminates the term "smart pointer".

PR #951, opened 2021-03-26
Left join

blp:
Add description of left join to tutorial.

### DIFF
--- a/doc/tutorial/tutorial.md
+++ b/doc/tutorial/tutorial.md
@@ -1241,7 +1241,9 @@ as *antijoin*.
 
 A *left join* includes all of the records in one relation plus fields
 from matching records from a second relation.  In DDlog, we can build
-left joins using negation.  For example:
+left joins using negation.  For example, we can extend the previous
+example with a new relation `EndpointSanitization` that includes every
+endpoint plus whether it is sanitized:
 
 ```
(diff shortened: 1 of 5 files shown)


mbudiu-vmw (approved):

doc/tutorial/tutorial.md
```diff
@@ -1221,22 +1221,45 @@ vec.filter(|v| v.len() > 1)   // [[1,2,3], [4,5,6]]
...
 as *antijoin*.
 
+#### Left join
+
+A *left join* includes all of the records in one relation plus fields
+from matching records from a second relation.  In DDlog, we can build
+left joins using negation.  For example:
+
+```
+output relation EndpointSanitization(ep: string, sanitized: bool)
```
ryzhyk, line 1247: Thanks for improving the docs!

To someone who jumped straight to the Left join section, it may not be obvious that there is a negation hiding inside `SanitizedEndpoint`. It may be a good idea to either point the reader to the definition of `SanitizedEndpoint` above or just inline its definition here:

```
EndpointSanitization(endpoint, true) :- 
    EndpointString(endpoint),
    not Blocklisted(endpoint).
```

blp: OK, I pushed an additional commit that does that. I don't know whether you'd prefer me to squash the new commit; let me know if so.

Issue #938, opened 2021-03-13
Error on valid FlatMap


Kixiron:
With this code

```prolog
typedef Row = Row { n: usize }
typedef Col = Col { n: usize }

typedef Cell = Cell { row: Row, col: Col }

typedef Value = Value { n: usize }

input relation InitialCells(cell: Cell, value: Value)

relation SolvedCells(cell: Cell)
SolvedCells(cell) :- InitialCells(cell, _).

relation PossibleValues(cell: Cell, value: Value)
PossibleValues(cell, value) :- (var cell, var value) = FlatMap(all_possibilities()).

function all_possibilities(): Vec<(Cell, Value)> {
    var possibilities = vec_with_capacity(9 * 9 * 9);
    for (row in range_vec(1, 9, 1)) {
        for (col in range_vec(1, 9, 1)) {
            for (value in range_vec(1, 9, 1)) {
                var cell = Cell {
                    .row = Row { row },
                    .col = Col { col },
                };
                var value = Value { value };

                possibilities.push((cell, value));
            }
        }
    };

    possibilities
}

```

DDlog gives this error

```
> ddlog -i code.dl

error: code.dl:14.1-16.1: Rule must start with a positive literal
PossibleValues(cell, value) :- (var cell, var value) = FlatMap(all_possibilities()).
^
```

It seems the error comes from [`ruleValidate`](https://github.com/vmware/differential-datalog/blob/master/src/Language/DifferentialDatalog/Validate.hs#L304-L313) but I'm not sure why since this was explicitly implemented in a154c94d49dd695270deba01026e6a8564013022

ryzhyk:
This is indeed invalid DDlog at the moment.  This is the same issue as #770.

Kixiron:
How is the flat map a negative rule, the error makes sense for a negation but not really on this

ryzhyk:
This applies to any rule that doesn't start with a positive literal.  Injecting a singleton relation in the body of such a rule makes it possible to evaluate it using the existing compilation algorithm.  In particular, we compile `FlatMap`, down to DD's `flatmap` operator, which requires an input collection.  `Singleton` would serve as such an input collection. Your rule would become:

```
PossibleValues(cell, value) :- Singleton(),
                                              (var cell, var value) = FlatMap(all_possibilities()).
```

Kixiron:
Ah, gotcha. As a side note we should probably special case this kind of situation into a `.to_stream()`/`.as_collection()`

[...]

Issue #97, opened 2018-09-19
Rules that express all constants are not allowed


mbudiu-vmw:
```
RHeapAllocation_Type(_heap, _type) :- var _heap = "<<main method array>>", var _type = "java.lang.String[]".
```

mbudiu-vmw:
Is this solved?

ryzhyk:
Sort of. We can now have rules that only have the head and not the body, e.g.,

```
Rel(5,true, {var x = true; x or false}).
```

But we do not yet allow rules that have the body, but the body only consists of constant expressions, e.g., this is _not_ allowed:

```
Rel(a,b,c) :- a = 5, b = true, c = false.
```

mbudiu-vmw:
These two do not seem to be semantically equivalent; the first one does not have field names.
Otherwise we could write a pass to convert the second to the first.

ryzhyk:
Yes, it should be easy to just compile the second form directly to DD rather than reducing it to the first form.

ryzhyk:
Closed via #953

PR #953, opened 2021-03-31
Allow rules that don't start with a positive literal.

### LINKED ISSUE #770: Automatically inject Singleton relation in rules that do not start with a positive literal
Perform the following conversions during parsing:
- `R() :- not R2(...).` into `R() :- Singleton(), not R2(...)`
- `R() :- var x = 5.` into `R() :- Singleton(), var x = 5.`

Currently, users have to perform similar conversion by hand.
This will also address #97.

ryzhyk:
See commit messages.

Resolves #770, #97, and #938.

CC @blp

### DIFF
--- a/src/Language/DifferentialDatalog/Validate.hs
+++ b/src/Language/DifferentialDatalog/Validate.hs
@@ -51,6 +51,9 @@ import Language.DifferentialDatalog.Attribute
 import Language.DifferentialDatalog.Error
 import Language.DifferentialDatalog.Config
 
+sINGLETON_RELATION :: String
+sINGLETON_RELATION = "ddlog_std::Singleton"
+
 -- | Validate Datalog program
 validate :: (MonadError String me, ?cfg::Config) => DatalogProgram -> me DatalogProgram
 validate d = do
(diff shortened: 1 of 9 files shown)


src/Language/DifferentialDatalog/Validate.hs
```diff
@@ -303,15 +306,21 @@ indexValidate d idx = do
...
 
 ruleValidate :: (MonadError String me, ?cfg::Config) => DatalogProgram -> Rule -> me Rule
-ruleValidate d rl@Rule{..} = do
-    when (not $ null ruleRHS) $ do
-        case head ruleRHS of
-             RHSLiteral True _ -> return ()
-             _                 -> err d (pos rl) "Rule must start with a positive literal"
-    mapIdxM_ (ruleRHSValidate d rl) ruleRHS
-    mapM_ (ruleLHSValidate d rl) ruleLHS
+ruleValidate d rl = do
```
mbudiu-vmw, line 309: One would expect validation not to modify the program.
We should have a clearer staging of program rewriting.

mbudiu-vmw (approved):
I should modify the souffle translator too to get rid of the Dummy relation.

mbudiu-vmw:
It looks like the Windows haskell download still does not work.

mbudiu-vmw:
https://stackoverflow.com/questions/32654493/stack-haskell-throws-tlsexception-in-windows

ryzhyk:
> https://stackoverflow.com/questions/32654493/stack-haskell-throws-tlsexception-in-windows

I saw this, but I don't know what's the command line equivalent of using internet explorer to navigate to a page to fetch the missing certificate.

PR #954, opened 2021-03-31
Attempt to fix certificate problems on windows

mbudiu-vmw:
I don't know how to test this locally, so I am reduced to trial and error.


tools/install-stack.sh
```diff
@@ -23,13 +23,15 @@ fetch_stack_windows() {
...
 # builds.
+pwd
 mkdir -p ~/.local/bin
 if [ "$(uname)" = "Darwin" ]; then
   retry fetch_stack_osx
 elif [ "$(uname)" = "Linux" ]; then
   retry fetch_stack_linux
 else
   retry fetch_stack_windows
+  choco install haskell-dev
```
ryzhyk, line 34: `choco` is not available on all Windows systems.  The script should only run this command if `$TRAVIS` env variable is defined.

ryzhyk:
@mbudiu-vmw , any chance we could finish and merge this?

mbudiu-vmw:
The specs says that there should be, though.

PR #958, opened 2021-04-20
D3log transient

convolvatron:
for Leonid


d3log/Makefile
```diff
@@ -0,0 +1,15 @@
+
```
ryzhyk, line 1: Since we're in this ecosystem, you probably want to use `cargo-make`

d3log/src/json_framer.rs
```diff
@@ -0,0 +1,106 @@
+use std::io::{ErrorKind, Error};
+use phf::{phf_map, phf_set};
+
+#[derive(Clone)]
+#[derive(Copy)]
+#[derive(PartialEq)]
```
Kixiron, line 6:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq)]
```

d3log/src/json_framer.rs
```diff
@@ -0,0 +1,106 @@
...
+                            let p = match std::str::from_utf8(&self.reassembly[..]){
+                                Ok(x) => x.to_string(),
+                                Err(_x) => return Err(Error::new(ErrorKind::Other, "utf8 error"))
+                            };
+                            n.push(p);
+                            self.reassembly.truncate(0);
+                        }
+                    }
+                    None => ()
+                }
```
Kixiron, line 93:
```suggestion
                if let Some(&k) = ENDS.get(c) {
                    if *k != self.w.pop().expect("mismatched grouping") {
                        return Err(Error::new(ErrorKind::Other, "mismatched grouping"));
                    }

                    if self.w.is_empty() {
                        let string = std::str::from_utf8(&self.reassembly[..])
                            .map_err(|_| Error::new(ErrorKind::Other, "utf8 error"))?
                            .to_owned();

                        n.push(string);
                        self.reassembly.truncate(0);
                    }
                }
```

d3log/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+
+type Node = D3logLocationId;
+
+use std::convert::TryFrom;
+fn matrix(mat:Vec::<Vec::<u64>>) -> Result<Batch, String> {
+    let mut b = Batch::new(DeltaMap::new());
+    let mut i = 0;
+    let relid = Relations::try_from("matrix::Matrix").map_err(|_| format!("Unknown relation {}", "Matrix"))?;
+
+    for c in mat.iter() {
```
Kixiron, line 27:
```suggestion
    for (i, c) in mat.iter().enumerate() {
```

d3log/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+use mm_ddlog::*;
+use differential_datalog::DeltaMap;
+use differential_datalog::{D3logLocationId};
+
+type Node = D3logLocationId;
+
+use std::convert::TryFrom;
+fn matrix(mat:Vec::<Vec::<u64>>) -> Result<Batch, String> {
+    let mut b = Batch::new(DeltaMap::new());
+    let mut i = 0;
```
Kixiron, line 24:
```suggestion
```

d3log/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+    let relid = Relations::try_from("matrix::Matrix").map_err(|_| format!("Unknown relation {}", "Matrix"))?;
+
+    for c in mat.iter() {
+        let mut j = 0;
+        for v in c.iter() {
+            b.insert(relid as usize, Matrix{i, j, v:(*v) as u32}.into_ddvalue(), 1);
+            j = j+1;
+        }
+        i=i+1;
```
Kixiron, line 33: Rust also has `+=` btw, so `i += 1` is nicer than `i = i + 1`

```suggestion
```

d3log/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+fn matrix(mat:Vec::<Vec::<u64>>) -> Result<Batch, String> {
+    let mut b = Batch::new(DeltaMap::new());
+    let mut i = 0;
+    let relid = Relations::try_from("matrix::Matrix").map_err(|_| format!("Unknown relation {}", "Matrix"))?;
+
+    for c in mat.iter() {
+        let mut j = 0;
+        for v in c.iter() {
+            b.insert(relid as usize, Matrix{i, j, v:(*v) as u32}.into_ddvalue(), 1);
+            j = j+1;
```
Kixiron, line 31:
```suggestion
```

d3log/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+
+use std::convert::TryFrom;
+fn matrix(mat:Vec::<Vec::<u64>>) -> Result<Batch, String> {
+    let mut b = Batch::new(DeltaMap::new());
+    let mut i = 0;
+    let relid = Relations::try_from("matrix::Matrix").map_err(|_| format!("Unknown relation {}", "Matrix"))?;
+
+    for c in mat.iter() {
+        let mut j = 0;
+        for v in c.iter() {
```
Kixiron, line 29:
```suggestion
        for (j, &v) in c.iter().enumerate() {
```

d3log/src/main.rs
```diff
@@ -0,0 +1,57 @@
...
+            b.insert(relid as usize, Matrix{i, j, v:(*v) as u32}.into_ddvalue(), 1);
```
Kixiron, line 30:
```suggestion
            b.insert(relid as usize, Matrix{i, j, v: v as u32}.into_ddvalue(), 1);
```

d3log/src/tcp_network.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    // prefer big endian for generic reasons, but we are travelling through
+    // a constricted u64 pipe
+    u128::from_le_bytes([i[0], i[1], i[2], i[3], p[0], p[1], 0, 0, 0, 0, 0,
+                         0,0,0,0,0])
+}
+
+pub fn nid_to_socketaddr(n:Node) -> SocketAddr {
+    let b = n.to_be_bytes();
+    SocketAddr::new(IpAddr::V4(Ipv4Addr::new(b[0], b[1], b[2], b[3])), u16::from_be_bytes([b[4], b[5]]))
```
Kixiron, line 45:
```suggestion
    let [a, b, c, d, e, f, ..] = n.to_be_bytes();
    SocketAddr::new(IpAddr::V4(Ipv4Addr::new(a, b, c, d)), u16::from_be_bytes([e, f]))
```

d3log/src/tcp_network.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+pub struct TcpNetwork {
+    peers : Arc::<Mutex::<HashMap<Node, Arc::<Mutex::<TcpStream>>>>>
```
Kixiron, line 49: `::` aren't required in types
```suggestion
    peers : Arc<Mutex<HashMap<Node, Arc<Mutex<TcpStream>>>>>
```

Kixiron, line 49: The double-synchronization here is fairly suspect to me

d3log/src/tcp_network.rs
```diff
@@ -0,0 +1,114 @@
...
+               std::io::Error> {
+            println!("send {}", nid);
+            Ok(Box::pin(async move {
+                let p = &mut (*self.peers).lock().await;
+                let encoded = Vec::<u8>::new();
+                // actually this creates a connection, so...kinda yeah
+                let c = self.soft_intern(nid).await.expect("cant fail");
+                let mut cl = c.lock().await;
+                cl.write_all(&encoded).await
+            }))
```
Kixiron, line 89: This can just be a future instead of returning one
```suggestion
    pub async fn send(&self, nid:Node, _b:Arc::<Mutex::<Batch>>) -> Result<(), std::io::Error> {
        println!("send {}", nid);

        let p = &mut (*self.peers).lock().await;
        let encoded = Vec::<u8>::new();
        // actually this creates a connection, so...kinda yeah
        let c = self.soft_intern(nid).await.expect("cant fail");
        let mut cl = c.lock().await;
        cl.write_all(&encoded).await
```

d3log/src/tcp_network.rs
```diff
@@ -0,0 +1,114 @@
...
+    // system defined relations
+
+    let w = Workers{x:nid}.into_ddvalue();
+    output_json(&(singleton("d3::Workers", &w)?)).await?;
+    loop {
+        let (socket, a) = listener.accept().await?;
+        output_json(&(singleton("Connection", &(address_to_nid(a) as u64).ddvalue())?)).await?;
+        // xxx - socket errors shouldn't stop the listener
+        process_socket(n, socket).await?;
+    }
```
Kixiron, line 111: No exit condition?

d3log/src/transact.rs
```diff
@@ -0,0 +1,130 @@
...
+                    n:TcpNetwork::new(),
+                    h:hddlog, // arc?
+                    progress:Vec::new(),
+                }
+            }
+            Err(err) => {
+                // this is supposed to be management protocol
+                panic!("Failed to run differential datalog: {}", err);
+            }
+        }
```
Kixiron, line 58:
```suggestion
        let (hddlog, _init_output) = HDDlog::run(1, false)
            .unwrap_or_else(|err| panic!("Failed to run differential datalog: {}", err));

        TransactionManager{
            n: TcpNetwork::new(),
            h: hddlog, // arc?
            progress: Vec::new(),
        }
```

d3log/src/transact.rs
```diff
@@ -0,0 +1,130 @@
...
+                let node_batch =  match output.get(&loc_id.unwrap()) {
+                    Some(n) => n,
+                    None => {
+                        let loc = loc_id.unwrap();
+                        println!("add node bnatch {}", loc as u64);
+                        let n = Arc::new(Mutex::new(Batch::new(DeltaMap::new())));
+                        output.insert(loc, n);
+                        &output[&loc_id.expect("not actually")]
+                    }
+                };
```
Kixiron, line 84:
```suggestion
                let node_batch =  output.get(&loc_id.unwrap())
                    .unwrap_or_else(|| {
                        let loc = loc_id.unwrap();
                        println!("add node bnatch {}", loc as u64);
                        let n = Arc::new(Mutex::new(Batch::new(DeltaMap::new())));
                        output.insert(loc, n);
                        &output[&loc_id.expect("not actually")]
                    });
```

d3log/src/transact.rs
```diff
@@ -0,0 +1,130 @@
...
+        let upd = output.get(&nid);// get_mut?
+        // at minimum we should fork/join here
+        let b = upd.expect("cant happen");
+        // xxx - lock held across blocking operation?
+        // we should let these sends complete async - needs some sort of collection bucket that
+        // calls poll?
+        let n = {&(&(*tm.lock().await)).n};
+        let t = match n.send(*nid,b.clone()) {
+            Ok(x) => x, Err(x) => return Err(x)
+        };
```
Kixiron, line 106:
```suggestion
        let t = n.send(*nid,b.clone())?;
```

d3log/src/transact.rs
```diff
@@ -0,0 +1,130 @@
...
+    let mut upd = Vec::new();
+    for (relid, v, _) in input {
+        upd.push(Update::Insert {relid, v});
+    }
+    h.transaction_start()?;
+    match h.apply_updates(&mut upd.clone().drain(..)) {
+        Ok(()) =>  Ok(Batch::new(h.transaction_commit_dump_changes()?)),
+        Err(err) => {
+            println!("Failed to update differential datalog: {}", err);
+            return Err(err);
```
Kixiron, line 127:
```suggestion
            Err(err)
```

Kixiron:
At a glance you should probably

- Run rustfmt & clippy over the new code
- Maybe move the `d3log` project into the `/rust` folder? I'm not sure what @ryzhyk wants the project layout to be, we should probably figure that out sometime
- Consider using more descriptive variable names?

d3log/src/transact.rs
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+    }
+}
+
+    
+// no asynch trait
+// maybe this belongs in network?
+
+pub async fn forward(tm : Arc::<Mutex::<TransactionManager>>,  input: Batch)  -> Result<(), std::io::Error> {
```
ryzhyk, line 66: This should be a method of `TransactionManager` even if the first argument is not called `self`.

ryzhyk (commented):
Got this far today.

d3log/src/batch.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+impl<'de> Deserialize<'de> for Batch {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where D: Deserializer<'de> {
+        let b:Batch = deserializer.deserialize_any(BatchVisitor{})?;
+        Ok(b)
+    }
+}
```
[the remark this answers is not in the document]
ryzhyk: DeltaMap contains DDValue's internally. DDValue's don't have a generic Deserialize implementation; as we can only create one for a specific DDlog program. That's what `UpdateSerializer` is for. Happy to brainstorm a better solution, maybe using some kind of global type_id-to-deserialize_implementation table.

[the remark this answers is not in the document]
Kixiron: [`typetag`](https://crates.io/crates/typetag)?

d3log/src/tcp_network.rs
```diff
@@ -45,8 +45,8 @@ pub fn nid_to_socketaddr(n:Node) -> SocketAddr {
     SocketAddr::new(IpAddr::V4(Ipv4Addr::new(b[0], b[1], b[2], b[3])), u16::from_be_bytes([b[4], b[5]]))
 }
 
-pub struct TcpNetwork {
-    peers : Arc::<Mutex::<HashMap<Node, Arc::<Mutex::<TcpStream>>>>>
+pub struct TcpNetwork<'a> {
+    peers : &'a Arc::<Mutex::<HashMap<Node, Arc::<Mutex::<TcpStream>>>>>
```
Kixiron, line 49: What's the referencing for here? You shouldn't need lifetimes at all (much less getting a static one), you can just clone the `Arc` and that'll keep the data alive as long as references are outstanding

ryzhyk (commented):
Made some more progress, still haven't looked at the networking stuff.

d3log/src/tcp_network.rs
```diff
@@ -0,0 +1,125 @@
...
+//
+// there is an async_trait macro that promises to do this, but unsurprisingly
+// MutexGuard does not implement Send. So its unclear how to let such
+// a trait function mutate. punting.
+
+impl TcpNetwork<'_> {
+    pub async fn soft_intern(&self, nid:Node) -> Result<Arc::<Mutex::<TcpStream>>, std::io::Error> {
+        let p = &mut (*self.peers).lock().await;
+        
+        match p.get(&nid) {
```
ryzhyk, line 64: Again, the `Entry` API would be a cleaner choice here imo.

d3log/src/tcp_network.rs
ryzhyk: Why do we need to wrap this in `Arc<Mutex>` given that we wrap the entire `TcpNetwork` into that below?

d3log/src/tcp_network.rs
```diff
@@ -0,0 +1,125 @@
...
+pub struct ArcTcpNetwork<'a> {
+    n : Arc::<Mutex::<TcpNetwork<'a>>>
+}
+
+impl ArcTcpNetwork<'_> {
+    pub fn new () -> ArcTcpNetwork<'static>{
+        ArcTcpNetwork{n:Arc::new(Mutex::new(TcpNetwork::new()))}
+    }
+    
+    pub fn send(&'static self, nid:Node, _b:Arc::<Mutex::<Batch>>) ->
```
ryzhyk, line 89: Looks like this is still wip

d3log/src/transact.rs
```diff
@@ -0,0 +1,130 @@
...
+use async_std::sync::Mutex;
+use tokio::task;
+use differential_datalog::{DeltaMap, D3log, program::Update, DDlog, DDlogDynamic};
+
+pub type Timestamp = u64;
+
+
+pub struct TransactionManager {
+    // svcc needs the membership, so we are going to assign nids
+    progress : Vec<Timestamp>,
```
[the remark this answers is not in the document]
Kixiron: It should probably be an `Antichain` as well

d3log/src/child.rs
```diff
@@ -0,0 +1,138 @@
...
+                       return}
+        };
+    });
+}
+
+pub fn make_child() -> Result<(Fd, Fd), nix::Error> {
+    let (in_r, in_w) = pipe().unwrap();
+    let (out_r, out_w) = pipe().unwrap();
+
+    match unsafe{fork()} {
```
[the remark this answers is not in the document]
convolvatron: this one is ok. we're just doing the standard pattern of replacing the stdin and stdout of the child with the appropriate ends of pipes to the parent so it can talk to the children. its entirely possible we are leaking some parent file descriptors into the child if they aren't being closed on exec.

in the limit we should probably leave std/stdout alone for the children and use another channel for the control plane.

d3log/src/transact.rs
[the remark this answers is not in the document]
convolvatron: yes. this is just a placeholder for addtl per-batch metadata

d3log/src/batch.rs
```diff
@@ -0,0 +1,173 @@
...
+pub struct Batch {
+    // one might consider augmenting the values w/ t - kind of a
+    // performance problem for multisets
+    timestamp: u64, 
+    pub b: DeltaMap<differential_datalog::ddval::DDValue>,
+    // timestamp
+}
+
+impl Serialize for Batch {
+    // i would _like_ to expose an interface that used the names for relations
```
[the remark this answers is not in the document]
convolvatron: the issue isn't being to map the names, its that the compiler generates the deserialize path for UpdateSerializer, and that one demultiplexes on the ids. we'd have to generate another version or think of something else

d3log/src/batch.rs
```diff
@@ -0,0 +1,173 @@
...
+use std::fmt::Display;
+use std::io::{Error,ErrorKind};
+use std::collections::BTreeMap;
+
+// the other choice here would be a Vec<Update>?
+// need multiplicity
+// #[derive(Serialize, Deserialize)] - deltamap isn't serialize
+pub struct Batch {
+    // one might consider augmenting the values w/ t - kind of a
+    // performance problem for multisets
```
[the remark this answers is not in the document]
convolvatron: this is just the discussion we had about the cost of adding a clock to all tuples

d3log/src/batch.rs
```diff
@@ -0,0 +1,173 @@
+use differential_datalog::{DeltaMap, program::RelId, program::Update, ddval::DDValue};
+use mm_ddlog::*;
+use serde::{de, de::SeqAccess, de::Visitor, ser::SerializeTuple, Serialize, Deserialize, Serializer, Deserializer};
+use std::convert::TryFrom;
+use std::fmt;
+use std::fmt::Display;
+use std::io::{Error,ErrorKind};
+use std::collections::BTreeMap;
+
+// the other choice here would be a Vec<Update>?
```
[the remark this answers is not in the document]
convolvatron: yeah, that was half the point of implementing batch is just to paper over some of these differences. ideally for our case the ddlog input would be in the same format as the output

d3log/src/tcp_network.rs
convolvatron: when would you like to stop listening to new connections? are you think we should have an orderly shutdown button?

d3log/src/tcp_network.rs
convolvatron: this has all changed for the better I think. the new model (which compiles, but still has issues) is that send isn't async (which means a network can be a trait), and it asynchronously manages its own completions. 

it wasn't strictly necessary - but I wanted to actually try to have some asynchrony instead of just a bunch of async functions that always call await

d3log/src/tcp_network.rs
convolvatron: most of this spurious Arc-Mutex stuff Is me flailing around w/ the borrow checker. 

in this case its not completely clear though that one would have to hold the peer lock over any usage of a connection. I also don't know if reading and writing from different threads on the same socket needs to be locked (it kinda shouldn't..neither should calls to write(), but there may be additional internal state)

d3log/src/tcp_network.rs
convolvatron: yes, I have a new version that compiles, will update when I resolve all the review comments

d3log/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+
+[dependencies]
+tokio = {version = "1.4.0", features =["net", "macros", "rt-multi-thread", "io-util", "rt"]}
+tokio-fd = {version = "0.3.0"}
+rustop = { version = "1.0.2"}
+nix = { version = "0.20.0"}
+serde_json = "1.0.60"
+phf = {version="0.8", features = ["macros"]}
+#async-trait = "0.1.48"
+async-std = "1.9.0"
```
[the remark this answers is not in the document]
convolvatron: thanks. these should be tokio::sync::Mutex

d3log/Makefile
convolvatron: I will look at this, but I assume its safe to defer?

d3log/src/batch.rs
```diff
@@ -0,0 +1,173 @@
...
+    // performance problem for multisets
+    timestamp: u64, 
+    pub b: DeltaMap<differential_datalog::ddval::DDValue>,
+    // timestamp
+}
+
+impl Serialize for Batch {
+    // i would _like_ to expose an interface that used the names for relations
+    // so that external users dont have to be privy to the compiler id assignemnt
+    // xxx - u128 is not supported!
```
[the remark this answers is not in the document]
convolvatron: https://github.com/serde-rs/json/issues/448

looks like this is maybe still open?

d3log/src/tcp_network.rs
Kixiron: I would assume we'd want some sort of structured shutdown to do cleanup and all that fun stuff

mbudiu-vmw: The alternative view based on recovery oriented computing is that you will crash so there is no reason to have a clean showdown path. Just make sure you have a working recovery scheme.

d3log/src/transact.rs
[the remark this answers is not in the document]
convolvatron: idk why comments in GitHub can't be nested as threads .. anyways, could you go into more detail about why this should be Antichain?

d3log/src/tcp_network.rs
convolvatron: static and lifetimes are gone. I was trying to line things up so that lifetimes were static inside an async block so that the compiler would mark it send(). I got away with using clones of arcs.

d3log/src/main.rs
convolvatron: |                     v: v as u32,
   |                        -^^^^^^^
   |                        |
   |                        cannot cast `&u64` as `u32`
   |                        help: dereference the expression: `*v`

d3log/src/batch.rs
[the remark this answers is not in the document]
convolvatron: umm...I just changed this and its not throwing an error...no idea why I was having problems earlier. there _is_ an issue that according to json we can only send 64 bit floats, so depending on the actual implementation  - this may not make it to the other side intact

d3log/src/transact.rs
```diff
@@ -0,0 +1,130 @@
...
+pub async fn forward(tm : Arc::<Mutex::<TransactionManager>>,  input: Batch)  -> Result<(), std::io::Error> {
+    let mut output = HashMap::<Node, Arc::<Mutex::<Batch>>>::new();
+    for (rel, v, weight) in input {
+        let tma = &*tm.lock().await;
+        
+        // we dont really need an instance here, do we? - i guess the state comes from
+        // prog. we certainly dont need a lock on tm
+        match tma.h.d3log_localize_val(rel, v.clone()) {
+            Ok((loc_id, in_rel, inner_val)) => {
+                let node_batch =  match output.get(&loc_id.unwrap()) {
```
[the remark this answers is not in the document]
convolvatron: oh. cool.

d3log/src/tcp_network.rs
convolvatron: thanks - these aren't being carried forward

d3log/src/tcp_network.rs
convolvatron: I'm kinda with mbudiu. to me the cleanup in this case is primarily just to show that we _can_ do the cleanup (that we did all the accounting)....I can't think of a reason to actually free a structure or thread  which was explicitly allocated to live forever. you can imagine my frustration at waiting 60 seconds for a linux box to 'shut down'.  

but its not a big deal, willing to go w/ consensus here

d3log/src/transact.rs
convolvatron: this got substantially cleaner due to Leonids hash map entry suggestion, thanks for unwrap_or_else though

d3log/src/batch.rs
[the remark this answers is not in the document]
convolvatron: _oh_, right, this is actually important. based on current discussion, @ is really a kind of generic IPC mechanism. since we are explicitly allowing communication between different ddlog domains (compilations), it can't be the case that the compiler assigned enumeration is sufficient to locate the target relation. this is fundamentally a dynamic linkage, we should also discuss whether relation names are really sufficient (do we need to include schema, versioning, .. ?)

d3log/src/batch.rs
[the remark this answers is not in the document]
convolvatron: verified that this runs - I could swear .. never mind

d3log/src/batch.rs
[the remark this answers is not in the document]
ryzhyk: Tried that in the past, but it did not work well for me: https://github.com/dtolnay/typetag/issues/25

d3log/src/batch.rs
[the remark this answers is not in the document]
ryzhyk: Right.  This was originally designed to make a single DDlog program known at compile time run in a distributed fashion.  In order to dynamically instantiate and connect different programs we'd need more infrastructure.  Relation names/ids is just part of the problem. We'd also need to somehow enforce type safety without relying on the compiler.  Not in this PR I guess... Let's brainstorm this separately.

d3log/src/transact.rs
```diff
@@ -0,0 +1,130 @@
...
+        }
+    }
+}
+
+    
+// no asynch trait
+// maybe this belongs in network?
+
+pub async fn forward(tm : Arc::<Mutex::<TransactionManager>>,  input: Batch)  -> Result<(), std::io::Error> {
+    let mut output = HashMap::<Node, Arc::<Mutex::<Batch>>>::new();
```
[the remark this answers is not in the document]
convolvatron: I _think_ I can answer properly. Arc on its own is not Send or Sync, so it really forces you to use those two in combination to pass values in an async context, where the execution thread can really be anyone.

at least that's what I believe at the moment :)

ryzhyk:
@convolvatron , ok to close this PR, since D3log work this now lives in a [fork](https://github.com/convolvatron/differential-datalog)?

PR #964, opened 2021-04-26
Update flatbuffers installation

nandbarkin:
Use `make install` instead of `./tools/install-flatbuf.sh` as there is no such file `install-flatbuf.sh` within flatbuffers distribution.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,8 +122,8 @@ You are now ready to [start coding in DDlog](doc/tutorial/tutorial.md).
   that the `flatc` tool is in your `$PATH`.  Additionally, make sure that FlatBuffers
   Java classes are in your `$CLASSPATH`:
   ```
-  ./tools/install-flatbuf.sh
   cd flatbuffers
+  make install
   export CLASSPATH=`pwd`"/java":$CLASSPATH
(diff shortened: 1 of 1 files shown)


mbudiu-vmw:
The script is part of ddlog, not flatbuf. The script also downloads flatbuf, it does not just install it.

ryzhyk:
I agree with Mihai, the original script looks correct.

@nandbarkin , ok to close this PR?

nandbarkin:
@ryzhyk 
Thanks for the clarification. I needed to install flatbuffers to work with Java API.
However from the following text it wasn't clear whether flatbuffers was supposed to be downloaded manually or the script will do it or where exactly is the script.
So may be some better 

Google FlatBuffers library. Download and build FlatBuffers release 1.11.0 from github. Make sure that the flatc tool is in your $PATH. Additionally, make sure that FlatBuffers Java classes are in your `$CLASSPATH`:
```
./tools/install-flatbuf.sh
cd flatbuffers
export CLASSPATH=`pwd`"/java":$CLASSPATH
export PATH=`pwd`:$PATH
cd ..
```

PR #969, opened 2021-05-06
String conversions from/to utf8

mbudiu-vmw:
Also added a little print library function useful for debugging.

### DIFF
--- a/lib/print.dl
+++ b/lib/print.dl
@@ -1,0 +1,12 @@
+// These functions are useful for debugging.
+// print can be inserted in imperative code,
+// while printTrue can be inserted on the RHS of rules
+// e.g. O(x) :- I(v), printTrue(v), x = v + 1.
+
+#[has_side_effects]
+extern function print(msg: string): ()
+
+function printTrue(msg: string): bool {
+   print(msg);
+   true
+}
--- a/lib/print.dl
+++ b/lib/print.dl
@@ -6,6 +6,9 @@
 #[has_side_effects]
 extern function print(msg: string): ()
 
+#[has_side_effects]
+extern function debug(value: 'T): ()
+
 function printTrue(msg: string): bool {
    print(msg);
    true
--- a/lib/print.rs
+++ b/lib/print.rs
@@ -1,3 +1,9 @@
+use std::fmt::Debug;
+
 pub fn print(msg: &String) {
     print!("{}\n", msg)
+}
+
+pub fn debug<T: std::fmt::Debug>(value: &T) {
+    println!("{:?}", value);
 }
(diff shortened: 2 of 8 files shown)


lib/print.rs
```diff
@@ -0,0 +1,3 @@
+pub fn print(msg: &String) {
+    print!("{}\n", msg)
+}
```
Kixiron, line 3:
```suggestion
use std::fmt::Debug;

pub fn debug<T>(value: &T) {
    println!("{:?}", value);
}
```

Taking arbitrary values that are debuggable is heaps more versatile easier to use than only taking strings

lib/print.dl
```diff
@@ -0,0 +1,12 @@
...
+// while printTrue can be inserted on the RHS of rules
+// e.g. O(x) :- I(v), printTrue(v), x = v + 1.
+
+#[has_side_effects]
+extern function print(msg: string): ()
+
+function printTrue(msg: string): bool {
+   print(msg);
+   true
+}
```
Kixiron, line 12:
```suggestion
extern function debug<'T>(value: 'T): ()
```
Debug prints shouldn't be considered effectful and a specialization of the function that returns is redundant since we have `Inspect` clauses

Kixiron (changes requested):

lib/print.dl
ryzhyk: `has_side_effects` tells the compiler not to cache the result of the function, so it's absolutely necessary here.

lib/print.rs
ryzhyk: Good idea.  All DDlog types should implement `Debug`, I think, so this gives a way to print anything.
`print` might still be useful though, as it forces you to convert values to string using DDlog's `to_stream()` function, which is kind of similar to Rust's `Display`.
I vote we keep both.

lib/print.dl
ryzhyk: > specialization of the function that returns is redundant since we have Inspect clauses

`printTrue` can be used without `Inspect` though.  I don't mind either way.

ryzhyk:
PS. Please update CHANGELOG.

lib/print.dl
Kixiron: That's why printing should be "cached", so to speak: if printing forces re-computation you're no longer seeing the true behavior of the dataflow, you're seeing an altered version simply because our added a debug print

ryzhyk: All this attribute does is prevent converting function calls into `lazy_static`'s. It doesn't affect the dataflow in any way, but without it the output will only be produced the first time a rule containing the `print()` call is triggered.

PR #970, opened 2021-05-09
Timezone support

mbudiu-vmw:
I have also realized that no library files had copyright.
The previous implementation of time was based on a Rust crate called time, which used non-standard format specifiers and did not support time zones. So the implementation has been rewritten based on the chrono crate which uses C strftime format. However, the names of the functions were not changed, to prevent client breakage. The old library gave better error messages on malformed times.


mbudiu-vmw:
I should add more tests.
Was thinking about adding making tests easier to write; perhaps we should add some assert functions and have a compiler option to help track test coverage.

PR #975, opened 2021-05-13
Switched to triomphe for DDValue

Kixiron:
Triomphe doesn't allow weak references (which we don't use) so this lowers the overhead of `DDValue`s somewhat. Should have a decent memory boost, my benches also showed a nice little perf boost

```
twitter-micro/1 thread/150000 records
time:   [330.47 ms 332.58 ms 334.66 ms]
change: [-7.9898% -7.0025% -5.9926%] (p = 0.00 < 0.05)

twitter-micro/4 threads/200000 records
time:   [213.30 ms 214.84 ms 216.31 ms]
change: [-8.7423% -7.1495% -5.5009%] (p = 0.00 < 0.05)

twitter-micro/3 threads/200000 records
time:   [222.58 ms 224.66 ms 226.83 ms]
change: [-10.161% -8.6242% -7.1023%] (p = 0.00 < 0.05)

twitter-micro/2 threads/200000 records
time:   [287.95 ms 289.78 ms 292.07 ms]
change: [-9.8372% -8.5416% -7.1336%] (p = 0.00 < 0.05)

twitter-micro/1 thread/200000 records
time:   [440.43 ms 443.72 ms 447.15 ms]
change: [-8.0105% -6.7065% -5.2675%] (p = 0.00 < 0.05)

twitter-macro/4 threads/10000000 samples
time:   [150.55 s 162.49 s 173.91 s]
change: [-20.019% -11.893% -4.1555%] (p = 0.02 < 0.05)
```

### DIFF
--- a/rust/template/differential_datalog/src/ddval/ddval_convert.rs
+++ b/rust/template/differential_datalog/src/ddval/ddval_convert.rs
@@ -8,8 +8,8 @@ use std::{
     fmt::{self, Debug, Display, Formatter},
     hash::{Hash, Hasher},
     mem::{self, align_of, size_of, ManuallyDrop},
-    sync::Arc,
 };
+use triomphe::Arc;
 
 /// Trait to convert `DDVal` into concrete value type and back.
 pub trait DDValConvert: Sized {
@@ -149,7 +149,10 @@ where
             <*const usize>::cast::<Self>(&value.v).read()
         } else {
             let arc = Arc::from_raw(value.v as *const Self);
-            Arc::try_unwrap(arc).unwrap_or_else(|a| (*a).clone())
+
+            // If the `Arc` is uniquely held then simply take the inner value,
+            // otherwise clone it out
+            Arc::try_unwrap(arc).unwrap_or_else(|arc| Self::clone(&*arc))
         }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Changes
+
+- Made all reference counting (internal and through `ddlog_std::Ref`) use reference counters without weak counts
+  in an effort to reduce memory usage
+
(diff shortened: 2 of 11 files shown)


ryzhyk:
This is great!  I suppose, we can also use `triomphe` to replace `Arc` in the implementation of `Ref`?  I actually tried a similar crate (Arc without weak ref count) a while ago.  Unfortunately I don't remember what the issue was back then.  It may have required unstable features.

PS. Looking at Travis errors, this seems to break the C API.

ryzhyk:
This is a nice feature.  I'd love to merge it. @Kixiron , I'm not sure what local workflow you use.  I find it useful to run a subset of tests locally to catch most issues before pushing a commit to github, e.g., `./test.sh crates simple flatbuf0`.

rust/template/differential_datalog/src/ddval/ddval_convert.rs
```diff
@@ -149,7 +149,8 @@ where
             <*const usize>::cast::<Self>(&value.v).read()
         } else {
             let arc = Arc::from_raw(value.v as *const Self);
-            Arc::try_unwrap(arc).unwrap_or_else(|a| (*a).clone())
+            // FIXME: An equivalent of `Arc::try_unwrap()` for triomphe
+            Self::clone(&*arc)
```
ryzhyk, line 153: Isn't this terrible for performance: we're cloning every time even when we don't need to (which I expect to be the common case)?

ryzhyk (commented):
Now that it works, I actually looked at the code.  The extra clone bothers me, as `from_ddval` is used pretty ubiquitously in generated code.

Kixiron:
Yes, I was going to write an equivalent of `std::sync::Arc`'s `.try_unwrap()` but I never got around to it, I'll do it next week

Kixiron:
Opened [triomphe/#21](https://github.com/Manishearth/triomphe/pull/21)

Kixiron:
Got the features upstreamed into `triomphe` and upgraded to the newer version, we now use `Arc::try_unwrap(arc).unwrap_or_else(|arc| Self::clone(&*arc))` instead of unconditionally cloning

PR #982, opened 2021-05-24
Fix tutorial commit output

booxter:
'commit' doesn't print changes unless 'dump_changes' is used. Also, the
output of 'commit dump_changes' and 'dump' commands is slightly
different now.


mbudiu-vmw (approved):
I assume you have actually tested these.

Issue #983, opened 2021-05-24
Loosen restriction on `MapFunc` closure function type and also other types for more flexibility


qishen:
Hi, I'm trying to reuse some of the low-level DDlog APIs to run data with a tree-like data structure in the dataflow and I need less restriction on the closure function meaning I may want to pass an external variable into the `MapFunc` but the current type of `MapFunc` won't allow it. My workaround is to change the type to `MapDynFunc` shown below with dynamic dispatch and the capability to pass around between threads. 

It's very useful for situations when you need an external value passed into the closure to decide how to filter, map, or arrange the collection. Let me know what you guys think about this improvement and I can send a PR to possibly change the type of all closure functions.

```Rust
/// Function type used to map the content of a relation
/// (see `XFormCollection::Map`).
pub type MapFunc = fn(DDValue) -> DDValue;
pub type MapDynFunc = Arc<dyn Fn(DDValue) -> DDValue + Sync + Send>;

...

pub enum XFormCollection {
    ...
    /// Apply `mfun` to each element in the collection
    Map {
        description: Cow<'static, str>,
        mfun: MapDynFunc,
        next: Box<Option<XFormCollection>>,
    },
}

...

XFormCollection::Map {
    ref description,
    ref mfun,
    ref next,
} => {
    // Another workaround to avoid static lifetime for closure function.
    let mfn = mfun.clone();
    let mfn_closure = move |v: DDValue| {
        mfn.call((v,))
    };
    let mapped = with_prof_context(&description, move || col.map(mfn_closure));
    Self::streamless_xform_collection(mapped, &*next, arrangements, lookup_collection)
}

...

// We may need to pass a variable into the closure in some use cases.
let external_var = 1;
next: Box::new(Some(XFormCollection::Map {
    description: Cow::from("hello"),
    mfun: Arc::new(move |value: DDValue| -> DDValue {
        let inner_var = external_var;
        value
    }),
    next: Box::new(None)
})
```

ryzhyk:
Hmm, this is an unusual request. The `differential-datalog` crate that contains these definitions was designed for use by ddlog-generated code and is really not meant to be used manually this way.  It lacks any kind of ergonomics and offers numerous ways to shoot yourself in the foot.  Just curious: why do you want to use this crate instead of using differential dataflow directly?

[...]

PR #984, opened 2021-05-26
Loosen function type

qishen:
This is an incomplete PR for #983 to loosen the restriction on function types. For now I only changed four function types `FilterMapFunc`, `FilterFunc`, `JoinFunc`, `ArrangeFunc` for a review and will add another commit to change the rest of the function types if you guys think everything is good without breaking anything.

### DIFF
--- a/rust/template/differential_datalog/src/dataflow/filter_map.rs
+++ b/rust/template/differential_datalog/src/dataflow/filter_map.rs
@@ -4,19 +4,23 @@ use timely::{
     Data,
 };
 
+pub type FilterMapFunc<D, D2> = std::sync::Arc<dyn Fn(D) -> Option<D2>>;
+
 pub trait FilterMap<D, D2> {
     type Output;
 
-    fn filter_map<L>(&self, logic: L) -> Self::Output
-    where
-        L: FnMut(D) -> Option<D2> + 'static,
+    // fn filter_map<L>(&self, logic: L) -> Self::Output
+    // where
+    //     L: FnMut(D) -> Option<D2> + 'static,
+    fn filter_map(&self, logic: FilterMapFunc<D, D2>) -> Self::Output
     {
         self.filter_map_named("FilterMap", logic)
     }
 
-    fn filter_map_named<L>(&self, name: &str, logic: L) -> Self::Output
-    where
-        L: FnMut(D) -> Option<D2> + 'static;
+    // fn filter_map_named<L>(&self, name: &str, logic: L) -> Self::Output
+    // where
+    //     L: FnMut(D) -> Option<D2> + 'static;
+    fn filter_map_named(&self, name: &str, logic: FilterMapFunc<D, D2>) -> Self::Output;
 }
 
 impl<S, D, D2> FilterMap<D, D2> for Stream<S, D>
@@ -27,9 +31,10 @@ where
 {
     type Output = Stream<S, D2>;
 
-    fn filter_map_named<L>(&self, name: &str, mut logic: L) -> Self::Output
-    where
-        L: FnMut(D) -> Option<D2> + 'static,
+    // fn filter_map_named<L>(&self, name: &str, mut logic: L) -> Self::Output
+    // where
+    //     L: FnMut(D) -> Option<D2> + 'static,
+    fn filter_map_named(&self, name: &str, logic: FilterMapFunc<D, D2>) -> Self::Output
     {
         let mut buffer = Vec::new();
 
@@ -56,14 +61,17 @@ where
 {
     type Output = Collection<S, D2, R>;
 
-    fn filter_map_named<L>(&self, name: &str, mut logic: L) -> Self::Output
-    where
-        L: FnMut(D) -> Option<D2> + 'static,
+    // fn filter_map_named<L>(&self, name: &str, mut logic: L) -> Self::Output
+    // where
+    //     L: FnMut(D) -> Option<D2> + 'static,
+    fn filter_map_named(&self, name: &str, logic: FilterMapFunc<D, D2>) -> Self::Output
     {
         self.inner
-            .filter_map_named(name, move |(data, time, diff)| {
-                logic(data).map(|data| (data, time, diff))
-            })
+            .filter_map_named(name, 
+                std::sync::Arc::new(move |(data, time, diff)| {
+                    logic(data).map(|data| (data, time, diff))
+                })
+            )
             .as_collection()
     }
 }
--- a/rust/template/differential_datalog/src/dataflow/filter_map.rs
+++ b/rust/template/differential_datalog/src/dataflow/filter_map.rs
@@ -4,8 +4,6 @@ use timely::{
     Data,
 };
 
-pub type FilterMapFunc<D, D2> = std::sync::Arc<dyn Fn(D) -> Option<D2>>;
-
 pub trait FilterMap<D, D2> {
     type Output;
 
--- a/rust/template/differential_datalog/src/dataflow/filter_map.rs
+++ b/rust/template/differential_datalog/src/dataflow/filter_map.rs
@@ -12,7 +12,6 @@ pub trait FilterMap<D, D2> {
     fn filter_map<L>(&self, logic: L) -> Self::Output
     where
         L: FnMut(D) -> Option<D2> + 'static,
-    // fn filter_map(&self, logic: FilterMapFunc<D, D2>) -> Self::Output
     {
         self.filter_map_named("FilterMap", logic)
     }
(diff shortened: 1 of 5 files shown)


ryzhyk:
This won't work unless you also change the Haskell Compile.hs to generate these closures.

Also, I'm still not sure how this is better than using ddlog :)

rust/template/differential_datalog/src/dataflow/filter_map.rs
```diff
@@ -4,19 +4,23 @@ use timely::{
     Data,
 };
 
+pub type FilterMapFunc<D, D2> = std::sync::Arc<dyn Fn(D) -> Option<D2>>;
+
 pub trait FilterMap<D, D2> {
     type Output;
 
-    fn filter_map<L>(&self, logic: L) -> Self::Output
```
Kixiron, line 10: You don't need to alter the dataflow operators to do this

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -110,49 +110,62 @@ pub type ArrId = (RelId, usize);
 /// Function type used to map the content of a relation
 /// (see `XFormCollection::Map`).
 pub type MapFunc = fn(DDValue) -> DDValue;
+pub type MapDynFunc = Arc<dyn Fn(DDValue) -> DDValue + Sync + Send>;
```
Kixiron, line 113: I actually think boxing and having the user take responsibility for synchronizing is better & lower overhead

Kixiron (changes requested):

rust/template/differential_datalog/src/program/mod.rs
qishen: Changing from `Arc` to `Box` causes some troubles because `MapDynFunc` is used in the field of `XFormCollection` which requires `Clone` but you can't add `Clone` to trait object inside `MapDynFunc`.

rust/template/differential_datalog/src/dataflow/filter_map.rs
qishen: I'm a little confused with this comment. Which part I don't need to alter here?

Kixiron: The function signatures here shouldn't change, I'll make code suggestions

rust/template/differential_datalog/src/dataflow/filter_map.rs
```diff
@@ -4,19 +4,23 @@ use timely::{
...
-    where
-        L: FnMut(D) -> Option<D2> + 'static,
+    // fn filter_map<L>(&self, logic: L) -> Self::Output
+    // where
+    //     L: FnMut(D) -> Option<D2> + 'static,
+    fn filter_map(&self, logic: FilterMapFunc<D, D2>) -> Self::Output
```
Kixiron, line 15:
```suggestion
    fn filter_map<L>(&self, logic: L) -> Self::Output
    where
        L: FnMut(D) -> Option<D2> + 'static,
```

rust/template/differential_datalog/src/dataflow/filter_map.rs
```diff
@@ -4,19 +4,23 @@ use timely::{
...
         self.filter_map_named("FilterMap", logic)
     }
 
-    fn filter_map_named<L>(&self, name: &str, logic: L) -> Self::Output
-    where
-        L: FnMut(D) -> Option<D2> + 'static;
+    // fn filter_map_named<L>(&self, name: &str, logic: L) -> Self::Output
+    // where
+    //     L: FnMut(D) -> Option<D2> + 'static;
+    fn filter_map_named(&self, name: &str, logic: FilterMapFunc<D, D2>) -> Self::Output;
```
Kixiron, line 23:
```suggestion
    fn filter_map_named<L>(&self, name: &str, logic: L) -> Self::Output
    where
        L: FnMut(D) -> Option<D2> + 'static;
```

rust/template/differential_datalog/src/dataflow/filter_map.rs
```diff
@@ -27,9 +31,10 @@ where
 {
     type Output = Stream<S, D2>;
 
-    fn filter_map_named<L>(&self, name: &str, mut logic: L) -> Self::Output
-    where
-        L: FnMut(D) -> Option<D2> + 'static,
+    // fn filter_map_named<L>(&self, name: &str, mut logic: L) -> Self::Output
+    // where
+    //     L: FnMut(D) -> Option<D2> + 'static,
+    fn filter_map_named(&self, name: &str, logic: FilterMapFunc<D, D2>) -> Self::Output
```
Kixiron, line 37:
```suggestion
    fn filter_map_named<L>(&self, name: &str, mut logic: L) -> Self::Output
    where
        L: FnMut(D) -> Option<D2> + 'static,
```

rust/template/differential_datalog/src/dataflow/filter_map.rs
```diff
@@ -56,14 +61,17 @@ where
 {
     type Output = Collection<S, D2, R>;
 
-    fn filter_map_named<L>(&self, name: &str, mut logic: L) -> Self::Output
-    where
-        L: FnMut(D) -> Option<D2> + 'static,
+    // fn filter_map_named<L>(&self, name: &str, mut logic: L) -> Self::Output
+    // where
+    //     L: FnMut(D) -> Option<D2> + 'static,
+    fn filter_map_named(&self, name: &str, logic: FilterMapFunc<D, D2>) -> Self::Output
```
Kixiron, line 67:
```suggestion
    fn filter_map_named<L>(&self, name: &str, mut logic: L) -> Self::Output
    where
        L: FnMut(D) -> Option<D2> + 'static,
```

rust/template/differential_datalog/src/dataflow/filter_map.rs
```diff
@@ -56,14 +61,17 @@ where
...
     {
         self.inner
-            .filter_map_named(name, move |(data, time, diff)| {
-                logic(data).map(|data| (data, time, diff))
-            })
+            .filter_map_named(name, 
+                std::sync::Arc::new(move |(data, time, diff)| {
+                    logic(data).map(|data| (data, time, diff))
+                })
+            )
```
Kixiron, line 74:
```suggestion
            .filter_map_named(name, move |(data, time, diff)| {
                 logic(data).map(|data| (data, time, diff))
            })
```

Kixiron (changes requested):

qishen:
> 
> 
> This won't work unless you also change the Haskell Compile.hs to generate these closures.

Oh I see the code generation part in `Compile.hs` that fills some code snippet templates. The task to change `Compile.hs` looks a little intimidating to me and I have no experience with Haskell. I don't want to add too much burden to your guys for a requested feature but I appreciate it if one of you can help me on `Compile.hs` later. I think somewhere in the code there is a TODO mentioning changing to dynamic dispatch for function types too so I guess this improvement is not a bad idea.

> Also, I'm still not sure how this is better than using ddlog :)

Well I'm in the middle of convincing myself to use ddlog directly. 

```
input relation NN(id: u32)
output relation EE(src: NN, dst: NN)

EE(NN(a+1), NN(b+1)) :- EE(NN(a), NN(b)).
```

A quick question: can we write rules like the one above? I get the error message: `error: graph.dl:13.28-13.30: Missing field 'id' in constructor NN
EE(NN(a+1), NN(b+1)) :- EE(NN(a), NN(b)).`

mbudiu-vmw:
Constructors in DDlog use curly braces:
`EE(NN{a+1}, NN{b+1}) :- EE(NN{a}, NN{b}).`

qishen:
> 
> 
> Constructors in DDlog use curly braces:
> `EE(NN{a+1}, NN{b+1}) :- EE(NN{a}, NN{b}).`

Thanks @mbudiu-vmw and you may want to fix them in the documentation too because in many places I saw the nested expression in parentheses.

mbudiu-vmw:
We are supposed to have test code for all the documentation code, but perhaps we missed some cases.

ryzhyk:
@qishen , where did you see that in the docs?

qishen:
@ryzhyk In the language reference. I think I was wrong because `f()` is actually a function not a constructor.
```
R(x,y) :- S(f(x)), T(x, y). // illegal, as f(x) is not a pattern.
R(x,y) :- S(x), T(f(x), y). // ok, x is introduced before being used in f(x)
``

ryzhyk:
That's right, function calls use parens, constructors use curly braces.

mbudiu-vmw:
Changing the Haskell file should probably not be too difficult, you just need to look in Compile.hs, and if you just grep for the function types you have changed it may be sufficient to figure out where you need to make the changes.

qishen:
@Kixiron I changed `Arc` to `Box` with `dyn_clone` only for the map function and I'll change the rest if you think that's the correct way to write it.

Kixiron:
I'm good with either of them so I'll let @ryzhyk decide, I don't think enforcing synchronization by default is useful and these closures would only be cloned as many times as there are ddlog workers so the cloning overhead is minimal. If a user needs sync I think it's simple enough to have them do it themselves instead of ending up with what's essentially an `Arc<Arc<UserData>>`

ryzhyk:
@qishen , are you still interested in this PR or is it ok to close it?

qishen:
@ryzhyk I think it's ok to close it since I'm not using ddlog as a library anymore. Thanks!

PR #998, opened 2021-06-16
[Tutorial] Clarify distinction between rules and functions

amytai:


doc/tutorial/tutorial.md
```diff
@@ -117,6 +117,20 @@ A comma in a rule is read as "and".  Rules end with a mandatory dot.
...
 
+### Computation
+In DDlog, the two main compute constructs are
+[rules](../language_reference/language_reference.md#rules) and
+[functions](../language_reference/language_reference.md#functions).
+
+Rules describe how to compute an output relation from its input relations.
+The available computations expressable in rules are similar to common queries
+in traditional relational databases,
+including select, join, filter, map, and groupby.
```
mbudiu-vmw, line 128: and aggregates

doc/tutorial/tutorial.md
```diff
@@ -117,6 +117,20 @@ A comma in a rule is read as "and".  Rules end with a mandatory dot.
...
+
+Functions in DDlog are similar to user-defined functions in relational languages
+and enable imperative expression of computation via traditional constructs
+such as control flow and complex data types.
```
mbudiu-vmw, line 132: I would add that functions are designed to always operate on values that come from some relation.

mbudiu-vmw (changes requested):

doc/tutorial/tutorial.md
amytai: done

doc/tutorial/tutorial.md
amytai: done

mbudiu-vmw (approved):

doc/tutorial/tutorial.md
```diff
@@ -117,6 +117,21 @@ A comma in a rule is read as "and".  Rules end with a mandatory dot.
 In database terminology the above rule computes an inner join of
 the `Word1` and `Word2` relations on the `cat` field.
 
+### Computation
+In DDlog, the two main compute constructs are
+[rules](../language_reference/language_reference.md#rules) and
```
ryzhyk, line 122: This is great, thanks!  Since the language reference document is not particularly useful at the moment, I'd prefer that these links pointed to relevant sections of this document.  I don't think we have a separate section for rules, which is ok, as we just gave and example of a rule above, but there's a section for functions.

amytai: Incidentally, I created this pull request because I was struggling to find good documentation about rules syntax, and the language manual was particularly enlightening (for me). But of course, I can remove the links if you prefer.

I was also thinking about linking to relevant sections in the tutorial (`Advanced rules` section for rules and `Functions` for functions). However, especially for functions, the relevant material is spread a little across multiple sections. Perhaps we could use [here](https://github.com/vmware/differential-datalog/blob/master/doc/tutorial/tutorial.md#expressions) as the link for functions. What do you think?

ryzhyk: In that case, I'll go ahead and merge it as is.

PR #999, opened 2021-06-16
Replace Travis CI with GitHub workflow.

ryzhyk:
Also two minor triomphe-related fixes.  See commit messages.


.github/workflows/main.yml
```diff
@@ -0,0 +1,85 @@
...
+        # build_distro.sh uses 'ghead' on MacOS
+      - name: Install coreutils on MacOS
+        if: ${{ github.event_name == 'release'  &&  runner.os == 'macOS' }}
+        run: brew install coreutils
+      - name: Build distro
+        if: ${{ github.event_name == 'release' }}
+        run: ./build_distro.sh
+      - name: Upload distro
+        if: ${{ github.event_name == 'release' }}
+        uses: alexellis/upload-assets@0.3.0
```
mbudiu-vmw, line 81: this is like a magic language. if you know the right incantations, it works.

mbudiu-vmw (approved):
I approve, but really I do not know how to read these actions.

Kixiron (approved):
Looks good as a minimal workflow, we can add fancy stuff later on

.github/workflows/main.yml
Kixiron, line 81:
```suggestion
      - name: Upload distro
        if: ${{ github.event_name == 'release' && github.ref == 'refs/heads/master' }}
        uses: alexellis/upload-assets@0.3.0
```
This should make sure that it only tries to upload when it's on master instead of failing on other branches and forks

Kixiron (changes requested):

.github/workflows/main.yml
ryzhyk: `event_name == 'release'` means that this will get triggered with we manually create a release on github, in which case we want to build release artifacts regardless of the branch.

PR #1000, opened 2021-06-16
Improved the generated code

Kixiron:
* The generated code is more robust now, it uses more absolute paths which means that type confusion is less likely
* The generated code is also easier to read now which satisfies my OCD
* Primary task before I got really distracted: Relations and arrangement name strings now include their ddlog source location so that users can look at a rule's profiling info and see where it came from in their written code
* Runtime type checking only happens in debug mode within generated code, types are checked statically by the ddlog compiler so we're good
* The spans of input events are now logged using timely's logging infrastructure, should help to get better data out of ddshow

### DIFF
--- a/src/Language/DifferentialDatalog/Rule.hs
+++ b/src/Language/DifferentialDatalog/Rule.hs
@@ -42,11 +42,13 @@ module Language.DifferentialDatalog.Rule (
     ruleBodyIsStream
 ) where
 
+import Prelude hiding ((<>))
 import qualified Data.Set as S
 import Data.List
 import Data.Functor.Identity
 --import Debug.Trace
 import Text.PrettyPrint
+import Text.Parsec (sourceName, sourceLine, sourceColumn)
 
 import Language.DifferentialDatalog.PP
 import Language.DifferentialDatalog.Syntax
@@ -73,7 +75,12 @@ rulePPPrefix rl len = commaSep $ map pp $ take len $ ruleRHS $ ruleStripTypeAnno
 
 -- | Pretty-print rule without type annotations.
 rulePPStripped :: Rule -> Doc
-rulePPStripped rl = pp $ ruleStripTypeAnnotations rl
+rulePPStripped rule = pp (ruleStripTypeAnnotations rule) <+> char '@' <+> file <> char ':' <> line <> char ':' <> column
+    where
+        (position, _) = rulePos rule
+        file = text (sourceName position)
+        line = int (sourceLine position)
+        column = int (sourceColumn position)
 
 -- | New variables declared in the 'i'th conjunct in the right-hand
 -- side of a rule.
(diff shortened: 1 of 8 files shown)


Kixiron:
For reference, what used to be roughly this
```rust
    let Arithm = program::Relation {
                     name: std::borrow::Cow::from("Arithm"),
                     input: false,
                     distinct: true,
                     caching_mode: program::CachingMode::Set,
                     key_func: None,
                     id: 18,
                     rules: vec![
                     ],
                     arrangements: vec![
                     ],
                     change_cb:    Some(std::sync::Arc::clone(&__update_cb))
                 };
    let BMethod = program::Relation {
                      name: std::borrow::Cow::from("BMethod"),
                      input: true,
                      distinct: false,
                      caching_mode: program::CachingMode::Set,
                      key_func: None,
                      id: 19,
                      rules: vec![
                      ],
                      arrangements: vec![
                          types::__Arng_BMethod_0.clone()
                      ],
                      change_cb:    None
                  };
```
Is now this
```rust
    let Arithm = ::differential_datalog::program::Relation {
        name: ::std::borrow::Cow::from("Arithm @ .\\simple.dl:1033:1"),
        input: false,
        distinct: true,
        caching_mode: ::differential_datalog::program::CachingMode::Set,
        key_func: None,
        id: 18,
        rules: vec![
        ],
        arrangements: vec![
        ],
        change_cb: ::core::option::Option::Some(::std::sync::Arc::clone(&__update_cb)),
    };
    let BMethod = ::differential_datalog::program::Relation {
        name: ::std::borrow::Cow::from("BMethod @ .\\simple.dl:1010:1"),
        input: true,
        distinct: false,
        caching_mode: ::differential_datalog::program::CachingMode::Set,
        key_func: None,
        id: 19,
        rules: vec![
        ],
        arrangements: vec![
            types::__Arng_BMethod_0.clone(),
        ],
        change_cb: ::core::option::Option::None,
    };
```

src/Language/DifferentialDatalog/Rule.hs
```diff
@@ -73,7 +75,12 @@ rulePPPrefix rl len = commaSep $ map pp $ take len $ ruleRHS $ ruleStripTypeAnno
 
 -- | Pretty-print rule without type annotations.
 rulePPStripped :: Rule -> Doc
-rulePPStripped rl = pp $ ruleStripTypeAnnotations rl
+rulePPStripped rule = pp (ruleStripTypeAnnotations rule) <+> char '@' <+> file <> char ':' <> line <> char ':' <> column
```
ryzhyk, line 78: What's wrong with `pp $ rulePos rule`?

Kixiron: I don't know, I don't think it has the same output, does it?

ryzhyk: My bad, I was thinking of this function in `Pos.hs`.  It prints the entire range, not just its start.

```
spos :: (WithPos a) => a -> String
spos x = let (s,e) = pos x
         in sourceName s ++ ":" ++ (show $ sourceLine s) ++ "." ++ (show $ sourceColumn s) ++ "-"
                                ++ (show $ sourceLine e) ++ "." ++ (show $ sourceColumn e)
```

PR #1007, opened 2021-06-24
Various fixes

arkivm:
* Minor fixes to doc/workshop
* Fix trailing semicolon warning in compiler


doc/datalog2.0-workshop/.gitignore
```diff
@@ -2,4 +2,4 @@
 *.blg
 *.log
 *.aux
-top.pdf
+paper.pdf
```
mbudiu-vmw, line 5: This has been published already.

arkivm: It is mainly here to avoid showing the paper.pdf as modified every time the latex sources are built. So I thought, we can only commit paper.pdf if need be with `git add -f`. It's a little ugly, I agree. I can remove paper.pdf from gitignore, but we should be a bit cautious when rebuilding the latex sources (to not add paper.pdf without any diffs).

Issue #1008, opened 2021-06-24
Richer metadata in `XFormArrangement`, `XFormCollection`.


ryzhyk:
`XFormArrangement` and `XFormCollection`, as well as `Relation` and `Arrangement` should contain a field pointing to the source code fragment(s).  This information can be used to generate more informative profiles.

cc @roymanish

Kixiron:
I'm working on this rn

PR #1009, opened 2021-06-24
More relation & arrangement names

Kixiron:
Adds source locations to joins and arrangements

See #1008

### DIFF
--- a/src/Language/DifferentialDatalog/Rule.hs
+++ b/src/Language/DifferentialDatalog/Rule.hs
@@ -66,16 +66,19 @@ import Language.DifferentialDatalog.DatalogProgram
 -- resulting rule may no longer type check and should not be used for
 -- anything other then pretty-printing).
 ruleStripTypeAnnotations :: Rule -> Rule
-ruleStripTypeAnnotations rl@Rule{..} =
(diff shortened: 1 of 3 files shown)


src/Language/DifferentialDatalog/Rule.hs
```diff
@@ -66,16 +66,19 @@ import Language.DifferentialDatalog.DatalogProgram
...
-rulePPPrefix rl len = commaSep $ map pp $ take len $ ruleRHS $ ruleStripTypeAnnotations rl
+rulePPPrefix rule len = commaSep (map pp $ take len $ ruleRHS $ ruleStripTypeAnnotations rule) <+> location rule
 
 -- | Pretty-print rule without type annotations.
 rulePPStripped :: Rule -> Doc
-rulePPStripped rule = pp (ruleStripTypeAnnotations rule) <+> char '@' <+> file <> char ':' <> line <> char ':' <> column
+rulePPStripped rule = pp (ruleStripTypeAnnotations rule) <+> location rule
+
+location :: Rule -> Doc
+location rule = char '@' <+> file <> char ':' <> line <> char ':' <> column
```
ryzhyk, line 81: I'm still not sure what's wrong with printing the whole range. This way tools would be able to automatically extract the entire rule.

PR #1010, opened 2021-06-26
Major refactor: Move HDDlog to the differential_datalog crate

Kixiron:
Moves the `HDDlog` struct from the generated code into the main `differential_datalog` crate

Note: This is a very rough PR right now, I've only fixed up the immediate compile errors and none of the generated code or tests yet


ryzhyk:
This is great.  I started working on this some time ago, but abandoned the project due to the lack of time.  Would you like me to take a look at the patches now or shall I wait till you're done fixing things up?

Kixiron:
The code I've already written shouldn't change that much, I've refactored the rust side of things as of now and just need to update code generation to match it

test/datalog_tests/rust_api_test/src/main.rs
```diff
@@ -43,7 +52,15 @@ fn main() -> Result<(), String> {
     //   of output relations.  Should only be used for debugging in order to dump
     //   the contents of output tables using `HDDlog::dump_table()`.  Otherwise,
     //   indexes are the preferred way to achieve this.
-    let (hddlog, init_state) = HDDlog::run(1, false)?;
+    let (hddlog, init_state) = HDDlog::new(
```
ryzhyk, line 55: We should generate a convenience function for this.  There's no way we can expose this API to users.

PR #1012, opened 2021-06-29
Correct documentation around fully qualified names

amytai:
Module names are delimited by `::` rather than `.`.

I'm not sure if there are other places in actual documentation where this needs to be updated, but if so, we should add to this PR.


ryzhyk:
Good catch, thanks!

Kixiron (approved):

PR #1013, opened 2021-06-30
graph: Document the order of output tuples for connected components.

blp:
I made a wrong assumption and this would have save some debugging.


ryzhyk:
Thanks!

PR #1015, opened 2021-07-02
[WIP] adding 'source' feature to CLI

smadaminov:
This PR is about adding a 'source' feature to the CLI. It is still work-in-progress as there are some tests, which are not passed and a bit more need to be done on the code side. But it should provide an idea of the way I was thinking about implementing 'source' feature.

It is built around of replacing single loop with two separate async tasks where there is one parsing task and other one is a provider of commands. Such division potentially helps to add more command sources such as remotely connected user.

It may also ease adding 'display' feature, which can print changes on every commit transactions.

There is more work need to be done if implement in this way but before investing more time, I would like to get some feedback if possible.

### DIFF
--- a/rust/template/cmd_parser/lib.rs
+++ b/rust/template/cmd_parser/lib.rs
@@ -2,9 +2,8 @@
 
 mod parse;
 
-use std::io;
-use std::io::BufRead;
-use std::io::BufReader;
+use std::fs::{ self, File };
+use std::io::{ self, BufRead, BufReader };
 
 pub use parse::*;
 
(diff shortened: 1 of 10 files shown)


ryzhyk:
Thanks, @smadaminov!  

- I'm not sure I'm completely onboard with the async re-design.  It adds a bunch of complexity, but doesn't seem to improve functionality or performance. Would you be able to elaborate on how this will simplify adding new types of input in the future?

- Also, I am curious where you're going with the namespace stack feature.  I am guessing, this is meant to allow using local names rather than fully qualified names in CLI.  Will this require some new syntax or do you plan that the `source` command will have a side effect of switching module scopes?

- Regarding the TODO list in the source code, e.g.,
  `"start" => check if first start or not, and give warning that nested transactions currently not supported`.
  We already implement this check internally in HDDlog.  Is the purpose to impose additional constraints on the `source` command, e.g., that it shouldn't break up transactions in the outer file?  If we take that approach instead of interpreting `source` simply  as "paste the contents of the imported file here", then what is the exact set of rules you think we should enforce?  I am personally in favor of the least restrictive design, so that `source` can be used in creative ways to build command scripts.

Will add more comments inline.

rust/template/cmd_parser/lib.rs
```diff
@@ -2,31 +2,55 @@
...
+    TempStore,
+}
+
+#[derive(Debug)]
+struct CliCmdReq {
+    cli_cmd: Vec<u8>,
+    resp: Option<oneshot::Sender<Result<bool, String>>>,
+}
+
+fn send_cmd_to_parser(tx: &mpsc::Sender<CliCmdReq>, cmd: CliCmdReq) -> Result<bool, String> {
```
ryzhyk, line 32: I think a better name would be `send_cmd_to_ddlog` or `send_command_to_executor`, since the receiver both parses and executes the command.

rust/template/cmd_parser/lib.rs
```diff
@@ -39,88 +63,290 @@ where
...
+            let mut cmd_tokens = line.split_whitespace();
+            let first_cmd = match cmd_tokens.next() {
+                Some(cmd) => cmd.replace(";", ""),
+                None => { continue; },
+            };
+
+            match first_cmd.as_str() {
+                // TODO: what to do if source fails? partially executed OK, or do we want complete rollback?
+                "source" => {
+                    let filename = match cmd_tokens.next() {
```
ryzhyk, line 187: This won't work for filenames containing whitespaces.

rust/template/cmd_parser/lib.rs
```diff
@@ -39,88 +63,290 @@ where
...
+                        Some(filename) => filename.replace(";", ""),
+                        None => {
+                            eprintln!("Please provide source filename");
+                            eprintln!("Usage: source <filename>");
+                            continue;
```
ryzhyk, line 192: We probably want to fail and exit here when receiving commands from a pipe (but not in the TTY mode).

rust/template/cmd_parser/lib.rs
```diff
@@ -39,88 +63,290 @@ where
...
+                        },
+                    };
+                    
+                    if let Ok(f) = File::open(filename) {
+                        sourcing = true;
+                        original_input = input;
```
ryzhyk, line 198: I know this is work in progress, but I guess the plan is to maintain a stack of inputs?  How is that related to the namespace stack?

ryzhyk:
@smadaminov , closing this for now.  Feel free to reopen whenever ready.

PR #1016, opened 2021-07-02
DDShow integration

ryzhyk:

Still working on:
- tests
- docs
- Java bindings.

This breaks backwards compatibility by not enabling profiling by default.  The user must explicitly choose profiling mode (self-profiling or ddshow) whether starting a program via API or command line.

CLI options to configure profiling are kind of complicated, but simply specifying `--ddshow` (and `--differential` to profile arrangement size) should be sufficient for most purposes.


rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -1026,22 +1023,26 @@ impl Program {
...
                     profiling_data.clone(),
                     Arc::clone(&request_recv),
                     Arc::clone(&reply_send),
                 );
 
-                worker.run()
+                worker.run().map_err(|e| {
+                    eprintln!("Worker thread failed: {}", e);
+                    e
+                })
```
Kixiron, line 1045: This will make things fail if `TIMELY_WORKER_ADDR` happens to be set in the environment since timely sets some stuff up by default, I ran into the same issue in ddshow and [this was my workaround](https://github.com/Kixiron/ddshow/blob/master/src/main.rs#L70-L82). As a plus doing this also makes the stuff [here](https://github.com/vmware/differential-datalog/blob/43f7d2c34dee916ff5a9ccb0bcdd540723f82e3b/rust/template/differential_datalog/src/program/worker.rs#L524-L527) redundant

rust/template/differential_datalog/src/program/worker.rs
```diff
@@ -451,33 +455,82 @@ impl<'a> DDlogWorker<'a> {
                     }
                 },
             );
-        } else if let ProfilingKind::TimelyProfiling {
-            differential_dataflow,
-        } = self.config.profiling_kind
+        } else if let ProfilingConfig::TimelyProfiling {
+            timely_destination,
+            timely_progress_destination: _,
+            differential_destination,
```
Kixiron, line 461:
```suggestion
            differential_destination,
            ..
```

Kixiron (approved):
Looks great, thanks!

rust/template/differential_datalog/src/program/mod.rs
ryzhyk: Ah, glad you caught this.  So the solution is to use `execute_from` instead of `execute`?

ryzhyk:
I probably shouldn't merge this until I've added Java bindings; otherwise there's no way right now to enable profiling from Java (same for Go, but I don't think we have many go users).

rust/template/src/main.rs
```diff
@@ -244,31 +252,223 @@ fn main() -> Result<(), String> {
         opt init_snapshot:bool=true, desc:"Do not dump initial output snapshot.";                                                   // --no-init-snapshot
         opt print:bool=true, desc:"Backwards compatibility. The value of this flag is ignored.";                                    // --no-print
         opt workers:usize=1, short:'w', desc:"The number of worker threads. Default is 1.";                                         // --workers or -w
+        opt idle_merge_effort:Option<isize>, desc:"Set Differential Dataflow's 'idle_merge_effort' parameter.This flag takes precedence over the '$DIFFERENTIAL_EAGER_MERGE' environment variable.";
```
mbudiu-vmw, line 255: missing space after dot

rust/template/src/main.rs
```diff
@@ -244,31 +252,223 @@ fn main() -> Result<(), String> {
...
+        opt profile_timely:bool=false, desc:"Use external Timely Dataflow profiler.";
+        opt profile_differential:bool=false, desc:"Use external Differential Dataflow profiler. Implies '--profile-timely'";
+        opt self_profiler:bool, desc:"Enable DDlog internal profiler. This option is mutually exclusive with '--profile-timely'.";
+        opt timely_profiler_socket:Option<String>, desc:"Socket address to send Timely Dataflow profiling events. Default (if '--profile-timely' is specified) is '127.0.0.1:51317'. Implies '--profile-timely'.";
+        opt timely_trace_dir:Option<String>, desc:"Path to a directory to store Timely Dataflow profiling events, e.g., './timely_trace'. Implies '--profile-timely'.";
+        opt differential_profiler_socket:Option<String>, desc:"Socket address to send Differential Dataflow profiling events. Default (if '--profile-differential' is specified is '127.0.0.1:51318'. Implies '--profile-differential'.";
```
mbudiu-vmw, line 261: So I guess you cannot use both a file and a socket

mbudiu-vmw (approved):

rust/template/src/main.rs
ryzhyk: Nope, the ddshow client-side library doesn't support that.

rust/template/differential_datalog/src/program/mod.rs
Kixiron: Yep, it's a little annoying since you have to do stuff manually but that's the gist

ryzhyk:
@Kixiron , would you be able to look at the last two commits (profiling tutorial & new benchmark)? Thanks!

java/ddlogapi.c
```diff
@@ -192,6 +192,96 @@ JNIEXPORT jlong JNICALL Java_ddlogapi_DDlogAPI_ddlog_1run(
     return handle;
 }
 
+ddlog_log_destination log_config(JNIEnv *env, jint mode, jstring addr_str) {
+    switch (mode) {
+        // socket
```
convolvatron, line 197: idk anything about the context here, but in an ideal world these would be some kidn of enummeration {socket, disk, ...}

java/ddlogapi.c
```diff
@@ -192,6 +192,96 @@ JNIEXPORT jlong JNICALL Java_ddlogapi_DDlogAPI_ddlog_1run(
...
+    jint differentialLogMode, jstring differentialLogAddr) {
+
+    jlong handle = 0;
+
+    if (workers <= 0)
+        workers = 1;
+
+    ddlog_profiling_config profiling_config;
+
+    switch (profilingMode) {
```
convolvatron, line 235: ditto on the enum

rust/ddlog_benches/benches/twitter.rs
```diff
@@ -45,6 +49,52 @@ fn twitter_micro(c: &mut Criterion) {
...
+                        || {
+                            (
+                                twitter::init_config(Config {
+                                    num_timely_workers: thread_count,
+                                    profiling_config: ProfilingConfig::TimelyProfiling {
+                                        timely_destination: LoggingDestination::Disk {
+                                            directory: "timely_trace".to_string(),
+                                        },
+                                        timely_progress_destination: None,
+                                        differential_destination: None,
```
Kixiron, line 82: Probably should add one that includes ddflow

java/ddlogapi/DDlogConfig.java
```diff
@@ -0,0 +1,214 @@
...
+        this.enableDebugRegions = enableDebugRegions;
+        this.differentialIdleMergeEffort = differentialIdleMergeEffort;
+        this.profilingConfig = profilingConfig;
+    }
+
+    /**
+     * Logging mode for timely and differential profiling event streams.
+     * WARNING: The following constants are hardcoded in ddlogapi.c
+     */
+    public enum LogMode {
```
convolvatron, line 37: enum again - i know you're patching that here, but they should really be that way at the source (?)

README.md
```diff
@@ -78,7 +78,8 @@ but somewhat limits the flexibility, as changes to the relational schema or rule
 - [How to](doc/java_api.md) use DDlog from Java.
 - [How to](doc/c_tutorial/c_tutorial.rst) use DDlog from C.
 - [How to](go/README.md) use DDlog from Go and [Go API documentation](https://pkg.go.dev/github.com/vmware/differential-datalog/go/pkg/ddlog).
-- [How to](test/datalog_tests/rust_api_test) using DDlog from Rust (by example)
+- [How to](test/datalog_tests/rust_api_test) use DDlog from Rust (by example)
+- [Tutorial](doc/profiling.md) on profiling DDlog programs
```
Kixiron, line 82: I don't see `profiling.md` anywhere, have you committed it yet?

convolvatron (approved):
I can't help that some of this could just be passed through instead of picked up and translated and constructed, and descontructed. but thats really the whole point here - glue. so if there were a less-gluier path i'm sure you would have chosen it. 

I guess I would have thought about using something more fact or property-list like for the config - would certainly cut down on boilerplate, but I guess that's the java way :)

rust/template/src/main.rs
Kixiron: It does/could, you'd just need to make a tee'd writer a la
```rust
struct Tee<W1, W2> {
    first: W1,
    second: W2,
}

impl<W1, W2> Write for Tee<W1, W2>
where
    W1: Write,
    W2: Write,
{ ... }

ddshow_sink::enable_timely_logging(Tee::new(File::open(..), TcpStream::connect(..)));
```

java/ddlogapi.c
Kixiron: Yah, this does look like something that should be handled by `cbindgen`, but that's probably out of scope for this PR

rust/ddlog_benches/benches/twitter.rs
ryzhyk: Good point.  differential logs are usually much smaller, so I'll just enable both.

README.md
ryzhyk: Sorry, added it and pushed.

java/ddlogapi.c
ryzhyk: This is a Java-to-C bridge, not C-to-Rust, so `cbindgen` wouldn't help.

ryzhyk: @convolvatron , good point,  I can't use Java enum's from C (without writing even more ugly JNI code), but I have C enums I could use here, and in the other location you pointed out.

doc/profiling.md
```diff
@@ -0,0 +1,205 @@
...
+
+DDlog's profiling features are designed to help the programmer identify
+parts of the program that use the most CPU and memory.  DDlog supports
+two profilers:
+
+- DDlog's **self-profiler** runs as a thread inside the DDlog program and generates
+  textual memory and CPU profiles that can be output at any point
+  during the runtime of the program.
+
+- [**DDshow**](https://github.com/Kixiron/ddshow) is a stadalone profiler that
```
Kixiron, line 13:
```suggestion
- [**DDShow**](https://github.com/Kixiron/ddshow) is a standalone profiler that
```

Kixiron (approved):
Looks good, thanks!

doc/profiling.md
ryzhyk: Yeah, it's DDlog, but DDShow :)

PR #1018, opened 2021-07-05
Added Clone bounds to IntoRecord and Mutator derives

### LINKED ISSUE #1017: How to add trait bound on generic type?
The compiler complains that `N` needs to be restricted by trait bound `Clone` when I'm calling the Rust APIs. Is there any way I can add the trait bound on the generic type in ddlog program or I have to change it to an extern type in the native Rust constructor with whatever trait bounds I want to add?
```
typedef List<'N> = List { node: 'N, nxt: ListNxt<'N> } | EMPTY
typedef ListNxt<'N> = ListNxt { nxt: Ref<List<'N>> }
```

Below is the generated code from ddlog program that does not enforce `Clone` on the generic type `N`.
```Rust
#[derive(Eq, Ord, Clone, Hash, PartialEq, PartialOrd, IntoRecord, Mutator, Serialize, Deserialize, FromRecord)]
#[ddlog(rename = "List")]
pub enum List<N> {
    #[ddlog(rename = "List")]
    List {
        node: N,
        nxt: ListNxt<N>
    },
    #[ddlog(rename = "EMPTY")]
    EMPTY
}
impl <N: ::ddlog_rt::Val> abomonation::Abomonation for List<N>{}
```

Kixiron:
* Adds `Clone` bounds into the derived `IntoRecord` and `Mutator` derives
  * Adds a regression test to the derive suite for #1017
* Implements `IntoRecord`, `FromRecord` and `Mutator` for `std::sync::Arc`, `triomphe::Arc`, `std::rc::Rc`, `std::cell::Cell` and `std::cell::RecCell`

Fixes #1017

### DIFF
--- a/rust/template/differential_datalog/src/record/mod.rs
+++ b/rust/template/differential_datalog/src/record/mod.rs
@@ -8,16 +8,21 @@ use ordered_float::OrderedFloat;
 use serde::{Deserialize, Serialize};
 use std::{
     borrow::Cow,
+    cell::{Cell, RefCell},
     collections::{btree_map, BTreeMap, BTreeSet},
     fmt::{self, Write},
     iter::FromIterator,
-    str, vec,
+    rc::Rc,
+    str,
+    sync::Arc as StdArc,
+    vec,
 };
 #[cfg(feature = "c_api")]
 use std::{
     ffi::{CStr, CString},
     ptr, slice,
 };
+use triomphe::Arc;
 
 pub type Name = Cow<'static, str>;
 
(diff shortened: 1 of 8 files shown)


rust/template/differential_datalog/src/record/mod.rs
```diff
@@ -826,6 +831,170 @@ impl<T: FromRecord + Ord> Mutator<BTreeSet<T>> for Record {
     }
 }
 
+impl<T> FromRecord for Arc<T>
```
ryzhyk, line 834: I was surprised to see these impls given that we already have all the traits implemented for `Ref<>` in `ddlog_std.rs`, but I guess there's no harm in having them: if someone wants to construct extern types using `Arc` and friends, they will be able to take advantage of these existing impls.

ryzhyk (approved):
Looks good, thanks! I am going to add the list example that motivated this fix to one of our tests before merging.

Kixiron:
I added a regression test if that's what you mean

PR #1021, opened 2021-07-16
Several unrelated improvements.

ryzhyk:
Fixed some regressions, including a nasty performance bug, and added a couple of features. I want to publish this as a minor release 0.42.1.

### DIFF
--- a/lib/ddlog_std.dl
+++ b/lib/ddlog_std.dl
@@ -752,6 +752,18 @@ function keys(m: Map<'K, 'V>): Vec<'K> {
     map_keys(m)
 }
 
+function values(m: Map<'K, 'V>): Vec<'V> {
+    map_values(m)
+}
+
+function nth_key(m: Map<'K, 'V>, n: usize): Option<'K> {
+    map_nth_key(m, n)
+}
+
+function nth_value(m: Map<'K, 'V>, n: usize): Option<'V> {
+    map_nth_value(m, n)
+}
+
 /*
  * Set
  */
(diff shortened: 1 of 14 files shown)


lib/ddlog_std.dl
```diff
@@ -752,6 +752,14 @@ function keys(m: Map<'K, 'V>): Vec<'K> {
     map_keys(m)
 }
 
+function values(m: Map<'K, 'V>): Vec<'V> {
+    map_values(m)
+}
+
+function nth(m: Map<'K, 'V>, n: usize): Option<'V> {
```
mbudiu-vmw, line 759: Why not nth_value?

mbudiu-vmw (approved):

lib/ddlog_std.rs
```diff
@@ -997,6 +997,16 @@ pub fn map_keys<K: Ord + Clone, V>(map: &Map<K, V>) -> Vec<K> {
...
 }
 
+pub fn map_values<K: Ord, V: Clone>(map: &Map<K, V>) -> Vec<V> {
+    Vec {
+        vec: map.x.values().cloned().collect(),
+    }
+}
+
+pub fn map_nth<K: Ord, V: Clone>(m: &Map<K, V>, n: &std_usize) -> Option<V> {
+    option2std(m.x.iter().nth(*n as usize).map(|(_, v)| v.clone()))
```
Kixiron, line 1007: I'd be wary of this function since the order of values in a map is non-deterministic

ryzhyk: We can't have maps with nondet ordering in ddlog because we allow iterating over maps. This is a btreemap.

lib/ddlog_std.dl
ryzhyk: Yeah, `nth_value` is less ambiguous.  I'll do that.

PR #1023, opened 2021-07-19
Remove old d3log

arkivm:
Remove old
* implementation of d3log
* test cases
* test cases from gitlab-CI

This PR is to make way for the new d3log implementation.


mbudiu-vmw:
Can't we keep it around and put the new one on the side until it provides at least equivalent functionality?

ryzhyk:
I support deleting it to reduce maintenance overheads. And I think the new version that will be merged soon is actually more functional.

test.sh
```diff
@@ -325,36 +324,6 @@ souffle_tests7() {
 
 # 'd3log' test group.
 
-tcp_channel() {
```
ryzhyk, line 328: You also want to remove stuff at line 107 (github doesn't let me annotate lines that haven't changed)

ryzhyk (approved):
Looks good, thanks!

ryzhyk:
I'd also suggest merging the three commits in this PR into one.

test.sh
arkivm: fixed. thanks

PR #1028, opened 2021-07-23
D3log merge

convolvatron:
Initial merge of d3log machinery (excluding unix process instances and tcp ports for sake of review brevity)- to be brought back in later.

This commit establishes a basic runtime to support @ locality relations, provides support for DDValue and Record based facts, and implements a bus to broadcast management facts to all instances. 

more examples and a better readme are really required


rust/template/d3log/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+            RecordBatch::from(self.eval.clone(), b.clone())
+        );
+        self.dispatch.send(b.clone());
+
+        {
+            self.queue.lock().expect("lock").push_back(b.clone());
+        }
+
+        loop {
+            let b = {
```
convolvatron, line 109: this scheduling has already shows itself to be misguided when running with multiple threads. will update.

convolvatron: unsurprisingly, using a proper mpsc here works just fine...will fix here

rust/template/d3log/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+        }),
+    )?;
+
+    broadcast.clone().subscribe(eval_port.clone());
+    broadcast.clone().subscribe(Arc::new(AccumulatePort {
+        eval: eval.clone(),
+        b: accu_batch.clone(),
+    }));
+    eval_port.send(fact!(d3_application::Myself, me => uuid.into_record()));
+    Ok((broadcast, init_batch, eval_port, dispatch, forwarder))
```
convolvatron, line 227: I'm going to go ahead and make an Instance structure so that we don't have to carry around the 8 or 9 objects explicitly. these structures tend to collect junk, but its just a lot of noise right now. plus some of the giant blocks of clones can just become Instance.clone()

rust/template/src/d3main.rs
```diff
@@ -0,0 +1,270 @@
...
+
+    fn deserialize_batch(&self, s: Vec<u8>) -> Result<DDValueBatch, Error> {
+        let s = std::str::from_utf8(&s)?;
+        let v: SerializeBatchWrapper = serde_json::from_str(&s)?;
+        Ok(v.b)
+    }
+}
+
+pub fn start_d3log() -> Result<(), Error> {
+    let (uuid, is_parent) = if let Some(uuid) = std::env::var_os("uuid") {
```
ryzhyk, line 225: We discussed that this is a temporary hack.  I propose the following solution for the startup protocol:

1. Ignore initial outputs.  I don't think they are important and I don't think we should rely on them for anything in D3log.  In fact they may disappear from DDlog altogether at some point.

2. `d3_application.dl` should include an input relation (`D3Config`?) that defines its startup configuration.  The runtime populates this relation with a single record in the very first transaction and never touches is again (and the application can rely on this behavior).  The record consists of at least `parent: Option<Node>`.  `parent = None` indicates that we are the root of the local hierarchy. Additionally it can contain arbitrary state passed to it by parent, e.g., `argv: Vec<string>`.

One advantage of this design is that we can do more things, like advertising the Stdout port relationally.

rust/template/d3log/src/dispatch.rs
```diff
@@ -0,0 +1,66 @@
...
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
+
+type DispatchMap = HashMap<String, Vec<(u64, Port)>>;
+
+#[derive(Clone)]
+pub struct Dispatch {
+    eval: Evaluator, // to be removed
+    count: Arc<AtomicUsize>,
+    handlers: Arc<Mutex<DispatchMap>>,
```
ryzhyk, line 17: Do we really need a `Mutex` as opposed to a `RwLock` here (and possibly in other places in the code)?

rust/template/d3log/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    error::Error,
+    forwarder::Forwarder,
+    record_batch::RecordBatch,
+};
+
+pub type Node = D3logLocationId;
+
+pub trait EvaluatorTrait {
+    fn ddvalue_from_record(&self, id: String, r: Record) -> Result<DDValue, Error>;
+    fn eval(&self, input: Batch) -> Result<Batch, Error>;
```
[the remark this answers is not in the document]
ryzhyk: This will also simplify the implementation of `Dispatch`.

rust/template/d3log/src/broadcast.rs
```diff
@@ -0,0 +1,74 @@
...
+
+pub struct Ingress {
+    index: usize,
+    broadcast: Arc<Broadcast>,
+}
+
+#[derive(Clone, Default)]
+pub struct Broadcast {
+    count: Arc<AtomicUsize>,
+    ports: Arc<Mutex<Vec<(Port, usize)>>>,
```
ryzhyk, line 19: `RwLock`?

rust/template/d3log/src/broadcast.rs
```diff
@@ -0,0 +1,74 @@
...
+// fix for simple pairwise loops. This will need an additional distributed coordination mechanism in
+// order to maintain a consistent spanning tree (and a strategy for avoiding storms for temporariliy
+// inconsistent topologies
+
+use crate::{Batch, Port, Transport};
+
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Mutex};
+
+pub struct Ingress {
```
ryzhyk, line 11: If I understand correctly, this is needed so that broadcast group members can send messages to all group members excluding self.  Please document this here (and also in `trait PubSub`) for ease of understanding.

rust/template/d3log/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+    broadcast: Arc<Broadcast>,
+    // since all kinds of children will need a copy of the management state,
+    // consider generalizing
+    accumulator: Arc<Mutex<DDValueBatch>>,
+}
+
+// we're just throwing this into the same runtime - do we want/need scheduling isolation?
+// xxx handle deletes
+impl Transport for ThreadInstance {
+    fn send(&self, b: Batch) {
```
ryzhyk, line 144: This is basically `fork`, right?  Let's discuss this face-to-face, I'm struggling to form a big picture here.

ryzhyk (commented):
First-pass review.

rust/template/d3log/README
```diff
@@ -0,0 +1,101 @@
...
+Using D3log:
+----------------
+
+ Instances:
+ ----------------
+ D3log adds an additional primitive type -D3LogLocationID - to represent an
+ instance in the system. Each instance is a separate DDlog evaluator, and we
+ provide an extension to allow instances to exchange relational data.
+
+ Instances can be passed in from external systems as facts in a user defined
```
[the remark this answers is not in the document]
convolvatron: yes.

rust/template/d3log/README
```diff
@@ -0,0 +1,101 @@
...
+
+        [show an example of how the differential machinery updates the distribution]
+
+
+ Compiling D3log programs:
+ ----------------
+
+ to compile a ddlog program with the d3log localization extension, use the ddlog flag -d3log
+
+ to compile the generated rust program, 'cargo build --features distribution' in the generated directory.
```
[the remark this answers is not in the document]
convolvatron: yeah, we talked about this a little and didn't come up with this option. that would be better.

rust/template/d3log/README
```diff
@@ -0,0 +1,101 @@
...
+  correct approach
+
+  json_framer is a terrible idea. however, i'd like to do async i/o on
+  streams of self-framing json objects. serde supports self framed
+  objects (StreamDeserializer), but since it isn't in async land, it
+  doesn't look possible to implement io:Read without burning a thread to
+  block.
+
+  is it important that we implement all the methods in
+  differential_datalog/src/program/update.rs (DeleteKey, etc)?
```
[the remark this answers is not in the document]
convolvatron: thanks!

PR #1033, opened 2021-07-30
Update nproc to be Darwin-compatible

amytai:


mbudiu-vmw (approved):
This is the only change?

ryzhyk:
> This is the only change?

I don't think this is meant to fix the flatbuf compilation problem, just the `make` command.

PR #1035, opened 2021-08-02
Strip array subtype in Presto -> H2 compiler

amytai:
H2 SQL doesn't like subtypes for the ARRAY column type.
In order for the JooqConnection to accept the result of this translator, we need to strip subtypes from Presto ARRAY(*) column types.

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/TranslateCreateTableDialect.java
+++ b/sql/src/main/java/com/vmware/ddlog/TranslateCreateTableDialect.java
@@ -13,6 +13,8 @@ import com.facebook.presto.sql.tree.TableElement;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 class TranslateCreateTableDialect extends AstVisitor<String, String> {
     @Override
@@ -22,6 +24,15 @@ class TranslateCreateTableDialect extends AstVisitor<String, String> {
         for (final TableElement element : node.getElements()) {
             if (element instanceof ColumnDefinition) {
                 final ColumnDefinition cd = (ColumnDefinition) element;
+                String h2Type = cd.getType();
+
+                // We need to strip the array subtype for H2
+                final Pattern arrayType = Pattern.compile("ARRAY\\((.+)\\)");
+                Matcher m = arrayType.matcher(cd.getType());
+                if (m.find()) {
+                    h2Type = "array";
+                }
+
                 if (cd.getProperties().size() == 1) {
                     final String propertyName = cd.getProperties().get(0).getName().getValue();
                     final String propertyValue = cd.getProperties().get(0).getValue().toString();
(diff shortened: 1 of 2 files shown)


sql/src/main/java/com/vmware/ddlog/TranslateCreateTableDialect.java
```diff
@@ -22,6 +24,15 @@ protected String visitCreateTable(final CreateTable node, final String sql) {
         for (final TableElement element : node.getElements()) {
             if (element instanceof ColumnDefinition) {
                 final ColumnDefinition cd = (ColumnDefinition) element;
+                String h2Type = cd.getType();
+
+                // We need to strip the array subtype for H2
+                final Pattern arrayType = Pattern.compile("ARRAY\\((.+)\\)");
+                Matcher m = arrayType.matcher(cd.getType());
```
mbudiu-vmw, line 31: you already have h2Type, you should use it here.

sql/src/main/java/com/vmware/ddlog/TranslateCreateTableDialect.java
```diff
@@ -22,6 +24,15 @@ protected String visitCreateTable(final CreateTable node, final String sql) {
...
+                if (m.find()) {
+                    h2Type = "array";
```
mbudiu-vmw, line 33: So there is such a type "array"?
This seems wrong to me.

mbudiu-vmw (approved):
I am approving, but it would be nice to have a more principled approach.
One of these software components is braindead. Which one is it?

sql/src/main/java/com/vmware/ddlog/TranslateCreateTableDialect.java
amytai: fixed

sql/src/main/java/com/vmware/ddlog/TranslateCreateTableDialect.java
amytai: What's your concern? We can discuss before merging.

Neither H2 nor Presto have very well-specified SQL languages, but it appears H2 has a simple "array" type which [accepts no type](http://www.h2database.com/html/datatypes.html#array_type).

mbudiu-vmw: So perhaps we should add a configuration option for the code generator, indicating the expected SQL back-end.
When we compile for H2 we strip, otherwise we don't.

amytai: Ah yes, that will be in the next PR. I think I need that for the JooqProvider, because we are doing a silly thing right now where we (manually) convert SQL from H2 to Presto dialect, which gets converted back to H2 inside this ddlog-sql code. Whereas you should just have a flag for whether the current connection is using H2 or Presto, and translate accordingly.

However, I don't necessarily think you need a flag for this particular class. My understanding is that this particular class explicitly assumes it is translating from Presto --> H2.

mbudiu-vmw (approved):

sql/src/main/java/com/vmware/ddlog/TranslateCreateTableDialect.java
mbudiu-vmw: I don't remember any longer, but if that's true, then this is fine.

Issue #1036, opened 2021-08-03
ddlog-sql: NullPointerException when trying to insert null for varchar column value


amytai:
Repro diff on DDlog commit 0ccf6394
```
diff --git a/sql/src/test/java/ddlog/JooqProviderTest.java b/sql/src/test/java/ddlog/JooqProviderTest.java
index 300490f2..511831f9 100644
--- a/sql/src/test/java/ddlog/JooqProviderTest.java
+++ b/sql/src/test/java/ddlog/JooqProviderTest.java
@@ -64,7 +64,7 @@ public class JooqProviderTest {

     @BeforeClass
     public static void setup() throws IOException, DDlogException {
-        String s1 = "create table hosts (id varchar(36) with (primary_key = true), capacity integer, up boolean)";
+        String s1 = "create table hosts (id varchar(36) with (primary_key = true), capacity integer, up boolean, tester varchar(36))";
         String v2 = "create view hostsv as select distinct * from hosts";
         String v1 = "create view good_hosts as select distinct * from hosts where capacity < 10";
         String checkArrayParse = "create table junk (testCol integer array)";
@@ -136,10 +136,10 @@ public class JooqProviderTest {
     public void testSqlOpsWithBindings() {
         // Insert statements.
         create.insertInto(table("hosts"))
-              .values("n1", 10, true)
+              .values("n1", 10, true, null)
               .execute();
-        create.batch(create.insertInto(table("hosts")).values("n54", 18, false),
-                     create.insertInto(table("hosts")).values("n9", 2, true))
+        create.batch(create.insertInto(table("hosts")).values("n54", 18, false, null),
+                     create.insertInto(table("hosts")).values("n9", 2, true, null))
               .execute();
```

Partial stack trace of error:
```
java.lang.NullPointerException
	at org.jooq_3.13.1.DEFAULT.debug(Unknown Source)
	at ddlogapi.DDlogRecord.checkNull(DDlogRecord.java:26)
	at ddlogapi.DDlogRecord.<init>(DDlogRecord.java:95)
	at com.vmware.ddlog.DDlogJooqProvider.toValue(DDlogJooqProvider.java:511)
	at com.vmware.ddlog.DDlogJooqProvider.access$700(DDlogJooqProvider.java:86)
	at com.vmware.ddlog.DDlogJooqProvider$QueryVisitor.visitInsert(DDlogJooqProvider.java:266)
[...]

PR #1037, opened 2021-08-04
ddlog-sql: `not null` column annotation gets lost dialect translation

amytai:
`not null` column annotation gets lost in Presto -> H2 translation. This causes any insertions into a column with `not null` to fail on the DDlog side, because the types are not the same (DDlog expects a type T without an option, but the JooqProvider will end up generating a DDlog `Option<T>` object for a record).

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/TranslateCreateTableDialect.java
+++ b/sql/src/main/java/com/vmware/ddlog/TranslateCreateTableDialect.java
@@ -47,7 +47,7 @@ class TranslateCreateTableDialect extends AstVisitor<String, String> {
                     throw new RuntimeException(String.format("Unsupported properties %s in sql: %s",
                             cd.getProperties(), sql));
                 }
-                columnsToCreate.add(String.format("%s %s", cd.getName().getValue(), h2Type));
+                columnsToCreate.add(String.format("%s %s %s", cd.getName().getValue(), h2Type, cd.isNullable() ? "" : "not null"));
             }
         }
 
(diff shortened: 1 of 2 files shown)


mbudiu-vmw (approved):
What happens (and should happen) if you try to insert a null value?
Do you get the expected error?

amytai:
The new test (`testNotNullColumns`) fails with the following error:
```
org.jooq.exception.DataAccessException: SQL [insert into not_null values (5, 'test_string')]; ddlogapi.DDlogException: ddlog_apply_updates(): error: invalid command Insert(RelId(4), PosStruct("Tnot_null", [PosStruct("ddlog_std::Some", [Int(BigInt { sign: Plus, data: BigUint { data: [5] } })]), PosStruct("ddlog_std::Some", [String("test_string")])])): not an int PosStruct("ddlog_std::Some", [Int(BigInt { sign: Plus, data: BigUint { data: [5] } })])
```

which is the error I described in the PR body (i.e. insertion path is creating an Option<T> field in a DDlog record, when DDlog expects T field).

mbudiu-vmw:
So is this error acceptable? This is not easy to parse. DDlog fails to construct a value that contains a null at all for this row.

amytai:
Oh sorry, I misread your first comment.

If I try to insert a null into this code, I get the exception described in #1036 1036

ryzhyk:
@amytai , @mbudiu-vmw, can we add a test for this?

mbudiu-vmw:
I am pretty sure this PR includes a test. It's an extension of an existing test.

PR #1039, opened 2021-08-09
Convert string literals to lazy statics.

ryzhyk:
Since we map DDlog's `string` type to Rust's `String`, the
straightforward translation of string literals instantiates a `String`
every time the literal is evaluated: `String::from("foo")`.  This
involves a dynamic memory allocation, which is highly inefficient.  The
right thing to do in this case is to use `&'static str` and insert
conversions to/from `String` only where necessary.  Unfortunately, this
requires a significant compiler refactoring: we'd have to introduce the
notion that `T` and `&T` can be represented as different Rust types and
inject type conversions where necessary.

We implement a much simpler but slightly less efficient optimization that
simply converts string literals into lazy statics, so that memory
allocation is only performed once.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Optimizations
+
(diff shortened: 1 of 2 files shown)


mbudiu-vmw (approved):
Do you want to file an issue for the full optimization?

Kixiron (approved):
It's not ideal but it's good for a patch fix, we can cook up something more thorough later

PR #1048, opened 2021-08-16
Eliminate redundant string concatenation.

ryzhyk:
String literals that start with interpolated experession, e.g, `"${x}"`
are parsed into `"" ++ x.to_string()`, which translated to an extra
`string_append` call in Rust.  We cannot easily fix this in the parser,
as type inference relies on the first operand being a string to inject
the string coercion method (`to_string`).  We therefore drop the first
operand if it's an empty string in the code generator.

### DIFF
--- a/src/Language/DifferentialDatalog/Compile.hs
+++ b/src/Language/DifferentialDatalog/Compile.hs
@@ -3316,9 +3316,15 @@ mkExpr' d ctx e@EBinOp{..} = (v', EVal)
     t2 = exprType' d (CtxBinOpR e' ctx) (E $ sel3 exprRight)
     v = case exprBOp of
              Concat | t == tString
-                    -> case sel3 exprRight of
-                            EString _ s -> "::ddlog_rt::string_append_str(" <> e1 <> ", r###\"" <> pp s <> "\"###)"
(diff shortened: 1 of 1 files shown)


src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -3316,9 +3316,15 @@ mkExpr' d ctx e@EBinOp{..} = (v', EVal)
...
              Concat | t == tString
-                    -> case sel3 exprRight of
-                            EString _ s -> "::ddlog_rt::string_append_str(" <> e1 <> ", r###\"" <> pp s <> "\"###)"
-                            _           -> "::ddlog_rt::string_append(" <> e1 <> "," <+> ref exprRight <> ")"
+                    -> -- The parser scans expressions like '"${x}"' as '"" ++ x.to_string()'.
+                       -- The empty string operand helps type inference figure
+                       -- out that this is string concatenation, so we cannot
+                       -- eliminate it in the parser.  Drop it here instead.
+                       if E (sel3 exprLeft) == eString ""
+                       then val exprRight
```
mbudiu-vmw, line 3324: Technically you can drop empty strings in the right too.
There should be a strength reduction pass for this kind of stuff.
I guess rust is not smart enough to do it.

mbudiu-vmw (approved):

PR #1050, opened 2021-08-16
Optimize code generation for `match` expressions.

ryzhyk:
This was motivated by the following code generated for expressions of
the form `x?.field`:

```
match x {
        ddlog_std::Option::None{} => return (ddlog_std::Option::None{}),
        ddlog_std::Option::Some{x: ref __x} => (*__x).clone()
    }.field.clone()
```

which clones the entire `x`, which is potentially a large struct, just
to extract one field.

The optimization eliminates the clone and instead returns the result of
the `match` expression by reference.  This is only valid if all match
cases are the match expression are references:

```
match x {
        ddlog_std::Option::None{} => return (ddlog_std::Option::None{}),
        ddlog_std::Option::Some{x: ref __x} => __x
    }.field.clone()
```

### DIFF
--- a/src/Language/DifferentialDatalog/Compile.hs
+++ b/src/Language/DifferentialDatalog/Compile.hs
@@ -3238,8 +3238,7 @@ mkExpr' d ctx e@ESlice{..} = (mkSlice d (val exprOp, w) exprH exprL, EVal)
     e' = exprMap (E . sel3) e
     TBit _ w = exprType' d (CtxSlice e' ctx) $ E $ sel3 exprOp
 
--- Match expression is a reference
-mkExpr' d ctx e@EMatch{..} = (doc, EVal)
+mkExpr' d ctx e@EMatch{..} = (doc, if all_refs && not all_returns then EReference else EVal)
     where
     e' = exprMap (E . sel3) e
     m = deref exprMatchExpr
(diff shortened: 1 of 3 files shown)


Kixiron (approved):

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -3239,7 +3239,7 @@ mkExpr' d ctx e@ESlice{..} = (mkSlice d (val exprOp, w) exprH exprL, EVal)
     TBit _ w = exprType' d (CtxSlice e' ctx) $ E $ sel3 exprOp
 
 -- Match expression is a reference
-mkExpr' d ctx e@EMatch{..} = (doc, EVal)
+mkExpr' d ctx e@EMatch{..} = (doc, if all_refs && not all_returns then EReference else EVal)
```
mbudiu-vmw, line 3242: should the comment above be changed?

test/datalog_tests/simple3.dl
```diff
@@ -214,3 +214,79 @@ typedef Thosts = Thosts{id:Option<string>, capacity:Option<signed<64>>, up:Optio
 input relation Rhosts[Thosts] primary key (row) (row.id)
 output relation Rhostsv[Thosts]
 Rhostsv[v0] :- Rhosts[v],var v0 = v.
+
+
+/**********************************************************/
+
+// Test code generation for `match` expressions.
```
mbudiu-vmw, line 221: no ast files for this?

mbudiu-vmw (approved):

src/Language/DifferentialDatalog/Compile.hs
ryzhyk: good point

test/datalog_tests/simple3.dl
ryzhyk: we don't have those anymore, they were a paint to maintain.

PR #1052, opened 2021-08-17
Speedup JsonValue serialization.

ryzhyk:
`JsonValue` is the type we use to represent dynamically typed JSON in
DDlog.  We used to serialize it by converting to `serde::json::Value`
and serializing that.  This conversion is expensive and unnecessary, so
we implement `Serialize` for `JsonValue` natively instead.  We should
do the same for deserialize, but this is much more work, so we leave it
for when we really need it in the future.

One side effect of this change is that keys in JSON maps, which are
`istring`'s, are serialized in a deterministic, but unpredictable order,
instead of alphabetical order.

### DIFF
--- a/lib/json.rs
+++ b/lib/json.rs
@@ -45,50 +45,52 @@ pub fn to_json_value<T: serde::Serialize>(x: T) -> ddlog_std::Result<JsonValue, 
     res2std(serde_json::to_value(x).map(JsonValue::from))
 }
 
-pub struct ValueWrapper(serde_json::value::Value);
-
-impl serde::Serialize for ValueWrapper {
+impl serde::Serialize for JsonValue {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: serde::Serializer,
     {
         if serializer.is_human_readable() {
-            self.0.serialize(serializer)
+            match self {
+                JsonValue::JsonNull => serializer.serialize_unit(),
+                JsonValue::JsonBool { b } => serializer.serialize_bool(*b),
+                JsonValue::JsonNumber { ref n } => val_from_num(n.clone()).serialize(serializer),
+                JsonValue::JsonString { ref s } => serializer.serialize_str(s),
+                JsonValue::JsonArray { ref a } => a.serialize(serializer),
+                JsonValue::JsonObject { ref o } => {
+                    use serde::ser::SerializeMap;
+                    let mut map = serializer.serialize_map(Some(o.len()))?;
+                    for ddlog_std::tuple2(k, v) in o.iter() {
+                        map.serialize_entry(&k, &v)?;
+                    }
+                    map.end()
+                }
+            }
         } else {
-            serde_json::to_string(&self.0)
+            serde_json::to_string(self)
                 .map_err(|e| serde::ser::Error::custom(e))?
                 .serialize(serializer)
         }
     }
 }
 
-impl<'de> Deserialize<'de> for ValueWrapper {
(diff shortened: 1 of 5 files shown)


lib/json.rs
```diff
@@ -80,18 +94,6 @@ impl<'de> Deserialize<'de> for ValueWrapper {
     }
 }
 
-impl From<ValueWrapper> for JsonValue {
```
mbudiu-vmw, line 83: Is this where the benefit is coming from?

mbudiu-vmw (approved):

lib/json.rs
ryzhyk: yes

lib/json.rs
```diff
@@ -45,32 +45,46 @@ pub fn to_json_value<T: serde::Serialize>(x: T) -> ddlog_std::Result<JsonValue,
...
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: serde::Serializer,
     {
         if serializer.is_human_readable() {
-            self.0.serialize(serializer)
+            match self {
+                JsonValue::JsonNull => serializer.serialize_unit(),
+                JsonValue::JsonBool { b } => serializer.serialize_bool(*b),
+                JsonValue::JsonNumber { ref n } => val_from_num(n.clone()).serialize(serializer),
```
Kixiron, line 57: We should try and avoid this clone

lib/json.rs
```diff
@@ -45,32 +45,46 @@ pub fn to_json_value<T: serde::Serialize>(x: T) -> ddlog_std::Result<JsonValue,
...
 
-pub struct ValueWrapper(serde_json::value::Value);
-
-impl serde::Serialize for ValueWrapper {
+impl serde::Serialize for JsonValue {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
         S: serde::Serializer,
     {
         if serializer.is_human_readable() {
```
Kixiron, line 55: Do we need to explicitly handle this? Doesn't serde's infrastructure do the oh its own?

lib/json.rs
```diff
@@ -45,32 +45,46 @@ pub fn to_json_value<T: serde::Serialize>(x: T) -> ddlog_std::Result<JsonValue,
...
+                        map.serialize_entry(&k, &v)?;
+                    }
+                    map.end()
+                }
+            }
         } else {
-            serde_json::to_string(&self.0)
+            serde_json::to_string(self)
                 .map_err(|e| serde::ser::Error::custom(e))?
                 .serialize(serializer)
```
Kixiron, line 60: This looks really wrong, we shouldn't need this branch at all since the first branch just serializes the value and this branch will only end up calling the first branch when it serializes `self`

lib/json.rs
ryzhyk: This code could use some documentation, but it's there for a reason. The problem is that `serde::json::Value` cannot be deserialized from non-self-describing formats like bincode.  As a workaround, we first serialize the JSON value into a JSON string and store the string in bincode.  The next problem for which there is currently no good solution is that there is no way for the `Serialize`/`Deserialize` implementation to tell whether it is working with a self-describing format.  We use `is_human_readable` as a very rough and generally incorrect approximation of that.

lib/json.rs
ryzhyk: That's correct, it will use the first branch to serialize itself into a JSON string and then store the string inside the serializer.

lib/json.rs
ryzhyk: This one should be cheap, it's just cloning a numeric value.

PR #1053, opened 2021-08-18
Speedup hashing of interned objects.

ryzhyk:
We recently changed the implementation of `Hash` for interned objects to
hash the value of the object itself instead of hashing the pointer,
which is non-deterministic in some scenarios.  So obviously hashing
became much more expensive for interned types.  Of course the biggest
user of hashing is the internment library itself, as it hashes objects
in order to intern them.  So, interning an object whose fields are also
interned values became many times more expensive.  This for example
affected the performance of OVN in some scenarios.

To solve this, we change the way interned objects are hashed once again.
We already store a hash with each interned object, used to speedup
comparison operators by comparing by hash instead of by value.
Similarly, we now hash the interned object by hashing the stored hash
instead of hashing the object itself.  We also increase the stored hash
size to 64 bit.

### DIFF
--- a/lib/internment.rs
+++ b/lib/internment.rs
@@ -21,7 +21,7 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 */
 
-use ddlog_std::{hash32, Vec as DDlogVec};
+use ddlog_std::{hash64, Vec as DDlogVec};
 use differential_datalog::record::{self, Record};
 use internment::ArcIntern;
 use serde::{de::Deserializer, ser::Serializer};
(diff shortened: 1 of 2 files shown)


mbudiu-vmw (approved):

mbudiu-vmw:
Does this restore the OVN perf to its prior status?

blp:
> Does this restore the OVN perf to its prior status?

Yes, my measurements show that:

```
                       GB    elapsed time    CPU s
0.38                 16.8        5:13        277.5
0.42.1               16.5        5:11        271.2
0.42.1 + interning   12.9        4:44        248.0
0.43                 16.5       10:05        564.2
0.45.1               16.6        9:52        550.6
0.45.1 + interning   12.8       11:47        665.1
0.45.1 + 64b intern  12.9       11:21        640.3
0.45.1 + PR#1053     12.8        5:08        267.4
```

PR #1054, opened 2021-08-19
Java simulation of formal model

mbudiu-vmw:
This implementation is by no means complete, but I wanted to save persistently what I wrote so far.


mbudiu-vmw:
I will mark this as draft until I have written unit tests for all the classes involved.

java/streams/src/main/java/org/dbsp/circuits/operators/Operator.java
```diff
@@ -0,0 +1,154 @@
...
+    /**
+     * A new computation cycle has started, reset the internal state.
+     * Most operators do nothing.
+     */
+    public void reset() {}
+
+    public void connectTo(Operator to, int input) {
+        if (to.getInputWires().get(input) != null)
+            throw new RuntimeException(to.getName() + ": input " + input + " already connected");
+        Operator actual = to.getActualConsumer(to, input);
```
amytai, line 71: getActualConsumer should return (Operator, index) pair, where the index is the port number of the actual consumer

java/streams/src/main/java/org/dbsp/circuits/operators/IdOperator.java
```diff
@@ -0,0 +1,46 @@
...
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ */
+
+package org.dbsp.circuits.operators;
+
+import org.dbsp.algebraic.dynamicTyping.types.Type;
+
+/**
+ * An operator that works on streams.  It delays the input stream by 1 clock.
```
amytai, line 29: Delete 2nd sentence

mbudiu-vmw:
This should have addressed all the comments from the review.
I also added a couple more operators that are needed for computing on nested streams.

PR #1055, opened 2021-08-19
Change PrestoTranslator behavior when encountering an unsupported exp…

amytai:
…ression

Instead, the PrestoTranslator will now silently strip any expressions it cannot support.


mbudiu-vmw (approved):
I am not sure I like silent failures like this, but you know better what you need.

amytai:
I know, I also don't like it. Maybe we add a flag to the constructor whether to throw exception or silently discard, and default is to throw exception?

PR #1056, opened 2021-08-19
ovsdb2ddlog: --intern-strings option.

blp:
The new switch causes all strings in generated OVSDB tables to be emitted
as 'istring'.  This reduces memory use and can aid performance in programs
that use strings heavily.  In one OVN benchmark, this option reduced memory
use from 16.8 GB to 12.8 GB.

Based on a patch by Leonid Ryzhyk.

Signed-off-by: Ben Pfaff <bpfaff@vmware.com>

### DIFF
--- a/adapters/ovsdb/Main.hs
+++ b/adapters/ovsdb/Main.hs
@@ -41,6 +41,7 @@ import Language.DifferentialDatalog.Version
 
 data TOption = OVSFile         String
              | InternTable     String
+             | InternStrings
              | OutputTable     String
              | OutputOnlyTable String
              | MultisetTable   String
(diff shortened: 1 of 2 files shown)


ryzhyk:
Looks great, but we should also update the OVN test to use this feature. There are a couple of issues I ran into when trying to do that:
- There are many more tests now in the OVN test suite even if we only run DDlog tests, so I'd like to figure out a good subset
- There are still some nondeterministic test failures in CI. It'd be nice to investigate or at least disable them.
- We also need a new perf benchmark.  The one we use now is _really_ outdated.

@blp, I can work on all these, but will need a bit of help.  Let's chat offline!

ryzhyk:
@debnil , @blp , this is not urgent, but would be nice to one day add an additional ovsdb2ddlog test based on your nerpa work.

Issue #1057, opened 2021-08-22
Allow comments in DDLog interpreter


lykahb:
For better readability I'd like to be able to leave comments in the .dat files. Now the interpreter rejects the comments as an invalid syntax.

I have a workaround but it's rather ugly and doesn't work for the trailing comments.
```
test_ddlog/target/debug/test_cli < <(cat init.dat | grep -v "^//")
```

ryzhyk:
Lines starting with '#' are interpreted as comments by ddlog in  command files.

lykahb:
Thank you, that works.

Both Rust and Datalog use `//` for comments. Also, the top example in command_reference.md uses `//` for comments. So I didn't think of the alternatives. 

I opened a PR https://github.com/vmware/differential-datalog/pull/1058 with a fix for the examples.

ryzhyk:
Resolved via #1058.

PR #1058, opened 2021-08-23
Fix comment syntax in command examples

lykahb:
https://github.com/vmware/differential-datalog/issues/1057


ryzhyk:
Thanks for catching this!

PR #1066, opened 2021-08-30
ddlog-sql: translate array_length function to DDlog

amytai:
array_length is a non-standard SQL function that is alternately called array_length (in Postgres and H2 dialects) and cardinality (in Presto and Calcite dialects). Either instantiation of the function determines the length of the passed array.

For now, we need array_length for a use-case. If there are use-cases for `cardinality` in the future, we can add a similar function.

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/translator/ExpressionTranslationVisitor.java
+++ b/sql/src/main/java/com/vmware/ddlog/translator/ExpressionTranslationVisitor.java
@@ -469,6 +469,7 @@ public class ExpressionTranslationVisitor extends AstVisitor<DDlogExpression, Tr
                 return args.get(0).getType();
             case "count":
             case "count_distinct":
+            case "array_length":
                 if (args.size() == 0)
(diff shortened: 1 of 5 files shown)


sql/src/main/java/com/vmware/ddlog/translator/ExpressionTranslationVisitor.java
```diff
@@ -469,6 +469,7 @@ private static DDlogType functionResultType(Node node, String function, List<DDl
                 return args.get(0).getType();
             case "count":
             case "count_distinct":
+            case "array_length":
```
mbudiu-vmw, line 472: you could always add "cardinality" too.
That's a really bad name, btw.

mbudiu-vmw (approved):

amytai:
For simplicity, let's leave handling `cardinality` to another PR, when it's necessary.

mbudiu-vmw:
Can you please rebase on master so we can merge this?

amytai:
Rebased.

PR #1067, opened 2021-08-30
Use group_by operator instead of Aggregate in SQL compiler output

mbudiu-vmw:
Also, fix instances in the tutorial using Aggregate and replace with group_by.
Interestingly, most changes were already made in the test tutorial.dl.

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/ir/DDlogEApply.java
+++ b/sql/src/main/java/com/vmware/ddlog/ir/DDlogEApply.java
@@ -34,18 +34,31 @@ import java.util.List;
 public class DDlogEApply extends DDlogExpression {
     private final String func;
     private final List<DDlogExpression> args;
+    // If true the suffix notation with dot is preferred.
+    private boolean suffix;
 
-    public DDlogEApply(@Nullable Node node, String func, DDlogType type, DDlogExpression... args) {
+    public DDlogEApply(@Nullable Node node, String func, DDlogType type, boolean suffix, DDlogExpression... args) {
         super(node, type);
         this.func = func;
+        this.suffix = suffix;
         this.args = Arrays.asList(args);
+    }
+
+    public DDlogEApply(@Nullable Node node, String func, DDlogType type, DDlogExpression... args) {
+        this(node, func, type, false, args);
     }
 
     @Override
     public String toString() {
-        return this.func + "(" +
-                String.join(", ",
-                        Linq.map(this.args, DDlogExpression::toString)) + ")";
+        if (this.suffix && this.args.size() > 0) {
+            List<DDlogExpression> tail = this.args.subList(1, this.args.size());
+            return this.args.get(0).toString() + "." + this.func + "(" +
+                    String.join(", ", Linq.map(tail, DDlogExpression::toString)) + ")";
+        } else {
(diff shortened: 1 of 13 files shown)


mbudiu-vmw:
@amytai: can you please review the SQL compiler changes?

ryzhyk:
@mbudiu-vmw, @amytai, please make sure your SQL compiler improvements are documented in the `Unreleased` section of CHANGELOG.

mbudiu-vmw:
They aren't visible to users in any way; we are just avoiding using Aggregate for group_by.
It's not really worth a line in the changelog.
If anything the tutorial fixes are.

mbudiu-vmw:
BTW: @ryzhyk , you may need to review the tutorial changes.

ryzhyk:
> They aren't visible to users in any way; we are just avoiding using Aggregate for group_by.

I meant in general, not just this commit.

> BTW: @ryzhyk , you may need to review the tutorial changes.

Tutorial changes look good, thanks!

sql/src/main/java/com/vmware/ddlog/ir/DDlogEApply.java
```diff
@@ -34,18 +34,31 @@
 public class DDlogEApply extends DDlogExpression {
     private final String func;
     private final List<DDlogExpression> args;
+    // If true the suffix notation with dot is preferred.
+    private boolean suffix;
```
amytai, line 38: Is this used in this PR anywhere? If not, what's an intended use-case?

sql/src/main/java/com/vmware/ddlog/ir/DDlogRHSGroupby.java
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    @Override
+    public String toString() {
+        return "var " + this.var + " = " + this.rhsProject.toString() +
+                ".group_by((" + String.join(", ", this.rhsGroupByVars) + "))";
+    }
+
+    @Override
+    public boolean compare(DDlogRuleRHS val, IComparePolicy policy) {
```
amytai, line 50: Why do these nodes need to be Comparable?

amytai (approved):

sql/src/main/java/com/vmware/ddlog/ir/DDlogEApply.java
mbudiu-vmw: Not yet used, but may be, the result is easier to read.

sql/src/main/java/com/vmware/ddlog/ir/DDlogRHSGroupby.java
mbudiu-vmw: This is not the standard Java comparison. This was intended to make test comparisons more robust by checking if two DDlog programs are equivalent up to alpha-renaming. Then we would not need to change the test outputs by hand every time a new variable is renamed. But that would first require writing a DDlog parser, and we don't have that yet.

mbudiu-vmw:
Fixes #1068

mbudiu-vmw:
@amytai please review the last commit which hopefully fixes the issue you reported.

amytai (approved):

PR #1069, opened 2021-09-01
[self-profiler]: "Change profiling" feature. 

ryzhyk:
Added change profiling support to the DDlog self-profiler.  Unlike
arrangement size profiling, which tracks the number of records in each
arrangment, the change profile shows the amount of churn.  For example
adding one record and deleting one record will show up as two changes in
the change profile, but will cancel out in the size profile.
Identifying relations that see the most churn helps to understand the
performance of the program, as such relations are responsible for most
recomputation performed by DDlog.  In a typical profiling session, the
user identifies expensive operators in the CPU profile and looks up the
relations that are inputs to this operator in the change profile to
identify the ones with high churn.

The self-profiler now support the `profile change on/off` commands (also
available through the API), which enables change profiling for selected
transactions.  When change profiling is disabled, the recording stops,
but the previously accumulated profile is preserved.  By selectively
enabling change profiling for a subset of transactions, the user can
focus their analysis on specific parts of the program.

@blp

### DIFF
--- a/doc/profiling.md
+++ b/doc/profiling.md
@@ -63,6 +63,10 @@ addition to memory profiling.  CPU profiling is not enabled by default, as
 it can slow down the program somewhat, especially for large programs
 that handle many small updates.
 
+1. `profile change on/off;` - enables/disables recording of the number of
+insertions and deletions per arrangement, used to construct arrangement
+change profile (see below).
+
 1. `profile;` - returns information about program's CPU and memory usage.  CPU
 usage is expressed as the total amount of time DDlog spent evaluating each operator,
 assuming CPU profiling was enabled.  For example the following CPU profile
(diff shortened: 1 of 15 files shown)


doc/profiling.md
```diff
@@ -94,6 +98,22 @@ with the `LabeledNode(.node=child, .scc=childscc)` literal.
   451,529 and 372,446 records respectively (the numbered variables, e.g., `_0`)
   indicate one or more fields used to index the relation by.
 
+  In addition, if change profiling was enabled for some parts of the execution,
```
mbudiu-vmw, line 101: should this section have a little snippet of a change profile?

mbudiu-vmw (approved):

Kixiron (approved):
LGTM

doc/profiling.md
ryzhyk: ack

PR #1070, opened 2021-09-02
Cleaner and faster `group_by`.

ryzhyk:
Due to historical reasons and my stupidity, we ended up with an
implementation of `group_by` with odd semantics that was also expensive
to maintain.

The `group_by` transformation `<expr>.group_by(<vars>)` consists of
two operators: projection (by applying `<expr>` to all records) and
grouping records that share the same values of `<vars>`.  In the
previous implementation, grouping was performed before projection, so
given the following rule:

```
R(x, y, z),
var group = x.group_by(z).
```

and relation `R` containing records:

```
R(.x = 1, .y = 1, .z = 1), // weight =1
R(.x = 1, .y = 2, .z = 1), // weight =1
```

we produced a group with key `z=1` and two elements
`{(.x=1, .y=1), (.x=1, .y=2)}`, both with weight=1.  In this case,
`group.count()` would return 2.  The projection was applied while
enumerating the group, so that

```
for((x,w) in group) {...}
```

would encounter value `.x=1` with weight `1` twice.

In other words, this was kind of a multiset semantics where the same
value could appear multiple times, and the actual weight of the value in
the multiset had to be computed by summing up the weights of all
occurrences.  If one wanted to get the number of distinct elements in
the multiset or enumerate all distinct elements they had to first
convert the group into a set using `to_set()`.  All of this is really
easy to get wrong, and I've both encountered and written broken code
that didn't handle groups correctly.

Performance-wise, the biggest hit came from packing all visible
variables into the group at the cost of cloning them all, storing them
in an arrangement, and sorting values in a group (which DD does
internally).  Under some circumstances (large values or values with
expensive comparison operations) this could introduce huge overhead.
I've seen 100x in at least one benchmark.

The fix is to swap the order of operations:

1. Projection: takes a tuple of all variables visible in the current
   scope and applies `expr` to it to obtain a single value.  Returns
   a new collection where each record contains the group key (consisting
   of group-by variables) and the projected value. The weight of the
   record is the sum of all weights of records that are mapped to the
   same projected record.  In the above example, `y` is projected away
   and the projected relation only contains one record.
   ```
   R-projected(.z = 1, .x = 1), // weight = 2
   ```

2. Grouping: groups records that share the same key (internally,
   the projected relation is first arranged by key and the reduced).
   In our example this will yield a single group with one element
   with weight 2:
   ```
   R-grouped(Group{.key = 1, values = {(1, weight = 2)}})
   ```

One side effect of this is that the behavior of `Group::count()`
is different.  Instead of returning the number of values in a group
before project (2 in the example) it now returns the number of
unique values.  For example `().group_by(..).count()` always returns
1 in the new implementation, but could be greater than one before.
To avoid changing the behavior of existing code, I removed this function
altogether and added `count_unique()` instead, which is also a more
descriptive name for what the function actually does.

### DIFF
--- a/lib/ddlog_std.rs
+++ b/lib/ddlog_std.rs
@@ -1187,7 +1187,7 @@ pub fn hash128<T: Hash>(x: &T) -> u128 {
     ((w1 as u128) << 64) | (w2 as u128)
 }
 
-pub type ProjectFunc<X> = StdArc<dyn Fn(&DDValue) -> X + Send + Sync>;
+pub type ProjectFunc<X> = StdArc<dyn Fn(&DDValue) -> &X + Send + Sync>;
 
 /*
  * Group type (returned by the `group_by` operator).
@@ -1200,7 +1200,7 @@ pub type ProjectFunc<X> = StdArc<dyn Fn(&DDValue) -> X + Send + Sync>;
  * on both reference-backed and value-backed groups, we represent groups as
  * an enum and provide uniform API over both variants.
  *
- * There is a problem of managing the lifetime of a group.  Since one of the
+ * There is a problem in managing the lifetime of a group.  Since one of the
  * two variants contains references, the group type is parameterized by the
  * lifetime of these refs.  However, in order to be able to freely store and
  * pass groups to and from functions, we want `'static` lifetime.  Because
@@ -1250,7 +1250,7 @@ impl<K: Clone, V: Clone> Clone for Group<K, V> {
                 key: key.clone(),
                 group: group
                     .iter()
-                    .map(|(v, w)| tuple2(project(v), *w as DDWeight))
+                    .map(|(v, w)| tuple2(project(v).clone(), *w as DDWeight))
                     .collect(),
             },
             GroupEnum::ByVal { key, group } => GroupEnum::ByVal {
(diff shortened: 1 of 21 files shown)


ryzhyk:
@mbudiu-vmw , I need help fixing up the souffle converter for this.

lib/ddlog_std.rs
```diff
@@ -1514,7 +1514,7 @@ impl<V: Clone> Iterator for GroupIntoIter<V> {
         match self {
             GroupIntoIter::ByRef { iter, project } => match iter.next() {
                 None => None,
-                Some((x, w)) => Some(tuple2(project(x), *w as DDWeight)),
+                Some((x, w)) => Some(tuple2(project(x).clone(), *w as DDWeight)),
```
Kixiron, line 1517: I don't like all this cloning, do we have the ability to return references in ddlog right now?

Kixiron (approved):
LGTM

lib/ddlog_std.rs
ryzhyk: Agreed.  The clones were always there, hiding inside `project`. And yes, some of them can be removed now by changing various functions to return by-reference.  One unpleasant technicality is that we don't currently have a way to tell DDlog that group iterator returns `(&V, Weight)` tuples. We can do `(V, Weight)` or `&(V, Weight)`, but not `(&V, Weight)` or `(&V, &Weight)`.

I'll work on this once we've settled issues with the new semantics.

ryzhyk: Ok, this is trickier than I thought. Since this is not a regression, I'll merge as is and add a TODO in the code.

Issue #878, opened 2020-12-19
Incorrect result ?


jinyang175:
HI guys, 
Consider the following program:

```
input relation Z(a:string)
relation PQRI(a:string)
relation PLEY(a:string)
relation NFUV(a:string)
output relation OUT(a:string)



PQRI(v) :- Z(v), Z(nbj).
PLEY(o) :- PQRI(x), Z(o), Z(x).
NFUV(q) :- Z(fym), PLEY(q).
OUT(t) :- NFUV(ssz), PLEY(arv), PLEY(t).
```
If I run the program (after compiling in release mode), I get:

```
$ ./rules_orig_cli < facts.dat 
OUT{.a = "3u37ezDdBg"}
OUT{.a = "EbgMd2DV3K"}
OUT{.a = "uwsUc"}
OUT{.a = "zDdBgEbgMd"}
```

Now if i just add `PQRI(z), PQRI(x)` in rule `PLEY` to get a new program:

```
input relation Z(a:string)
relation PQRI(a:string)
relation PLEY(a:string)
relation NFUV(a:string)
output relation OUT(a:string)



PQRI(v) :- Z(v), Z(nbj).
PLEY(o) :- PQRI(x), Z(o), Z(x), PQRI(z), PQRI(x). // modified rule
NFUV(q) :- Z(fym), PLEY(q).
OUT(t) :- NFUV(ssz), PLEY(arv), PLEY(t).
```

after recompiling and running it again, now I get an empty result.

```
$ ./rules_orig_cli < facts.dat 

```
However, according to my understanding, this modification should not change the result of the computation. 
Right? or am I misunderstanding something?

```
$ cat facts.dat 
start;


insert Z("3u37ezDdBg"),
insert Z("uwsUc"),
insert Z("zDdBgEbgMd"),
insert Z("EbgMd2DV3K");

commit;

dump OUT;

```
I am running this on ddlog version: 5b1c9c1a70992032d271934f2652dcf6952b390d
Operating System: `Debian GNU/Linux 10 (buster)`
Kernel: `Linux 4.19.75.1.amd64-smp`
Architecture: `x86-64`

ryzhyk:
Yes, looks like a bug, thanks for reporting.  I will investigate.

ryzhyk:
So the good news is, this is a known... feature, yes, definitely feature. But clearly we are not handling this situation well.

[...]

PR #1075, opened 2021-09-10
Feature to use checked arithmetic on weights

mbudiu-vmw:
If this feature is enabled a ddlog program that overflows the weights will panic.
This partially addresses #878 - such programs will panic instead of giving the wrong result.


mbudiu-vmw:
I will resubmit this, got some wrong commits it looks like.

PR #1076, opened 2021-09-10
Overflow

### LINKED ISSUE #878: Incorrect result ?

mbudiu-vmw:
Feature to panic on weight overflow.
This partly fixes #878: instead of giving an incorrect result the program will crash.
Still, the user has no (simple) recourse.


rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,7 +89,111 @@ type TKeyAgent<S> = TraceAgent<KeyTrace<S>>;
...
+    }
+}
+
+impl Neg for CheckedWeight {
+    type Output = Self;
+
+    fn neg(self) -> Self::Output {
+        // intentional panic on overflow
+        CheckedWeight {
+            value: 0_i32.checked_sub(self.value).expect("Weight overflow"),
```
Kixiron, line 147: This is redundant, negation can't fail since the positive and negative integer spaces are equal. Any positive number can be converted into a negative one

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,7 +89,111 @@ type TKeyAgent<S> = TraceAgent<KeyTrace<S>>;
 type TValEnter<P, T> = TraceEnter<TValAgent<P>, T>;
 type TKeyEnter<P, T> = TraceEnter<TKeyAgent<P>, T>;
 
-/// Diff associated with records in differential dataflow
+#[derive(
+    Abomonation, Copy, Ord, PartialOrd, Eq, PartialEq, Hash, Debug, Clone, Serialize, Deserialize,
+)]
+pub struct CheckedWeight {
```
Kixiron, line 95:
```suggestion
#[derive(
    Abomonation, Copy, Ord, PartialOrd, Eq, PartialEq, Hash, Debug, Clone, Serialize, Deserialize,
)]
#[repr(transparent)]
pub struct CheckedWeight {
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,7 +89,111 @@ type TKeyAgent<S> = TraceAgent<KeyTrace<S>>;
...
+    pub value: i32,
+}
+
+/// CheckedWeight trait implementations
+
+impl Semigroup for CheckedWeight {
```
Kixiron, line 101:
```suggestion
impl Semigroup for CheckedWeight {
```
This doc comment doesn't add anything

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,7 +89,111 @@ type TKeyAgent<S> = TraceAgent<KeyTrace<S>>;
...
+            .expect("Weight overflow");
+    }
+}
+
+impl Add for CheckedWeight {
+    type Output = Self;
+
+    fn add(self, other: Self) -> Self {
+        // intentional panic on overflow
+        CheckedWeight {
```
Kixiron, line 122:
```suggestion
        Self {
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,7 +89,111 @@ type TKeyAgent<S> = TraceAgent<KeyTrace<S>>;
...
+                .expect("Weight overflow"),
+        }
+    }
+}
+
+impl Mul for CheckedWeight {
+    type Output = Self;
+    fn mul(self, rhs: Self) -> Self::Output {
+        // intentional panic on overflow
+        CheckedWeight {
```
Kixiron, line 135:
```suggestion
        Self {
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,7 +89,111 @@ type TKeyAgent<S> = TraceAgent<KeyTrace<S>>;
...
+        }
+    }
+}
+
+impl Neg for CheckedWeight {
+    type Output = Self;
+
+    fn neg(self) -> Self::Output {
+        // intentional panic on overflow
+        CheckedWeight {
```
Kixiron, line 146:
```suggestion
        Self {
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,7 +89,111 @@ type TKeyAgent<S> = TraceAgent<KeyTrace<S>>;
...
+            value: 0_i32.checked_sub(self.value).expect("Weight overflow"),
+        }
+    }
+}
+
+impl Monoid for CheckedWeight {
+    fn zero() -> Self {
+        CheckedWeight { value: 0 }
```
Kixiron, line 154:
```suggestion
        Self { value: 0 }
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,7 +89,111 @@ type TKeyAgent<S> = TraceAgent<KeyTrace<S>>;
...
+    }
+}
+
+impl One for CheckedWeight {
+    fn one() -> Self {
+        CheckedWeight { value: 1 }
```
Kixiron, line 160:
```suggestion
        Self { value: 1 }
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,7 +89,111 @@ type TKeyAgent<S> = TraceAgent<KeyTrace<S>>;
...
+    }
+}
+
+impl Zero for CheckedWeight {
+    fn zero() -> Self {
+        CheckedWeight { value: 0 }
+    }
+    fn is_zero(&self) -> bool {
```
Kixiron, line 168:
```suggestion
        Self { value: 0 }
    }

    fn is_zero(&self) -> bool {
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,7 +89,111 @@ type TKeyAgent<S> = TraceAgent<KeyTrace<S>>;
...
+        self.value == 0
+    }
+}
+
+impl From<i32> for CheckedWeight {
+    fn from(item: i32) -> Self {
+        CheckedWeight { value: item }
```
Kixiron, line 175:
```suggestion
    fn from(value: i32) -> Self {
        Self { value }
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,7 +89,111 @@ type TKeyAgent<S> = TraceAgent<KeyTrace<S>>;
...
+    }
+}
+
+impl From<i8> for CheckedWeight {
+    fn from(item: i8) -> Self {
+        CheckedWeight { value: item as i32 }
```
Kixiron, line 181:
```suggestion
        Self { value: item as i32 }
```

rust/template/differential_datalog/src/program/worker.rs
```diff
@@ -66,7 +67,13 @@ struct SessionData {
...
         ArrId,
         TraceAgent<
-            Spine<DDValue, DDValue, u32, i32, Rc<OrdValBatch<DDValue, DDValue, u32, i32, u32>>>,
+            Spine<
+                DDValue,
+                DDValue,
+                u32,
+                Weight,
+                Rc<OrdValBatch<DDValue, DDValue, u32, Weight, u32>>,
+            >,
```
Kixiron, line 76: I believe this can be replaced with `OrdValSpine`

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -1898,7 +2006,7 @@ impl Program {
                                 &collection_keys,
                                 arr.filter(move |_, v| f(v)),
                                 |k, key| *key = k.clone(),
-                                move |_, w1, v2, w2| (jfun(v2), w1 * w2),
+                                move |_, w1, v2, w2| (jfun(v2), (*w1) * (*w2)),
```
Kixiron, line 2009:
```suggestion
                                move |_, &w1, v2, &w2| (jfun(v2), w1 * w2),
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -1887,7 +1995,7 @@ impl Program {
                                 &collection_keys,
                                 arr.clone(),
                                 |k, key| *key = k.clone(),
-                                move |_, w1, v2, w2| (jfun(v2), w1 * w2),
+                                move |_, w1, v2, w2| (jfun(v2), (*w1) * (*w2)),
```
Kixiron, line 1998:
```suggestion
                                move |_, &w1, v2, &w2| (jfun(v2), w1 * w2),
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -1847,7 +1955,7 @@ impl Program {
                                 &collection_with_keys,
                                 arr.filter(move |_, v| f(v)),
                                 |(k, _), key| *key = k.clone(),
-                                move |v1, w1, v2, w2| (jfun(v2, &v1.1), w1 * w2),
+                                move |v1, w1, v2, w2| (jfun(v2, &v1.1), (*w1) * (*w2)),
```
Kixiron, line 1958:
```suggestion
                                move |v1, &w1, v2, &w2| (jfun(v2, &v1.1), w1 * w2),
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -1836,7 +1944,7 @@ impl Program {
                                 &collection_with_keys,
                                 arr.clone(),
                                 |(k, _), key| *key = k.clone(),
-                                move |v1, w1, v2, w2| (jfun(v2, &v1.1), w1 * w2),
+                                move |v1, w1, v2, w2| (jfun(v2, &v1.1), (*w1) * (*w2)),
```
Kixiron, line 1947:
```suggestion
                                move |v1, &w1, v2, &w2| (jfun(v2, &v1.1), w1 * w2),
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -1628,7 +1736,7 @@ impl Program {
                         &collection_with_keys,
                         arr,
                         |(k, _), key| *key = k.clone(),
-                        move |v1, w1, _, w2| (jfun(&v1.1), w1 * w2),
+                        move |v1, w1, _, w2| (jfun(&v1.1), (*w1) * (*w2)),
```
Kixiron, line 1739:
```suggestion
                        move |v1, &w1, _, &w2| (jfun(&v1.1), w1 * w2),
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -1595,7 +1703,7 @@ impl Program {
                         &collection_with_keys,
                         arr,
                         |(k, _), key| *key = k.clone(),
-                        move |v1, w1, v2, w2| (jfun(&v1.1, v2), w1 * w2),
+                        move |v1, w1, v2, w2| (jfun(&v1.1, v2), (*w1) * (*w2)),
```
Kixiron, line 1706:
```suggestion
                        move |v1, &w1, v2, &w2| (jfun(&v1.1, v2), w1 * w2),
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -1402,7 +1510,7 @@ impl Program {
                         &collection_with_keys,
                         arr,
                         |(k, _), key| *key = k.clone(),
-                        move |v1, w1, _, w2| (jfun(&v1.1), w1 * w2),
+                        move |v1, w1, _, w2| (jfun(&v1.1), (*w1) * (*w2)),
```
Kixiron, line 1513:
```suggestion
                        move |v1, &w1, _, &w2| (jfun(&v1.1), w1 * w2),
```

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -1369,7 +1477,7 @@ impl Program {
                         &collection_with_keys,
                         arr,
                         |(k, _), key| *key = k.clone(),
-                        move |v1, w1, v2, w2| (jfun(&v1.1, v2), w1 * w2),
+                        move |v1, w1, v2, w2| (jfun(&v1.1, v2), (*w1) * (*w2)),
```
Kixiron, line 1480:
```suggestion
                        move |v1, &w1, v2, &w2| (jfun(&v1.1, v2), w1 * w2),
```

Kixiron (changes requested):
There's also an extraneous file, `test/datalog_tests/prof.profile`

ryzhyk:
@mbudiu-vmw, can you please measure how this feature affects the benchmarks in `differential-datalog/rust/ddlog_benches/`?

mbudiu-vmw:
I will measure perf, but we should also run all  other programs in this mode. We could also add configs for bigger weights, although even 64 bits is not much.

mbudiu-vmw:
The best thing about this patch is that it immediately suggests a safe solution which I plan to prototype: make the weight an `Either<int32, BigInt>`.

ryzhyk:
> I will measure perf, but we should also run all other programs in this mode. We could also add configs for bigger weights, although even 64 bits is not much.

This PR should configure a subset of the test suite to run with this feature enabled.

rust/template/differential_datalog/src/program/mod.rs
mbudiu-vmw: They are not really equal, but indeed the negative one is bigger than the positive one.

mbudiu-vmw: But this check is actually necessary, in case that self.value is MIN_INT: https://towardsdatascience.com/unsinged-signed-integers-and-casting-in-rust-9a847bfc398f#3e54

rust/template/differential_datalog/src/program/mod.rs
mbudiu-vmw: > 
> This PR should configure a subset of the test suite to run with this feature enabled.

Any suggestions? Do we have any tools to measure the perf difference?

ryzhyk: This is just for correctness, not performance.  Speaking of which, a test that actually overflows and crashes would be nice too.

As far as performance goes, the benchmarks in `differential-datalog/rust/ddlog_benches` is the best we have right now.  (see my other comment).

mbudiu-vmw: The test overflow.dl crashes with this feature. It is based on the bug report in #878

mbudiu-vmw:
On the twitter-micro benchmark the perf impact is in the noise: most tests have unchanged perf, some are faster, and some are slower (according to cargo bench).

mbudiu-vmw:
twitter-macro benchmark results are similar. This is measured in a VM, so there is significant noise.

rust/ddlog_benches/Makefile.toml
```diff
@@ -1,6 +1,6 @@
 # Runs all ddlog benchmarks
 [tasks.benchmarks]
-dependencies = ["download-data", "bench-twitter", "bench-livejournal"]
+dependencies = ["bench-twitter", "bench-livejournal"]
```
Kixiron, line 3: What's the reason for this change?

.gitignore
```diff
@@ -31,6 +31,7 @@ test/datalog_tests/*.dump
 test/datalog_tests/*.c
 test/datalog_tests/*_ddlog
 test/datalog_tests/debug.log
+test/datalog_tests/prof.profile
```
Kixiron, line 34: This should go in `.git/info/exclude` instead of the committed gitignore

rust/ddlog_benches/Makefile.toml
mbudiu-vmw: without this change 'cargo bench' fails.

Kixiron: What's the error message?

.gitignore
mbudiu-vmw:
```
[cargo-make] INFO - Running Task: download-data
downloading '' to 'data/' 
[cargo-make] ERROR - Error while running duckscript: Source: Unknown Line: 10 - Invalid URL:  provided.
[cargo-make] WARN - Build Failed.
```
There isn't really a task download-data, it is a task invoked by other tasks.

rust/ddlog_benches/Makefile.toml
Kixiron, line 3:
```suggestion
dependencies = [
    "download-twitter",
    "bench-twitter",
    "download-livejournal",
    "bench-livejournal",
]
```

mbudiu-vmw: Is this really necessary? It looks to me like when you run 'cargo make benchmarks' these will be done through transitive dependencies anyway. For me it works fine as it is in this PR.

.gitignore
mbudiu-vmw: Why should it be in .git/info/exclude? I assume no one wants to commit that file by mistake, as I did.

Kixiron: The file existing isn't part of the project, it's an artifact of your particular dev flow

mbudiu-vmw: It was generated when running some tests with their default flags, I didn't do anything special for that.

rust/template/differential_datalog/src/program/mod.rs
```diff
@@ -85,7 +89,110 @@ type TKeyAgent<S> = TraceAgent<KeyTrace<S>>;
 type TValEnter<P, T> = TraceEnter<TValAgent<P>, T>;
 type TKeyEnter<P, T> = TraceEnter<TKeyAgent<P>, T>;
 
-/// Diff associated with records in differential dataflow
+#[derive(
```
ryzhyk, line 92: Do we need `Serialize`/`Deserialize` for weights?  The auto-derived implementations will likely not do what you want here.  I think you also need `#[serde(transparent)]`

rust/template/differential_datalog/src/program/mod.rs
ryzhyk: I actually find the original version easier to read. @Kixiron , so you think `Self` is more idiomatic?
In any case, @mbudiu-vmw , since you made this change everywhere you probably want it here as well.

rust/template/differential_datalog/src/program/mod.rs
ryzhyk: Why not use `checked_neg`?

rust/template/differential_datalog/src/program/mod.rs
ryzhyk: Do you want to apply @Kixiron 's suggestion here and in other places below?

rust/template/differential_datalog/src/program/mod.rs
ryzhyk: Why do we have this `impl` specifically for `i8`?

rust/template/differential_datalog/src/program/mod.rs
mbudiu-vmw: In general I added stuff that seemed to be necessary or the program would not compile.
I don't really understand why Abomonation is needed either.
I can try to take it out and see what breaks.
I really don't understand why the auto-derived implementation would be wrong.

rust/template/differential_datalog/src/program/mod.rs
Kixiron: Yah, `Self` is more idiomatic and it's more resistant to annoyances while refactoring since it just refers to the current struct, one less thing to change

mbudiu-vmw:
There is no checked_neg in num.

ryzhyk:
> There is no checked_neg in num.

Can you use this: https://doc.rust-lang.org/stable/std/primitive.i32.html#method.checked_neg

PR #1078, opened 2021-09-10
Implement `unwrap_or` in Rust.

ryzhyk:
Implement `Option::unwrap_or` and `Result::unwrap_or` in Rust to avoid
redundant cloning.

### DIFF
--- a/lib/ddlog_std.dl
+++ b/lib/ddlog_std.dl
@@ -209,11 +209,8 @@ extern function default(): 'T
 
 /* Returns the contained `Some` value or a provided default.
  */
-function unwrap_or(x: Option<'A>, def: 'A): 'A = {
-    match (x) {
-        Some{v} -> v,
-        None -> def
-    }
+function unwrap_or(#[by_val] x: Option<'A>, #[by_val] def: 'A): 'A = {
+    option_unwrap_or(x, def)
 }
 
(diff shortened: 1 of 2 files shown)


lib/ddlog_std.dl
```diff
@@ -209,11 +209,8 @@ extern function default(): 'T
 
 /* Returns the contained `Some` value or a provided default.
  */
-function unwrap_or(x: Option<'A>, def: 'A): 'A = {
-    match (x) {
-        Some{v} -> v,
-        None -> def
-    }
+function unwrap_or(#[by_val] x: Option<'A>, def: 'A): 'A = {
```
mbudiu-vmw, line 212: are these used much?

mbudiu-vmw (approved):

lib/ddlog_std.dl
Kixiron, line 212: Why not make both params by value?

ryzhyk: yes, especially in projects that declare many structs with `Option<>` fields.

ryzhyk: It's probably going to make less of a difference, but there's no real downside either.  The main concern with making something `#[by_val]` is that the called may then need to clone if they only have access to the value by reference.  Here this should not be an issue.

Issue #1083, opened 2021-09-17
SQL to DDlog compiler misses table name in context for windowed queries over JOINed tables


amytai:
Query:
```
SELECT DISTINCT t1.column2, AVG(t2.column1) OVER (PARTITION by t1.column3) AS avg
         FROM t1
         JOIN t2
             ON t1.column1 = t2.column1
```

Error:
```
com.vmware.ddlog.translator.TranslationException: Could not resolve identifier
t1 line: 2 column: 17

	at com.vmware.ddlog.translator.ExpressionTranslationVisitor.visitIdentifier(ExpressionTranslationVisitor.java:349)
	at com.vmware.ddlog.translator.ExpressionTranslationVisitor.visitIdentifier(ExpressionTranslationVisitor.java:39)
	at com.facebook.presto.sql.tree.Identifier.accept(Identifier.java:70)
	at com.facebook.presto.sql.tree.AstVisitor.process(AstVisitor.java:27)
...
```

ryzhyk:
@amytai , ok to close this issue?

PR #1085, opened 2021-09-20
Fix for issue #1083

mbudiu-vmw:
When compiling window SQL queries we actually generate 3 sub-queries. The generated subqueries no longer operate on the same input tables, so we have to adjust the column names accordingly.

### DIFF
--- a/sql/src/test/java/ddlog/WindowTest.java
+++ b/sql/src/test/java/ddlog/WindowTest.java
@@ -338,4 +338,39 @@ public class WindowTest extends BaseQueriesTest {
                 "var aggResult15 = agg1(groupResult16),var v14 = TRtmp0{.gb1 = gb9,.sum = aggResult15.sum},var v17 = v14.";
         this.testTranslation(query, translation);
     }
+
+    @Test
+    public void joinWindowTest() {
+        String query = "SELECT DISTINCT t1.column2, AVG(t2.column1) OVER (PARTITION by t1.column3) AS avg\n" +
+                "         FROM t1 JOIN t2 ON t1.column1 = t2.column1\n";
+        String translation = this.header(false) +
+                "typedef Ttmp = Ttmp{column1:signed<64>, column2:string, column3:bool, column4:double, column10:signed<64>}\n" +
+                "typedef TRtmp = TRtmp{tmp:signed<64>, gb:bool, column2:string}\n" +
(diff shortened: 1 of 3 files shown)


mbudiu-vmw:
@amytai please review

mbudiu-vmw:
Fixes #1083

sql/src/test/java/ddlog/WindowTest.java
```diff
@@ -338,4 +338,39 @@ public void nestedAggTest() {
...
+                "relation Roverinput[TRtmp]\n" +
+                "relation Rtmp0[TRtmp0]\n" +
+                "relation Rover[TRtmp0]\n" +
+                "relation Rtmp2[TRtmp2]\n" +
+                "Roverinput[v6] :- Rt1[v2],Rt2[v3],(v2.column1 == v3.column1),true," +
+                "var v4 = Ttmp{.column1 = v2.column1,.column2 = v2.column2,.column3 = v2.column3,.column4 = v2.column4,.column10 = v3.column1}" +
+                ",var v5 = TRtmp{.tmp = v3.column1,.gb = v2.column3,.column2 = v2.column2},var v6 = v5.\n" +
+                "Rover[v11] :- Roverinput[v7],var gb8 = v7.gb,var groupResult = (v7).group_by((gb8))," +
+                "var aggResult = agg(groupResult),var v10 = TRtmp0{.gb = gb8,.avg = aggResult.avg},var v11 = v10.";
+        this.testTranslation(query, translation);
```
amytai, line 374: This is weird, why is there no output relation? When I apply these changes locally, I get
```
"Rv1[v16] :- Roverinput[v12],Rover[v13],(true and (v12.gb == v13.gb)),var v14 = Ttmp1{.tmp = v12.tmp,.gb = v12.gb,.column2 = v12.column2,.avg = v13.avg},var v15 = TRtmp2{.column2 = v12.column2,.min = v13.avg},var v16 = v15.";
 ```
as the output relation, which seems correct.

amytai (changes requested):

sql/src/test/java/ddlog/WindowTest.java
mbudiu-vmw: "CREATE VIEW" was missing.

amytai (approved):

PR #1087, opened 2021-09-22
Windows and Joins work when columns have alias

amytai:
Also make sure that columns in the DDlog output relation appear in the same order as in the original SQL create statement.

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/translator/TranslationVisitor.java
+++ b/sql/src/main/java/com/vmware/ddlog/translator/TranslationVisitor.java
@@ -1038,7 +1038,7 @@ class TranslationVisitor extends AstVisitor<DDlogIRNode, TranslationContext> {
         List<SelectItem> finalItems = new ArrayList<SelectItem>();
         SubstitutionRewriter rewriter = new SubstitutionRewriter(windowVisitor.substitutions);
         ExpressionTreeRewriter<Void> subst = new ExpressionTreeRewriter<Void>(rewriter);
-        ExpressionTreeRewriter<Void> dropTable = new ExpressionTreeRewriter<Void>(new ColumnContextEliminationRewriter());
+        ExpressionTreeRewriter<Void> dropTableExprRewriter = new ExpressionTreeRewriter<Void>(new ColumnContextEliminationRewriter());
 
         // Add columns to finalItems in the order that they appear in the original table.
         for (SelectItem originalColumn : query.getSelect().getSelectItems()) {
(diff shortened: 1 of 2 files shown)


sql/src/test/java/ddlog/WindowTest.java
```diff
@@ -381,4 +381,93 @@ public void joinWindowTest() {
...
+                "(var avg9 = (64'sd0, 64'sd0): (signed<64>, signed<64>));\n" +
+                "(for ((i, _) in g) {\n" +
+                "var v7 = i;\n" +
+                "(var incr = v7.tmp);\n" +
+                "(avg9 = agg_avg_signed_R(avg9, incr))}\n" +
+                ");\n" +
+                "(Tagg{.avg = avg_signed_R(avg9)})\n" +
+                "}\n" +
+                "\n" +
+                "input relation Rt1[Tt1]\n" +
```
mbudiu-vmw, line 407: usually I write `this.relations(false)`; it's just slightly more robust if you add more input relations.

sql/src/main/java/com/vmware/ddlog/translator/TranslationVisitor.java
```diff
@@ -1040,11 +1040,50 @@ private DDlogIRNode processWindows(QuerySpecification query,
...
+                }
+            } else if (originalColumn instanceof SingleColumn) {
+                SingleColumn scOrigCol = (SingleColumn) originalColumn;
+                boolean found = false;
+                for (SingleColumn sc: Utilities.concatenate(aggregateItems, windowItems)) {
+                    // Aggregate or window items should be referenced by their name in the OverInput or Over tables
+                    // created by this Visitor.
+                    if (scOrigCol.getExpression().equals(sc.getExpression())) {
+                        Expression repl = subst.rewrite(sc.getExpression(), null);
+                        Expression repl1 = dropTable.rewrite(repl, null);
```
mbudiu-vmw, line 1063: maybe we should rename the 'dropTable' variable as 'dropTableSpecRewriter' or something like that.

mbudiu-vmw (changes requested):

sql/src/test/java/ddlog/WindowTest.java
amytai: done

sql/src/main/java/com/vmware/ddlog/translator/TranslationVisitor.java
amytai: How about `dropTableExprRewriter`?

mbudiu-vmw (approved):

PR #1089, opened 2021-09-26
Translate vector-type fields properly in DDlogJooqProvider

amytai:

Previously, vector-type columns (i.e. ARRAY_AGG) would not get properly handled by the DDlogJooqProvider during writeback (DDlogJooqProvider::onChange), because the Jooq H2 backend stores the type of array columns as Object.class. Instead, in the writeback function, we infer the type of the column from the DDlog record, which should have all typing informaiton.

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
+++ b/sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
@@ -558,24 +558,49 @@ public final class DDlogJooqProvider implements MockDataProvider {
         }
     }
 
-    private static void setValue(final Field<?> field, final DDlogRecord in, final Record out) {
-        final DataType<?> dataType = field.getDataType();
-        switch (dataType.getSQLType()) {
-            case Types.BOOLEAN:
-                out.setValue((Field<Boolean>) field, in.getBoolean());
-                return;
-            case Types.INTEGER:
-                out.setValue((Field<Integer>) field, in.getInt().intValue());
-                return;
-            case Types.BIGINT:
-                out.setValue((Field<Long>) field, in.getInt().longValue());
-                return;
-            case Types.VARCHAR:
-                out.setValue((Field<String>) field, in.getString());
-                return;
-            default:
-                throw new DDlogJooqProviderException("Unknown datatype " + field);
+    // Return a value based on the DDlog type of the record
+    private static Object ddlogRecordToObject(final Field<?> field, final DDlogRecord in) {
+        if (in.isBool()) {
+            return in.getBoolean();
+        } else if (in.isInt()) {
+            if (field.getDataType().getSQLType() == Types.BIGINT) {
+                return in.getInt().longValue();
+            } else {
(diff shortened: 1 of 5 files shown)


sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -558,26 +558,51 @@ private static DDlogRecord toValue(final Field<?> field, final Object in) {
...
-                out.setValue((Field<Long>) field, in.getInt().longValue());
-                return;
-            case Types.VARCHAR:
-                out.setValue((Field<String>) field, in.getString());
-                return;
-            default:
-                throw new DDlogJooqProviderException("Unknown datatype " + field);
+    // Return a value based on the DDlog type of the record
+    private static Object ddlogRecordToObject(final Field<?> field, final DDlogRecord in) {
+        if (in.isBool()) {
```
mbudiu-vmw, line 563: How about adding a 'debug' flag which will cause the code to assert that the SQL type matches as well?

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -558,26 +558,51 @@ private static DDlogRecord toValue(final Field<?> field, final Object in) {
...
+            } else {
+                return in.getInt().intValue();
+            }
+        } else if (in.isDouble()) {
+            return in.getDouble();
+        } else if (in.isFloat()) {
+            return in.getFloat();
+        } else if (in.isString()) {
+            return in.getString();
+        } else if (in.isStruct()) {
```
mbudiu-vmw, line 577: how come that the original code didn't have this path?
were there no nulls so far in the tests?

sql/src/test/java/ddlog/JooqProviderTestBase.java
```diff
@@ -539,6 +539,34 @@ public void testInsertNullFails1() {
         }
     }
 
+    @Test
+    public void testArrayAggTypes() {
+        assert(create != null);
+        create.execute("insert into \nhosts values ('n1', 10, true)");
```
mbudiu-vmw, line 545: why the newline here?

sql/src/test/java/ddlog/JooqProviderTestBase.java
```diff
@@ -539,6 +539,34 @@ public void testInsertNullFails1() {
...
+        create.batch("insert into hosts values ('n54', 18, false)",
```
mbudiu-vmw, line 546: none of your tests has an array with more than 1 element, or with a null element.
is it possible to add these?

mbudiu-vmw (changes requested):

sql/src/test/java/ddlog/JooqProviderTestBase.java
amytai: Added both types of array results.

mbudiu-vmw (approved):

Issue #1092, opened 2021-09-30
Asymmetric ownership of records in API


mbudiu-vmw:
The function ddlog_query_index in ddlog.h does not take ownership of the key record.
This puts the burden on the user on deallocating the key.
This is different from all other API calls in ddlog.h.
Should this function also take ownership of the key record?

ryzhyk:
I think the idea is that the caller can perform multiple queries with the same key without re-creating it every time.

PR #1093, opened 2021-09-30
Add queryIndex Java API

mbudiu-vmw:


mbudiu-vmw:
@amytai @lalithsuresh 
This should fix #1091

java/ddlogapi/DDlogRecord.java
```diff
@@ -51,7 +51,14 @@ public void release() {
...
 
-    private long checkHandle() {
+    /**
+     * Same as 'release', for symmetry with the DDlogCommandVector API.
+     */
+    public void dispose() {
+        this.release();
+    }
+
+    long checkHandle() {
```
amytai, line 61: `checkHandle` dropped the private annotation. is that intentional?

java/test-index/A.java
```diff
@@ -0,0 +1,71 @@
...
+        // Deallocate the result
+        result.dispose();
+
+        // Dump the entire index.
+        this.api.dumpIndex("AII", i -> System.out.println("dump: " + i));
+        rec = new DDlogRecord(0);
+        // Reuse the same key to query the AIII index.
+        this.api.queryIndex("AIII", rec, i -> System.out.println("query 0: " + i));
+        // The key is no longer useful, we need to deallocate it explicitly.
+        rec.dispose();
```
amytai, line 45: Not actually sure why this is necessary. Where are you releasing this memory from? Isn't `rec` just an object to the JVM and will get GC'd accordingly?

Question applies to other `dispose` calls as well.

java/test-index/A.java
```diff
@@ -0,0 +1,71 @@
...
+
+        // Create a tuple key for the AIIII index
+        DDlogRecord r0 = new DDlogRecord(0);
+        DDlogRecord r1 = new DDlogRecord(1);
+        DDlogRecord tuple = DDlogRecord.makeTuple(r0, r1);
+        this.api.queryIndex("AIIII", tuple, i -> System.out.println("query (0,1): " + i));
+        // Deallocate the key
+        tuple.dispose();
+
```
amytai, line 60: Can you add a query for a key that isn't in an index?

java/ddlogapi.c
```diff
@@ -444,6 +444,11 @@ JNIEXPORT void JNICALL Java_ddlogapi_DDlogAPI_00024DDlogCommandVector_ddlog_1bat
     ddlog_free_record_updates((ddlog_record*)handle, size);
 }
 
+JNIEXPORT void JNICALL Java_ddlogapi_DDlogAPI_ddlog_1free(
```
amytai, line 447: where is this used?

amytai (changes requested):

mbudiu-vmw:
in dispose.

java/ddlogapi.c
mbudiu-vmw: It's used in dispose.

java/ddlogapi/DDlogRecord.java
mbudiu-vmw: Yes, because it needs to be called by DDlogAPI, which is in the same package.

java/test-index/A.java
mbudiu-vmw: The queryIndex api does not consume the keys, they have to be disposed by the caller.
See also #1092.
The DDlogRecord objects are allocated by the native world and are not GC-ed. That's why they have stringent lifetime requirements. Most often you don't see them because each object is used exactly once (after which the native API usually takes ownership and is responsible for deleting it). This API is different. The Java GC may collect the record, but the record contains a pointer which will never be freed by GC. In a language by C# this is done by objects implementing an IDisposable interface, but Java 8 does not have an equivalent.

java/test-index/A.java
mbudiu-vmw: Query for 1 above is such an example; it does not have any output since key 1 is not present.

lalithsuresh: It would be good to assert on expected return values from these queryIndex calls and check their state than to rely on print statements.

```
var result = new ArrayList<DDlogRecord>();
this.api.queryIndex("AIIII", tuple, result::add)
assertTrue(result.size() == 1 && result.get(0).<check contents>);
```

mbudiu-vmw: Actually the test results are checked automatically by the run.sh script, which captures stdout and compares it with `atest.dump.expected`. Many of the DDlog tests have this shape.

mbudiu-vmw: This is because there is no `assert` in the ddlog CLI testing tool.

java/ddlogapi.c
amytai: Well, dispose is a wrapper around release, which is why I was confused. `release` was not touched by this PR, so it must already have access to `ddlog_free` if it needed it?

mbudiu-vmw: release was actually never called by anyone.
This is the first use of it.
There should not be both of them, but I wanted to keep backwards compatibility.

amytai (approved):

ryzhyk:
Thanks! Please don't forget to update the CHANGELOG (I'll do it for this PR when preparing the next release).

mbudiu-vmw:
sorry. perhaps the commit check should require a diff in changelog.

PR #1095, opened 2021-10-04
Add functionality to translate create index statements from SQL to DDlog

amytai:

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/ir/DDlogIndexDeclaration.java
+++ b/sql/src/main/java/com/vmware/ddlog/ir/DDlogIndexDeclaration.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright (c) 2021 VMware, Inc.
+ * SPDX-License-Identifier: MIT
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in all
+ * copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+ * SOFTWARE.
+ *
+ */
+
+package com.vmware.ddlog.ir;
+
+import com.facebook.presto.sql.tree.Node;
+
+import javax.annotation.Nullable;
+import java.util.List;
+import java.util.stream.Collectors;
+
+/**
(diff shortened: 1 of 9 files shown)


sql/src/main/java/com/vmware/ddlog/ir/DDlogIndexDeclaration.java
```diff
@@ -0,0 +1,88 @@
...
+
+public class DDlogIndexDeclaration extends DDlogNode {
+
+    private final String name;
+    private final List<DDlogField> fields;
+    private final DDlogRelationDeclaration baseRelation;
+    private final String baseRelationTypeString;
+
+    public DDlogIndexDeclaration(@Nullable Node node, String name,
+                                    List<DDlogField> fields, DDlogRelationDeclaration baseRelation,
```
mbudiu-vmw, line 41: indentation is inconsistent.
I would also add a little comment describing what this does.

sql/src/main/java/com/vmware/ddlog/ir/DDlogProgram.java
```diff
@@ -36,33 +36,51 @@
...
-        String[] parts = new String[5];
-        parts[0] = String.join("\n", Linq.map(this.imports, DDlogImport::toString)) + "\n";
-        parts[1] = String.join("\n", Linq.map(this.typedefs, DDlogTypeDef::toString));
-        parts[2] = String.join("\n", Linq.map(this.functions, DDlogFunction::toString));
-        parts[3] = String.join("\n", Linq.map(this.relations, DDlogRelationDeclaration::toString));
-        parts[4] = String.join("\n", Linq.map(this.rules, DDlogRule::toString));
-        return String.join("\n", parts);
+        // We do this to preserve the correctness of existing tests, which are sensitive to white-space
+        if (this.indexes.size() == 0) {
+            String[] parts = new String[5];
```
mbudiu-vmw, line 67: This starts to look ugly. Perhaps just make a list and keep adding to it?

mbudiu-vmw (approved):

sql/src/main/java/com/vmware/ddlog/ir/DDlogProgram.java
amytai: Take a look at the update? No matter what, if we want to avoid changing the results of many of the tests, it will look somewhat ugly (i.e. check for whether `this.indexes.size() == 0`.

mbudiu-vmw: Was arguing against allocating exactly 6/5 elements

amytai:
Not sure if we can merge yet -- further testing shows that  index statements cannot come after rules in a DDlog program. Is this correct? If so, I'll need to change the order of the prints in DDlogProgram::toString

ryzhyk:
I don't think that's true. DDlog should not be sensitive to the order of declarations. If it is, it's a bug.

amytai:
Hmmm here's the repro (if it's indeed a bug, I'll copy this over to a separate issue):
```

**Comment by mbudiu-vmw** [comment]
The syntax of your index statement is wrong

**Comment by amytai** [comment]
I think it's correct? If I just move the index declarations up the program compiles fine.
```
import fp
import time
import sql
import sqlop

typedef Thosts = Thosts{id:Option<string>, capacity:Option<signed<64>>, up:Option<bool>}
typedef Tbase_array_table = Tbase_array_table{id:Option<string>, capacity:Option<signed<64>>, col3:Option<signed<64>>}

input relation Rbase_array_table[Tbase_array_table]
output relation Rhosts__view[Thosts]

index Ibase_array_idx(id:Option<string>) on Rbase_array_table[Tbase_array_table{id,_,_}]
index Ihosts_id_up(id:Option<string>,up:Option<bool>) on Rhosts[Thosts{id,_,up}]

Rhosts__view[v0] :- Rhosts[v],var v0 = v.
```

ryzhyk:
I'll have a look.

Unrelated question: I think that SQL `CREATE INDEX` has a very different semantics than DDlog indexes. The former is just a performance optimization, the latter has nothing to do with performance but rather enables the client to query the contents of tables non-incrementally.  I am curious what is the point of this feature?  Is the goal here to use the `CREATE INDEX` syntax to indirectly create DDlog indexes?

amytai:
Yeah, we aren't using them as a performance optimization at all. It's just to enable the DDlog backend to be able to service queries on non-primary keys, which is a feature DCM needs for selects and deletes.

ryzhyk:
I confirm there's a parser bug.  I'll try to fix today.

mbudiu-vmw:
Is this one ready to merge too?

PR #1096, opened 2021-10-04
Fix ddlog-sql test class naming

amytai:
Maven needs test classes to end in *Test, so previously JooqProviderTest<Calcite|Presto> and SetTests weren't actually running when `mvn test` is called. Renamed these test classes and added patch to JooqProviderPrestoTest to make a recent test pass.

This patch also means the number of tests run in IntelliJ is equal to the number of tests run by mvn from the command line.


mbudiu-vmw (approved):

mbudiu-vmw:
Can we merge this?

PR #1098, opened 2021-10-05
Fix parsing ambiguity in index declaration.

ryzhyk:
The following failed to parse:

```
R2[u] :- R1[v], var u = v.

index R1(id: string) on ...
```

The parser interpreted the dot followed by `index` as a field access
(`v.index`).  This is because we deliberately did not declare `index`
as a keyword as we want to allow using it as a valid identifier.

The solution is to perform some extra lookahead in the field access
expression whenever field name is equal to "index".

@amytai

### DIFF
--- a/test/datalog_tests/simple3.dl
+++ b/test/datalog_tests/simple3.dl
@@ -217,6 +217,13 @@ input relation Rhosts[Thosts] primary key (row) (row.id)
 output relation Rhostsv[Thosts]
 Rhostsv[v0] :- Rhosts[v],var v0 = v.
 
+// Make sure 'index' is not parsed as field name ('var v0 = v.index') in this context.
+index Rhosts_by_id(id: string) on Rhosts[Thosts{.id = Some{id}}]
+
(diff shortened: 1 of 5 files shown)


amytai:
Minor comment: what's the reason for not making index a keyboard? You want people to be able to call a relation `index`?

test/datalog_tests/simple3.dl
```diff
@@ -217,6 +217,9 @@ input relation Rhosts[Thosts] primary key (row) (row.id)
 output relation Rhostsv[Thosts]
 Rhostsv[v0] :- Rhosts[v],var v0 = v.
 
+// Make sure 'index' is not parsed as field name ('var v0 = v.index') in this context.
+index Rhosts_by_id(id: string) on Rhosts[Thosts{.id = Some{id}}]
```
mbudiu-vmw, line 221: is there also a positive test where index is a field name?

mbudiu-vmw (approved):

PR #1107, opened 2021-11-08
Self-profiler revamp

ryzhyk:
The self-profiler  remains a useful tool for troubleshooting DDlog performance
issues.  It runs with low overhead, allows enabling/disabling CPU and change
profiling at runtime to only instrument parts of the program, and because it is
integrated into the DDlog runtime it can precisely match each DD operator to the
corresponding DDlog operator.  DDshow does not currently have these features.

In this release we revamp the self profiler to improve its ergonomics.  The new
self-profiler has the following features:

- Produces profiles in the form of interactive HTML tables.  Each
  row in the table represents a DD operator and contains operator
  description, e.g., "Arrange relation 'Rel1' by 'x,y,z'" along with
  links to one or more source code location that this operator
  corresponds to (e.g., all locations where this specific arrangement
  of 'Rel1' is used).

- The new `dump_profile` API dumps the profile into an HTML file
  on the disk instead of returning it as a string.  All profiles
  generated by the same process are generated in the same folder, even
  if the process creates multiple instances of DDlog.  This folder
  also contains a complete snapshot of all DDlog code in the
  program, so that the profiler can show program sources even when
  the program does not run on the same system where it was compiled.

- Internally, the self-profiler represents profiles using a
  well-defined JSON format.  New APIs were added to extract
  each of the four profiles currently supported by DDlog (arrangement
  size profile, peak arrangement size profile, change profile, and
  CPU profile) in the JSON format, for automatic processing by
  third-party tools.


src/Language/DifferentialDatalog/Profile.hs
```diff
@@ -0,0 +1,258 @@
...
+dbgInfoArrangement :: (?specname::String, ?module_paths::ModulePathMap) => Expr -> [Pos] -> [Index] -> Doc
+dbgInfoArrangement pattern used_at indexes = 
+    "::ddlog_profiler::ArrangementDebugInfo::new(" <> cow (pp pattern) <> "," <+> used_pos <> "," <+> used_in_indexes <> ")"
+    where
+    used_pos = "&[" <> commaSep (map mkSourcePos used_at) <> "]"
+    used_in_indexes = "&[" <> commaSep (map (cow_str . name) indexes) <> "]"
+
+-- OperatorDebugInfo::Arrange - arrange the prefix of a rule.
+--
+-- * 'rl' - rule whose prefix is being arranged.
```
mbudiu-vmw, line 83: I don't see any rl

src/Language/DifferentialDatalog/Profile.hs
```diff
@@ -0,0 +1,258 @@
...
+dbgInfoArrange Rule{..} prefix_len key =
+    "::ddlog_profiler::OperatorDebugInfo::arrange(" <> cow key <> "," <> source_pos <> ")"
+    where
+    start = fst $ pos $ head ruleRHS
+    end = snd $ pos $ ruleRHS !! (prefix_len - 1)
+    source_pos = mkSourcePos (start, end)
+
+-- OperatorDebugInfo::Flatmap - Flatmap operator.
+--
+-- * 'rl' - rule.
```
mbudiu-vmw, line 96: same here, you mean Rule{ .. }

src/Language/DifferentialDatalog/Syntax.hs
```diff
@@ -702,17 +708,50 @@ data RuleRHS = RHSLiteral   {rhsPolarity :: Bool, rhsAtom :: Atom}
...
+    (==) _                        _                         = False
+
+rhsrank :: RuleRHS -> Int
+rhsrank RHSLiteral{}   = 0
+rhsrank RHSCondition{} = 1
+rhsrank RHSGroupBy{}   = 2
+rhsrank RHSFlatMap{}   = 3
+rhsrank RHSInspect{}   = 4
+
+instance Ord RuleRHS where
```
mbudiu-vmw, line 740: forgot who needs Ord

src/Language/DifferentialDatalog/Compile.hs
```diff
@@ -956,6 +977,20 @@ compileMainLib d d3log_rel_map nodes cstate =
...
 
+-- The 'SOURCES' static variable contains complete DDlog source code embedded
+-- inside the compiled binary.
+mkSources :: (?specname::String, ?modules::M.Map ModuleName DatalogModule) => Doc
+mkSources =
+    "static SOURCES: &'static [ddlog_profiler::SourceFile] = &[" $$
+    (nest' $ vcommaSep sources)                                  $$
+    "];"
+    where
+    sources = map (\DatalogModule{..} ->
```
mbudiu-vmw, line 988: should there exist an option to generate 'stripped' binaries, without this stuff?

mbudiu-vmw (approved):
This sure was a monster change.
I wonder how much we'd be able to reuse if we switch to a new backend...

rust/template/differential_datalog/src/api/c_api.rs
```diff
@@ -854,19 +869,168 @@ pub unsafe extern "C" fn ddlog_enable_timely_profiling(
...
+    CString::new(prof_file)
+        .map(CString::into_raw)
+        .unwrap_or_else(|e| {
+            prog.eprintln(&format!(
+                "Failed to convert profile filename to a C string: {}",
+                e
+            ));
+            ptr::null_mut()
+        })
+}
```
Kixiron, line 904: This can be simplified by taking and returning `Option<NonNull<_>>`s
```suggestion
pub unsafe extern "C" fn ddlog_dump_profile(
    prog: Option<NonNull<HDDlog>>,
    label: Option<NonNull<raw::c_char>>,
) -> Option<NonNull<raw::c_char>> {
    let prog = prog?.as_ref();
    let label = CStr::from_ptr(label?)
        .map_err(|err| prog.eprintln(&format!("invalid label string: {}", err)))
        .ok()?;

    let prof_file = prog
        .dump_profile(label.as_deref())
        .map_err(|err| prog.eprintln(&format!("Failed to dump profile: {}", err)))
        .ok()?;

    match CString::new(prof_file).map(CString::into_raw) {
        Ok(string) => NonNull::new(string),
        Err(err) => {
            prog.eprintln(&format!(
                "Failed to convert profile filename to a C string: {}",
                err,
            ));
            None
        }
    }
}
```

doc/profiling/profiling.md
```diff
@@ -0,0 +1,370 @@
...
+fragments in the corresponding profile entry:
+
+![Numbers relation arranged by ()](size-numbers-2.png)
+
+There are several other kinds of arrangements in the profile, e.g., the
+`group_by` operator arranges its input using group-by variables as a key.  Each
+profile entry contains a hyperlink to documentation that describes the meaning
+of the associated operator.
+
+Note that in general arrangement size can be larger than the number of
```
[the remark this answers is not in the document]
ryzhyk: Any arrangement that changes over time can have duplicate records with different timestamp.  With `DIFFERENTIAL_EAGER_MERGE` enabled and with enough slack time these will eventually get merged.

rust/template/differential_datalog/src/api/c_api.rs
ryzhyk: Nice!  We should apply this consistently throughout the API.

src/Language/DifferentialDatalog/Syntax.hs
ryzhyk: `Set` and `Map`.

rust/template/ddlog_profiler/profiler_ui/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "profiler_ui",
+  "version": "1.0.0",
+  "description": "DDlog profiler UI",
+  "main": "index.html",
+  "dependencies": {
```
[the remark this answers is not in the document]
ryzhyk: I moved all dependencies to devDependencies.  It seems to work.

src/Language/DifferentialDatalog/Compile.hs
ryzhyk: Probably, but I'm inclined to make it "future work".

src/Language/DifferentialDatalog/Syntax.hs
mbudiu-vmw: This is Ord for RuleRHS, not for the data.

ryzhyk: Any time we need e.g., a set of rules, we need `Ord Rule`, which in turn requires `Ord RuleRHS`.

PR #1113, opened 2021-11-12
Reduce dependency bloat

Kixiron:
* Started trying to remove `num` from the dependency tree
* Removed some of the "default dependencies" from generated crates, this should allow better compilation scheduling since crates don't have to wait on unnecessary dependencies
* Discovered that some files are in CLRF form, this will be fixed in a later PR
* Removed dependencies on `fnv` and `twox-hash` and replaced with a single dependence on `xxhash-rust`, it has zero dependencies and is very light
* Removed old dependence on `time` for all generated crates, this reduced the dependency tree a good bit
* Minor refactors to some massive match statements
* Added some documentation to the `DDVal`/`DDValue` innards
*


mbudiu-vmw:
Some of these files show every line as different. Is this due to CR/LF?
Thiis makes it hard to review.

Kixiron:
Yes, those are files that were changed from CLRF to LF, I'll make a PR that bites the bullet and standardizes things after this

lib/ddlog_std.rs
```diff
@@ -47,11 +47,12 @@ use std::{
     sync::Arc as StdArc,
     vec::{self, Vec as StdVec},
 };
-use twox_hash::{XxHash32, XxHash64};
+use xxhash_rust::{xxh3::Xxh3, xxh32::Xxh32};
 
 const XX_SEED1: u64 = 0x23b691a751d0e108;
 const XX_SEED2: u64 = 0x20b09801dce5ff84;
 
+#[inline]
```
ryzhyk, line 55: why don't we trust the compiler with all this inlining?

ryzhyk (commented):
There's a bunch of test failures.  Some of them are probably caused by full disks on CI machines.  Those should now be fixed.  Others probably just require changing reference test outputs due to the use of a different hashing function.

I think most of the maps that use FNV today can be regular default HashMaps.  I mostly used FNV to ensure deterministic iteration order. This is necessary e.g., when building dataflow graphs to make sure that each DD worker ends up with the same operator id assignment.  In most other cases there's no reason for non-standard hasher.

lib/ddlog_std.rs
Kixiron: We do, however this makes the functions viable for cross-crate inlining

PR #1114, opened 2021-11-12
Feature-gate dynamic deserialization.

ryzhyk:
We generate a `AnyDeserialize` implementation that allows
deserializing into instances of `Any` given relation id.
This impl is not used by most applications, but causes massive
code bloat.  The reason is that is uses erased_serde, which
forces all `Deserialize` impls to get specialized for
`erased_serde::Deserializer`.  We feature-gate this impl, so only
clients who want to use it can enable it.


Kixiron (approved):
LGTM

PR #1117, opened 2021-11-18
Efficient SQL join implementation

mbudiu-vmw:
Previously joins were implemented as cartesian products followed by filtering.
With this PR they are implemented as DDlog joins.
The PR is more complicated than expected because it had to rework the way objects are named and looked-up.
Joins were also incorrect: one should not join on null values. This is fixed in this PR.
Note: left joins are still not implemented, but they should be easier to do with this infrastructure.

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/DDlogHandle.java
+++ b/sql/src/main/java/com/vmware/ddlog/DDlogHandle.java
@@ -113,11 +113,7 @@ public class DDlogHandle {
      * This corresponds to the naming convention followed by the SQL -> DDlog compiler
      */
     public String relationNameToTableName(final String relationName) {
-        for (Map.Entry<String, DDlogRelationDeclaration> me : this.program.tableToRelation.entrySet()) {
-            if (me.getValue().getName().name.equals(relationName))
-                return me.getKey();
-        }
-        throw new RuntimeException("Table not found for relation " + relationName);
+        return this.program.relationNameToTableName(relationName);
     }
 
     public void queryIndex(String index, DDlogRecord key, Consumer<DDlogRecord> consumer) throws DDlogException {
--- a/sql/src/main/java/com/vmware/ddlog/DDlogHandle.java
+++ b/sql/src/main/java/com/vmware/ddlog/DDlogHandle.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright (c) 2021 VMware, Inc.
+ * SPDX-License-Identifier: MIT
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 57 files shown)


mbudiu-vmw:
This looks big, but there are many renames that percolate across many files.

lalithsuresh (approved):

sql/src/main/java/com/vmware/ddlog/DDlogHandle.java
```diff
@@ -0,0 +1,126 @@
...
+import java.util.Set;
+import java.util.function.Consumer;
+
+public class DDlogHandle {
+    DDlogProgram program;
+    DDlogAPI api;
+
+    // Unfortunately, `create index` statements have to be passed separately, because neither Calcite nor Presto
+    // supports them, so we must pass them as SQL strings.
+    public <R extends SqlStatement> DDlogHandle(final List<R> ddl, final ToPrestoTranslator<R> translator,
```
amytai, line 53: I can foresee problems with this class, which we can fix now or in a follow-up PR.

We want to avoid recompiling the DDlog program everytime we want to run benchmarks. i.e., if we know that /tmp/program.dl isn't going to change, we don't want to go through the SQL-->DDlog _and_ Rust compilation process. This _could_ be fixed with a simple boolean passed to this constructor (`compile` flag, which, if false, skips lines 56-70). 
However, this isn't enough because you've moved some DDlogProgram-specific functions into this class, such as ddlogTableTypeName, that require `this.program` to be not-null.

Another reason is, based on some idiosyncracy in DCM's workflow, we need to make a dummy "DDlogHandle", that has fields `this.program = null; this.api = null`.  So we need an empty constructor, for example.

amytai (changes requested):

sql/src/main/java/com/vmware/ddlog/DDlogHandle.java
lalithsuresh: We can also figure out how to remove the dummy workflow.

sql/src/main/java/com/vmware/ddlog/DDlogJooqProvider.java
```diff
@@ -148,6 +146,10 @@ public DDlogJooqProvider(final DDlogAPI dDlogAPI, final List<H2SqlStatement> sql
         this.whereHelper.seal();
     }
 
+    public static String normalizeTableName(String tableName) {
```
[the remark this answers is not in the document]
mbudiu-vmw: not sure I caught all of them, but now there is the infrastructure to do it

sql/src/main/java/com/vmware/ddlog/DDlogHandle.java
mbudiu-vmw: This PR is big enough as it is.

PR #1124, opened 2021-11-30
Fix bug; add anchor on each table row

mbudiu-vmw:


ryzhyk:
@mbudiu-vmw, any progress on the issue with anchor's not working on collapsed table rows?

mbudiu-vmw:
I believe that the solution is to actually check the url and if it contains an anchor to automatically expand the document to that path.

mbudiu-vmw:
Would starting with all cells expanded in that case be acceptable?

ryzhyk:
I think so.

mbudiu-vmw:
Fixes #1124

PR #1127, opened 2021-12-01
Add compile flag to DDlogHandle

amytai:
Because large DDlog programs can take long to compile, if we know that /tmp/program.dl won't change, a caller can choose not to recompile program.dl.


amytai:
We use this compile flag both in DCM tests and in e2e DCM executions.

mbudiu-vmw (approved):

PR #1139, opened 2021-12-10
Treat DECIMAL types as Long in DDlogJooqProvider

amytai:
Jooq translates BIGINT types into Types.DECIMAL, which means they should be treated as Longs to prevent overflow.

### DIFF
--- a/sql/src/test/java/ddlog/JooqProviderTestBase.java
+++ b/sql/src/test/java/ddlog/JooqProviderTestBase.java
@@ -687,4 +687,14 @@ public abstract class JooqProviderTestBase {
         assertTrue(readFromInput.contains(test3));
     }
 
+    @Test
+    public void testBigInt() {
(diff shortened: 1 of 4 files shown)


sql/src/test/java/ddlog/JooqProviderTestBase.java
```diff
@@ -687,4 +687,14 @@ public void testIndexSelect() {
         assertTrue(readFromInput.contains(test3));
     }
 
+    @Test
+    public void testBigInt() {
```
mbudiu-vmw, line 691: no tests that validate the output DDlog?

amytai: I don't think that's necessary. This test fails if you run without the patch.

mbudiu-vmw (approved):

PR #1144, opened 2022-01-25
sql: bump H2 and JOOQ to latest versions and address array syntax change

lalithsuresh:
The new H2 version uses the standard SQL syntax for arrays (that is, qualified with the type of the array). 

@mbudiu-vmw I addressed most issues, but two tests still fail. They both relate to a type-confusion on the return type of array_agg(). I couldn't immediately tell if it's to do with the jooq code or whether it's on the ddlog -> sql compiler side. Please have a look, otherwise I'll dig deeper later this week.

### DIFF
--- a/sql/src/main/java/com/vmware/ddlog/util/sql/PrestoToH2Translator.java
+++ b/sql/src/main/java/com/vmware/ddlog/util/sql/PrestoToH2Translator.java
@@ -64,7 +64,9 @@ class PrestoToH2 extends AstVisitor<String, String> {
                 final Pattern arrayType = Pattern.compile("ARRAY\\((.+)\\)");
                 Matcher m = arrayType.matcher(h2Type);
                 if (m.find()) {
-                    h2Type = "array";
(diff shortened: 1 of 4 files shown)


sql/src/main/java/com/vmware/ddlog/util/sql/PrestoToH2Translator.java
```diff
@@ -64,7 +64,9 @@ protected String visitCreateTable(final CreateTable node, final String sql) {
                 final Pattern arrayType = Pattern.compile("ARRAY\\((.+)\\)");
                 Matcher m = arrayType.matcher(h2Type);
                 if (m.find()) {
-                    h2Type = "array";
+                    final String match = m.group();
+                    final String arrayTypeStr = match.substring(6, match.length() - 1);
```
mbudiu-vmw, line 68: I think the right way to do this is to get group(1).

mbudiu-vmw (changes requested):

PR #1148, opened 2022-02-10
SQL compiler: Use interned strings

mbudiu-vmw:


lalithsuresh (approved):
I assume all the dynamic tests pass? 

Are there scenarios where we should not be using interned strings?

mbudiu-vmw:
Yes, the SQL tests pass.
For typical SQL programs I expect that intern strings should always help. Only if you create lots of temporary strings that you use in computations this could be detrimental.

PR #1153, opened 2022-02-17
[SQL compiler] Use int64 instead of bigint

mbudiu-vmw:


mbudiu-vmw:
@lalithsuresh this should be faster too

mbudiu-vmw:
I have piggybacked on this PR a set_agg aggregation function as well, to avoid having to rebase it after this is merged.

lalithsuresh:
@mbudiu-vmw could you try a dynamic test that calls set_agg and set_contains? :)

PR #1156, opened 2022-04-22
Upgrade libraries

mbudiu-vmw:

@lalithsuresh please take a look. This closes some security holes, so we should do it.
There seems to be at least one JOOQ bug introduced by this migration, and there is a (ugly) workaround.


lalithsuresh:
@mbudiu-vmw LGTM.

I've encountered the OTHER/VARCHAR issue as well when I tried to upgrade. The ugly workaround is the only option for now. Best to fix it later.

Issue #1158, opened 2022-07-18
Type constraints for right-hand operand of shift operators


blp:
The following program:
```
input relation X(x: bit<16>, y: bit<16>)
output relation Y(x: bit<16>, y: bit<16>, z: bit<16>)
Y(x, y, x << y) :- X(x, y).
```
yields the following compiler error that I don't expect (with release v1.2.3 and earlier releases):
```
error: /home/blp/nerpa/nerpa/ofp4/test.dl:3.6-3.7: Unsatisfiable bit-width constraint '16 = 0' in
expected type: bit<16>
actual type: bit<32>
in
expression 'y'
Y(x, y, x << y) :- X(x, y).
     ^
```
I can fix it by casting `y` to `u32` in the shift expression, like this:
```
Y(x, y, x << (y as u32)) :- X(x, y).
```
I would prefer that `<<` (and presumably `>>`) accept any integer type as its right-hand operand.

mbudiu-vmw:
Probably any unsigned integer type only could work. Signed values introduce additional complications.

blp:
That's reasonable.

blp:
If https://github.com/vmware/differential-datalog/pull/1159 gets merged, and this bug gets fixed, then the documentation added by the PR should be updated as well.

PR #1159, opened 2022-07-19
Document and add a test for bit-shift operator behavior.

blp:
This documents the current requirement that the right-hand operand of
<< and >> have type bit<32>.  I tend to consider that a bug (see
https://github.com/vmware/differential-datalog/issues/1158), so I
hope that it will be relaxed in the future (and the documentation
updated).

Signed-off-by: Ben Pfaff <bpfaff@vmware.com>


blp:
I don't know how to evaluate the failed tests. Are they related to something I did?

ryzhyk:
Thank you, @blp !

ryzhyk:
@blp PS. Test failures are unrelated to your PR.

PR #1160, opened 2022-07-23
Use LTS 18.28 for Apple M1

meme:
This allows Differential Datalog to be built on M1 Macs


ryzhyk:
Great, thank you, @meme!

